PR #23, opened 2024-04-05
Remove the patch that deleted fixed UIDs from istiod charts

dgn:
This patch is not needed anymore as we removed support for v1.20 - the lines it attempts to remove have been removed upstream as part of istio/istio#45394.


jwendell (approved):
This is the kind of change that we approve blindly as we don't have a smoke bookinfo test to validate sidecar injection

PR #25, opened 2024-04-08
Centralize handling of conflicts and other common errors

luksa:


pkg/reconciler/reconciler_test.go
```diff
@@ -216,6 +214,46 @@ func TestReconcile(t *testing.T) {
...
+			objects: []client.Object{
+				&v1alpha1.Istio{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:       key.Name,
+						Finalizers: []string{testFinalizer},
+					},
+				},
+			},
+			setup: func(g *WithT, mock *mockReconciler) {
+				mock.reconcileError = apierrors.NewForbidden(schema.GroupResource{}, "foo", fmt.Errorf("cannot set blockOwnerDeletion in this case because cannot find RESTMapping for APIVersion ..."))
```
luksa, line 248: @dgn Do you remember if this was the error that we were seeing?

pkg/reconciler/reconciler_test.go
```diff
@@ -216,6 +214,47 @@ func TestReconcile(t *testing.T) {
...
+						Name:       key.Name,
+						Finalizers: []string{testFinalizer},
+					},
+				},
+			},
+			setup: func(g *WithT, mock *mockReconciler) {
+				mock.reconcileError = apierrors.NewConflict(schema.GroupResource{}, "foo", fmt.Errorf("simulated conflict"))
+			},
+			assert: func(g *WithT, cl client.Client, result ctrl.Result, err error, mock *mockReconciler) {
+				g.Expect(result).To(Equal(reconcile.Result{RequeueAfter: 2 * time.Second}))
```
jwendell, line 231: nit: these numbers seem to be fragile to changes. Worth to put them in constants so once changed in one place we don't need to look around and change all places?

pkg/reconciler/reconciler_test.go
dgn: yes, pretty sure that was it

dgn (approved):
lgtm. adding a hold in case we want to add constants for the requeueAfter values (I think it makes sense)

/hold

pkg/reconciler/reconciler_test.go
luksa: It depends whether there is a hard requirement for some value to be exactly 2s. If yes, then the test should not use the same constant that's used in the code. Obviously this is not the case here. 

But I've been thinking about this and I'm now pretty sure that we should be using exponential backoff here.

dgn: Take a look at https://github.com/kubernetes-sigs/controller-runtime/issues/808

luksa: What specifically?

dgn: Okay nevermind. Looking at that issue (and code) I think controller-runtime will do exponential backoffs by default, but only for `Requeue:true`, not for `RequeueAfter` Results. Which is unfortunate, as we would need "requeue after at least duration x but on subsequent requeues do an exponential backoff"

PR #31, opened 2024-04-09
Add Debug information when e2e Test run Fails

### LINKED ISSUE #30: Improve debug information when e2e test run fail
Now we have some debug information when the e2e test run fails, but we need to improve and add more information that can be helpful to detect earlier the failure.

Current debug information:

- Istio CR Yaml
- Pods in Istio CR namespace
- Logs from sail-operator pod
- Logs from istiod pod

We need to add more debugging information as:

- Operator Pod YAML
- OC describe
- More logs if possible

fjglira:
Fixes #30

### DIFF
--- a/tests/e2e/util/common/e2e_utils.go
+++ b/tests/e2e/util/common/e2e_utils.go
@@ -36,6 +36,8 @@ var (
 	deploymentName        = env.Get("DEPLOYMENT_NAME", "sail-operator")
 	controlPlaneNamespace = env.Get("CONTROL_PLANE_NS", "istio-system")
 	istioName             = env.Get("ISTIO_NAME", "default")
+	istioCniNamespace     = env.Get("ISTIOCNI_NAMESPACE", "istio-cni")
+	istioCniName          = env.Get("ISTIOCNI_NAME", "default")
 )
 
 // key returns the client.ObjectKey for the given name and namespace. If no namespace is provided, it returns a key cluster scoped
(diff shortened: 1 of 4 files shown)


tests/e2e/util/common/e2e_utils.go
```diff
@@ -90,28 +92,57 @@ func CheckNamespaceEmpty(ctx SpecContext, cl client.Client, ns string) {
...
+	operator, err := kubectl.GetYAML(namespace, "deployment", deploymentName)
 	if err != nil {
-		GinkgoWriter.Println("Error getting Istio CR: ", err)
+		GinkgoWriter.Println("Error getting operator deployment yaml: ", err)
 	}
-	GinkgoWriter.Println("Istio CR: \n", resource)
+	GinkgoWriter.Println("Operator deployment: \n", operator)
 
-	output, err := kubectl.GetPods(controlPlaneNamespace, "-o wide")
+	describe, err := kubectl.Describe(namespace, "deployment", deploymentName)
```
luksa, line 105: Why do we need both the YAML and the description? They both show the same information. I'd remove the description and leave just the YAML.

tests/e2e/util/kubectl/kubectl.go
```diff
@@ -141,6 +141,21 @@ func GetPods(ns string, args ...string) (string, error) {
...
+func Describe(ns, kind, name string) (string, error) {
+	cmd := kubectl("describe %s %s %s", kind, name, nsflag(ns))
+	output, err := shell.ExecuteCommand(cmd)
+	if err != nil {
+		return "", fmt.Errorf("error describing resource: %w, output: %s", err, output)
+	}
+
+	return output, nil
+}
+
```
luksa, line 158: This is now unused code, which we should probably remove.

tests/e2e/util/common/e2e_utils.go
```diff
@@ -90,7 +92,23 @@ func CheckNamespaceEmpty(ctx SpecContext, cl client.Client, ns string) {
 
 func LogDebugInfo() {
 	// General debugging information to help diagnose the failure
-	// TODO: Add more debugging information for others resources
+	// TODO: Add the creation of file with this information to be attached to the test report
+
```
nrfox, line 96: `kubectl events -n <namespace>` is also generally useful debug info as well.

fjglira: Thanks. I'll add it

luksa (approved):

PR #33, opened 2024-04-10
Add `operator-name` target

dgn:
This is an internal target that will help us to change the name of the operator in our downstream repository. We're already using the `OPERATOR_NAME` variable everywhere except the operator-sdk's `PROJECT` file, so this fills the gap.


jwendell (commented):
Not sure if it's related to this PR or not, but the PROJECT file contains many references to maistra. Is this in the scope of this PR as well?

jwendell (approved):
Putting a hold in case you want to address the maistra thing on this pr.

/hold

dgn:
I'll update those references in another PR

PR #36, opened 2024-04-11
Create namespace if it doesn't exist in README example

nrfox:
Updates the README example to create the namespace if it doesn't exist. Otherwise when you just run the example without creating the namespace first the reconciliation will fail.


luksa (approved):

jwendell:
Note that the README in the bundle directory is more complete and targets end users.

/ok-to-test

PR #60, opened 2024-04-18
e2e tests: skip deletion of operator namespace on OCP

dgn:
When removing the namespace, all related container image repositories are deleted from the internal registry, causing following tests to fail.

### DIFF
--- a/tests/e2e/controlplane/control_plane_test.go
+++ b/tests/e2e/controlplane/control_plane_test.go
@@ -274,6 +274,10 @@ spec:
 			To(Succeed(), "Operator failed to be deleted")
 		GinkgoWriter.Println("Operator uninstalled")
 
+		if ocp {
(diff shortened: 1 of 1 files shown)


luksa (approved):

tests/e2e/controlplane/control_plane_test.go
```diff
@@ -274,6 +274,10 @@ spec:
 			To(Succeed(), "Operator failed to be deleted")
 		GinkgoWriter.Println("Operator uninstalled")
 
+		if ocp {
+			Success("Skipping deletion of operator namespace to avoid removal of operator container image from internal registry")
```
fjglira, line 278: Here, we should use GinkgoWriter instead of Success because Success is intended only for logging successful steps. The idea is to clearly indicate to anyone reading the test—and not just the result—that Success is a success step in the test case, and should not be used for general information or debugging purposes.

PR #65, opened 2024-04-19
Ensure control plane is deployed immediately when the target namespace is created

luksa:


controllers/istiorevision/istiorevision_controller.go
```diff
@@ -433,11 +437,27 @@ func istiodDeploymentKey(rev *v1alpha1.IstioRevision) client.ObjectKey {
...
+		if rev.Spec.Namespace == ns.GetName() {
+			requests = append(requests, reconcile.Request{NamespacedName: types.NamespacedName{Name: rev.Name}})
+		}
+	}
+
+	// Check if the namespace references an IstioRevision in its labels
 	revision := getReferencedRevisionFromNamespace(ns.GetLabels())
 	if revision != "" {
-		return []reconcile.Request{{NamespacedName: types.NamespacedName{Name: revision}}}
+		requests = append(requests, reconcile.Request{NamespacedName: types.NamespacedName{Name: revision}})
```
jwendell, line 458: does it make sense to use a set instead of raw arrays? Is there any chance to get duplicated entries here? if yes, would that be a problem?

luksa: The queue handles duplicates.

jwendell (approved):

PR #112, opened 2024-05-07
Automate the bump of stable Istio releases

jwendell:
Modify the existing `update-istio.sh` script to also look for new stable versions and update `versions.yaml` if any is found.

This automates the inclusion of new patch releases (currently done in a nightly job), but does not remove anything. Removals should be manual.

Note that it only looks for new patch releases of *existing* versions in `versions.yaml` file. It does not look for a new *minor* version, only patch versions. New minor or major versions should still be added manually.

### DIFF
--- a/hack/update-istio.sh
+++ b/hack/update-istio.sh
@@ -19,37 +19,113 @@ set -euo pipefail
 SLEEP_TIME=10
 VERSIONS_YAML_FILE=${VERSIONS_YAML_FILE:-"versions.yaml"}
 
-COMMIT=$(yq '.versions[] | select(.name == "latest") | "git ls-remote --heads " + .repo + ".git " + .branch + " | cut -f 1"' "${VERSIONS_YAML_FILE}" | sh)
-CURRENT=$(yq '.versions[] | select(.name == "latest") | .commit' "${VERSIONS_YAML_FILE}")
+# Add a new entry in versions.yaml file.
+# First argument is the new version (e.g. 1.22.5)
+# Second argument is the old version (e.g. 1.22.4)
+#
+# The new entry will be placed immediately before the old one
+function add_stable_version() {
+    echo "Adding new stable version: ${1}"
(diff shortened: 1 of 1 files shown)


hack/update-istio.sh
```diff
@@ -19,37 +19,89 @@ set -euo pipefail
...
+    "https://istio-release.storage.googleapis.com/charts/base-${1}.tgz",
+    "https://istio-release.storage.googleapis.com/charts/istiod-${1}.tgz",
+    "https://istio-release.storage.googleapis.com/charts/gateway-${1}.tgz",
+    "https://istio-release.storage.googleapis.com/charts/cni-${1}.tgz",
+    "https://istio-release.storage.googleapis.com/charts/ztunnel-${1}.tgz"
+    ]
+}
+END
+    )
+    yq -i ".versions += ${template}" "${VERSIONS_YAML_FILE}"
```
luksa, line 40: This adds the version to the end of the list. Since the version order is important, this should add the version in the correct spot.

jwendell: hmmm, ok. Why does the order matter? What would be the correct order?

luksa: https://github.com/istio-ecosystem/sail-operator/blob/main/versions.yaml#L11

+ the order in versions.yaml is the order in which the versions appear in drop-downs and other lists.

fjglira: we need the version.yaml in order because we used in the e2e test and if we add a upgrade e2e test will be very useful also to have it in order. If not we will need to do some extra steps to order the versions before to start making the updates

dgn: maybe we can add a comment to versions.yaml that explains that/why correct order is necessary? or we explain it in a markdown file somewhere

jwendell: done

dgn (approved):
works like a charm, thank you!

PR #118, opened 2024-05-10
Actually force-delete resources after test

asmigala:


fjglira (approved):
lgtm, if fails later the deletion we will see the error from the Delete func

Issue #110, opened 2024-05-07
Upload helm chart package to releases


dgn:
Whenever we create a release, we'll also want to create a helm chart tarball that we can release e.g. here on GitHub releases

Acceptance Criteria:
* generate a Helm chart tarball when a new tag is created in the repository
* upload the Helm chart tarball to GitHub Releases
* tests:
  * verify that the chart installs correctly
  * chart linting

fjglira:
This can be done by adding a make target called `helm-package` that runs the helm package command pointing to the helm chart directory and then using some sort of automation to move the tar to the release section here in github. Can I assign this task to me @dgn ?

dgn:
sure thing!

fjglira:
@dgn check please this PR https://github.com/istio-ecosystem/sail-operator/pull/137, I'm only adding a new target to do the helm-package. I think the automatic release can be handled by creating a new GitHub action workflow that handles the creation of the version every time we create a new tag in the repository (We need to check the proper workflow). but I'm not sure if we can create actions in this repository and if I can start to make some test using those workflows

dgn:
I'm not sure that we should introduce a new CI mechanism- if possible I'd prefer us to stick with Istio Prow.

I approved your PR, I think the main task will be setting up the automation though

fjglira:
Yes, the main task will be the automation. Ok, I'll check if there is a way to do it using prow. I'll let you know

fjglira:
@dgn after some research yesterday we have two option here:
* Use a github workflow to create a action that monitors the tag creation and every time that we create a new tag a new release is created invoking the make target to generate the file and add the asset to the release
* The second option is more complex but is as upstream istio does: they have a script that invokes a code made specifically to create releases. 
     *  They have a bash script located in the istio repo that invokes `release-builder publish --release "${WORK_DIR}/out"`. That command goes to release builder repo and more information is here: https://github.dev/istio/release-builder/blob/c0969495c8a87f7562887f6b1cd70bb997398fe9/README.md#L65-L66.
     * They have an external script and process that builds everything. 
     
 For us will be more simple because we only need to publish initially the package in the release section and what we publish is the tar package from the helm charts

bmangoen:
> * tests:
>   
>   * verify that the chart installs correctly
>   * chart linting

[...]

PR #137, opened 2024-05-27
Create make target for helm package

fjglira:
This is part of this task: https://github.com/istio-ecosystem/sail-operator/issues/110

The idea is to have a make target to be able to create the package with all the charts of the operator and create later a job/automation to be able to automatically create the release is the release section of github


fjglira:
@dgn do we want to be able to set `--version` to set a custom version or do we prefer to take the version from the chart (By using the `helm package` command it will take by default the version from the chart)

dgn (approved):
lgtm

PR #144, opened 2024-05-30
Move user documentation into separate file

dgn:


sridhargaddam (approved):

jwendell (commented):
Why not the other way around? Main README targets end users, CONTRIBUTING.md or DEVELOPERS.md targeting developers? It makes more sense to me.

PR #152, opened 2024-05-31
Return to updating istio and client-go

jwendell:
We are no longer blocked by cachito.


tools/update_deps.sh
```diff
@@ -33,8 +33,8 @@ make update-common
 
 # Update go dependencies (skipped until cachito supports 1.22, see https://issues.redhat.com/browse/OSSM-5988)
```
dgn, line 34: I guess we can update that comment

jwendell: 🤦

dgn (approved):

PR #162, opened 2024-06-12
Add install instructions to user docs

dgn:


docs/README.md
```diff
@@ -7,7 +7,27 @@ tbd
...
+
+1. In the OpenShift Console, navigate to the OperatorHub by clicking **Operator** -> **Operator Hub** in the left side-pane.
+
+1. Search for "sail".
+
+1. Locate the sail-operator, and click to select it.
+
+1. When the prompt that discusses the community operator appears, click **Continue**.
+
+1. Verify the sail-operator is version 0.1, and click **Install**.
```
jwendell, line 21: Should we skip this step? Having it means this doc will become outdated very often.

jwendell: I mean, skip the version check.

FilipB: +1 but we should probably remove other channels to avoid confusion, we still have there 3.0-nigthly and 3.0-dp1

dgn: I removed the step, clicking install is now part of the previous one

jwendell (approved):

PR #164, opened 2024-06-13
Fix e2e test failure with single version in versions.yaml

asmigala:


pkg/test/util/supportedversion/supportedversion.go
```diff
@@ -49,7 +49,9 @@ func init() {
 
 	List = versions.Versions
 	Default = List[0].Name
-	Old = List[1].Name
+	if len(List) > 1 {
+		Old = List[1].Name
+	}
```
luksa, line 54: This will cause `Old` to be `""`, which means that the upgrade tests will set `Istio.spec.version` to `""` ([here](https://github.com/istio-ecosystem/sail-operator/blob/498e3cf169def506a87f66bca5d7f82a5f332eba/tests/integration/api/istio_test.go#L249), for example), which will cause `Istio` to default to the only version that's available. Because the old and the new version end up being the same, the upgrade tests will probably fail (or, at least won't actually test upgrades).

We should probably explicitly skip the upgrade tests if only one version is available.

asmigala: yep, already working on a fix

asmigala: @luksa integration tests passing with the latest change with the downstream versions.yaml

luksa (approved):

PR #171, opened 2024-06-20
Add docs for sail-operator installation using CLI

bmangoen:


docs/README.md
```diff
@@ -23,6 +25,46 @@ tbd
...
+      namespace: openshift-operators
+    spec:
+      channel: "3.0-nightly"
+      installPlanApproval: Automatic
+      name: sailoperator
+      source: community-operators
+      sourceNamespace: openshift-marketplace
+    ```
+
+1. Verify the installation succeeded by inspecting the CSV file.
```
sridhargaddam, line 58: nit: If you happen to re-spin the PR.

```suggestion
1. Verify that the installation succeeded by inspecting the CSV file.
```

dgn (approved):

docs/README.md
```diff
@@ -23,6 +25,46 @@ tbd
...
+*Prerequisites*
+
+* You must have `admin` privileges.
+
+*Steps*
+
+1. Create the `openshift-operators` namespace (if it does not already exist).
+
+    ```bash
+    $ kubectl get ns openshift-operators --ignore-not-found || kubectl create namespace openshift-operators
```
luksa, line 39: Doesn't the `--ignore-not-found` option make the entire first command useless?

docs/README.md
```diff
@@ -23,6 +25,46 @@ tbd
 1. Click **Operators** -> **Installed Operators** to verify that the sail-operator 
 is installed. `Succeeded` should appear in the **Status** column.
 
+#### Installing using the CLI
+
+*Prerequisites*
+
+* You must have `admin` privileges.
```
luksa, line 32: What are `admin` privileges? Do you mean the `admin` ClusterRole? Does that suffice or does the user have to have the `cluster-admin` ClusterRole?

docs/README.md
luksa: Hmm, not just useless, but it actually prevents the `kubectl create` command from ever being invoked, since it always succeeds.

docs/README.md
bmangoen: I guess that because we want to install the operator in the `openshift-operators` namespace, the `cluster-admin` is needed. Going to fix it

docs/README.md
bmangoen: You're right, going to change this

luksa: I think you can just skip the first command. This isn't a shell script, so the user will usually know whether the namespace exists or not (and will thus know whether to run the `kubectl create` command or not). And even if they run it when they don't need to and get an error, that's not that bad. I think it's more important for the instructions to be as simple and short as possible (so, a simple `kubectl create` is better then `kubectl get || kubectl create`).

dgn: sorry I overlooked this. will this ever be the case though? In OCP that namespace should always exist

PR #175, opened 2024-06-24
Display a friendly message for missing K8s cluster

sridhargaddam:
Some of the Makefile commands require a running K8s cluster. When these commands are issued without an active cluster, we currently do not display a proper message, and users have to figure it out manually. This PR fixes that by verifying access to a running cluster and displaying a friendly message if not.


dgn (approved):
Thanks Sridhar!

PR #203, opened 2024-07-15
Add helm-publish make target

bmangoen:
According to this issue #110, we want to create a release and upload the helm charts package tarball as a release asset


fjglira (approved):
LGTM, this will be executed in a postsubmit job only in release branches?

bmangoen:
> LGTM, this will be executed in a postsubmit job only in release branches?
It is the idea to do so.

Hum we might have to separate the release/tag creation and the upload of the helm charts tarball because `gh release create <tag>` will fail the second time it will be executed (the tag would already exist).
WDYT @dgn ?

PR #209, opened 2024-07-16
Change API domain to sailoperator.io

dgn:


luksa:
Any chance we can use `sailoperator.io` instead of `sail-operator.io`? Having to type dashes in `apiVersion` always feels weird to me.

dgn:
> Any chance we can use `sailoperator.io` instead of `sail-operator.io`? Having to type dashes in `apiVersion` always feels weird to me.

Hmm. It's a good point but having the repo name as `sail-operator` and API domain as `sailoperator` would be inconsistent

PR #212, opened 2024-07-17
enhancements: add SEP for revision tag support

dgn:


enhancements/SEP2-revision-tags.md
```diff
@@ -0,0 +1,45 @@
...
+## Goals
+* Provide revision tag support in sail-operator APIs so users don't have to use istioctl for basic revision tag operations
+
+## Non-goals
+* Compatibility with manual revision tag creation using istioctl. There might be a way to disable the operator functionality to avoid conflicts when creating revision tags manually, but that's it - you either do it yourself or let the operator do it
+
+## Design
+
+### User Stories
+1. As a user of sail-operator's RevisionBased update strategy, I want to be able to use the `istio-injection=enabled` label on my namespaces and pods.
```
sridhargaddam, line 21:
```suggestion
1. As a user of sail-operator's RevisionBased update strategy, I want to be able to use the `istio-injection=enabled` label on my namespaces.
```

We can either remove pods from the above sentence (as that label is not applicable for pods) or maybe use `sidecar.istio.io/inject=true` for pods.

enhancements/SEP2-revision-tags.md
```diff
@@ -0,0 +1,45 @@
...
+1. As a platform engineer, I want my application teams to be able to use a fixed label for proxy injection without having to know which version of Istio is running in the cluster, so that I can perform upgrades in the background without the application teams having to make changes to use the new version.
+
+### API Changes
+We will add a new field `revisionTags` of type `[]string` to the `Istio` CRD. It will be located under `spec.updateStrategy.revisionTags`. Whenever the sail-operator deploys a new revision, it will update all the tags listed in the `revisionTags` field to point to this new revision. The default value for the field will be `{"default"}` - this way, the standard injection label `istio-injection=true` will work out of the box for RevisionBased deployments.
```
sridhargaddam, line 25: For RevisionBased deployments, i thought we would use "istio.io/rev=xxxx" and not "istio-injection=xxxx".

enhancements/SEP2-revision-tags.md
```diff
@@ -0,0 +1,45 @@
...
+
+We will also need to remove the `values.revisionTags` field (which is how the upstream charts expose this feature). We can still set that field to trigger creation of the tags, but we should not expose it to users.
+
+### Architecture
+We will need to update the sail-operators mechanism to detect revisions that are being used. Today, we only look at the `istio.io/rev` label's value to check which revisions are in use. But when revision tags are used, those values will be mere aliases, so we have to improve our detection mechanism. The most correct way is probably to look at the revision annotation on the pods that is set by Istio during injection. That requires inspecting every pod in the cluster, though. Another way could be to resolve the tags - the sail-operator knows which revisions the tags point to, after all - but only if tags are exclusively managed by the sail-operator.
+
+Revision tags must never overlap between `Istio` resources, so we'll need to make sure during reconciliation that they are unique and report an error in the status if they are not.
```
sridhargaddam, line 32: Not sure i understand. Can you please eloborate with an example what this sentence means?

enhancements/SEP2-revision-tags.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Alternatives Considered
+### values.revisionTags
+Istio has a `values.revisionTags` field that we even currently expose in our APIs. The problem is that we copy all values from the `Istio` resource to every `IstioRevision` and that means we would be facing duplicate revision tags when we create additional revisions in the sail-operator - so, some logic would be required to work around this problem. As it is a similar amount of effort, I prefer the explicit version of adding the field to the `Istio` CRD.
+
+## Implementation Plan
+- [ ] Improve revision usage detection to support revision tags
+- [ ] Add field revisionTags to Istio CRD
+- [ ] Implement tag creation and update
+- [ ] Add logic to prevent duplicate tags
```
sridhargaddam, line 42: - [ ] documentation

sridhargaddam (commented):
Thanks @dgn for the SEP.

enhancements/SEP2-revision-tags.md
dgn: a revision tag can only point to one revision. So if a user has multiple `Istio` resources, they cannot set the same revision tags on them, otherwise the operator wouldn't know which revision to tag. I'll try to improve the wording

enhancements/SEP2-revision-tags.md
dgn: allowing the use of the `istio-injection=enabled` (I see now I made a mistake in the text, it's `istio-injection=enabled`, not `true`) is one of the purposes of this change (see user story 2) - that only works if you tag the revision with the `default` tag though

dgn: I added an explanation of this special case in the overview and linked it here

enhancements/SEP2-revision-tags.md
```diff
@@ -0,0 +1,45 @@
...
+|---------------------------------------------------|--------------|------------|
+|WIP                                                | @dgn         | 2024-07-17 |
+
+# Revision Tag Support
+
+## Overview
+Upstream Istio supports the use of revision tags for multi-revision deployments and canary upgrades of Istio control planes. These tags serve as aliases for revisions and allow users to use stable revision names (e.g. `prod` or `default`), so they don't have to change their namespace and pod labels (in this case `istio.io/rev=prod` or `istio-injection=enabled`) when switching to a new version. Instead, by tagging a new revision with the correct tag and restarting their workloads, they can perform an Istio update without having to change their labels. This is especially useful in situations where the team managing the Istio control plane is separate from the teams managing the workloads.
+
+Each revision tag only ever points to exactly one Istio revision. Upstream, revision tags are created manually using `istioctl` and- as they only affect injection- are represented in the cluster by a MutatingWebhookConfiguration.
+
```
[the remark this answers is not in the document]
dgn: done

enhancements/SEP2-revision-tags.md
```diff
@@ -0,0 +1,55 @@
...
+Revision tags can have any name, there is only one special case: revisions tagged `default` are treated as if they had an empty revision name, thereby allowing the use of the standard namespace injection label `istio-injection=enabled`.
+
+Each revision tag only ever points to exactly one Istio revision. Upstream, revision tags are created manually using `istioctl` and- as they only affect injection- are represented in the cluster by a MutatingWebhookConfiguration.
+
+Here's an example how to create a `default` revision tag that points to the `1-21-1` revision using `istioctl`:
+
+```bash
+istioctl tag set default --revision 1-21-1
```
sridhargaddam, line 17: I'm getting the following error when running this command in a default istio deployment.

```
$ istioctl tag set default --revision 1-21-1
Error: cannot modify tag: cannot find MutatingWebhookConfiguration with revision "1-21-1"
```

and the MutatingWebhookConfiguration is

```
$ k get MutatingWebhookConfiguration 
NAME                         WEBHOOKS   AGE
istio-revision-tag-default   4          83s
istio-sidecar-injector       4          94s
$
```

dgn: did you set your istio control plane's revision to `1-21-1`? that is required

sridhargaddam: Got it, I missed this step. Thanks for clarifying @dgn

sridhargaddam:
LGTM, but will wait if others have any review comments.

FilipB (approved):
LGTM

fjglira (approved):
lgtm

enhancements/SEP2-revision-tags.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Architecture
+We will need to update the sail-operators mechanism to detect revisions that are being used. Today, we only look at the `istio.io/rev` label's value to check which revisions are in use. But when revision tags are used, those values will be mere aliases, so we have to improve our detection mechanism. The most correct way is probably to look at the revision annotation on the pods that is set by Istio during injection. That requires inspecting every pod in the cluster, though. Another way could be to resolve the tags - the sail-operator knows which revisions the tags point to, after all - but only if tags are exclusively managed by the sail-operator.
+
+Revision tags must never overlap between `Istio` resources, as every revision tag must point to exactly one revision, and if multiple `Istio` resources define the same revision tags, they would flip-flop between the respective revisions, depending on which resource was reconciled last. We'll need to make sure during reconciliation that they are unique and report an error in the status if they are not.
```
luksa, line 40: I believe they wouldn't flip-flop. Instead, there'll be an error during the reconciliation of the 2nd Istio instance. Helm will throw an error when it encounters an existing MutatingWebhookConfiguration representing the tag, because it's associated with an existing Helm release.

enhancements/SEP2-revision-tags.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Design
+
+### User Stories
+1. As a user of sail-operator's RevisionBased update strategy, I want to be able to use the `istio-injection=enabled` label on my namespaces.
+1. As a platform engineer, I want my application teams to be able to use a fixed label for proxy injection without having to know which version of Istio is running in the cluster, so that I can perform upgrades in the background without the application teams having to make changes to use the new version.
+
+### API Changes
+We will add a new field `revisionTags` of type `[]string` to the `Istio` CRD. It will be located under `spec.updateStrategy.revisionTags`. Whenever the sail-operator deploys a new revision, it will update all the tags listed in the `revisionTags` field to point to this new revision. The default value for the field will be `{"default"}` - this way, the standard namespace injection label `istio-injection=enabled` will work out of the box for RevisionBased deployments (see second paragraph of the [Overview](#overview)).
```
luksa, line 33: If the operator copies the tags over to the `IstioRevision` resource, this means that when you bump the version in the `Istio` resource, the operator will now also need to re-render the old revision. This is a significant change compared to the current behavior, where a version bump triggers only the deployment of a new IstioRevision and the associated control plane. 

Not sure what the implications of updating the old revision are. Maybe it's not a big deal, since the only consequence should be the removal of the MutatingWebhookConfiguration, but we should think about this (and/or experiment with it).

enhancements/SEP2-revision-tags.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Architecture
+We will need to update the sail-operators mechanism to detect revisions that are being used. Today, we only look at the `istio.io/rev` label's value to check which revisions are in use. But when revision tags are used, those values will be mere aliases, so we have to improve our detection mechanism. The most correct way is probably to look at the revision annotation on the pods that is set by Istio during injection. That requires inspecting every pod in the cluster, though. Another way could be to resolve the tags - the sail-operator knows which revisions the tags point to, after all - but only if tags are exclusively managed by the sail-operator.
+
+Revision tags must never overlap between `Istio` resources, as every revision tag must point to exactly one revision, and if multiple `Istio` resources define the same revision tags, they would flip-flop between the respective revisions, depending on which resource was reconciled last. We'll need to make sure during reconciliation that they are unique and report an error in the status if they are not.
+
+## Alternatives Considered
```
luksa, line 42: What about introducing a new resource kind (e.g. `IstioRevisionTag`), which contains a reference to an `IstioRevision`? One benefit of this approach is that a tag can never point to more than one revision and there's an easy way to list all tags in the cluster. One downside is that users will have to create a default `IstioRevisionTag` to enable the use of the `istio-injection: enabled` label (this is in contrast to the current proposal, where `spec.revisionTags` defaults to `["default"]`).

enhancements/SEP2-revision-tags.md
```diff
@@ -0,0 +1,55 @@
...
+
+We will also need to remove the `values.revisionTags` field (which is how the upstream charts expose this feature). We can still set that field to trigger creation of the tags, but we should not expose it to users.
+
+### Architecture
+We will need to update the sail-operators mechanism to detect revisions that are being used. Today, we only look at the `istio.io/rev` label's value to check which revisions are in use. But when revision tags are used, those values will be mere aliases, so we have to improve our detection mechanism. The most correct way is probably to look at the revision annotation on the pods that is set by Istio during injection. That requires inspecting every pod in the cluster, though. Another way could be to resolve the tags - the sail-operator knows which revisions the tags point to, after all - but only if tags are exclusively managed by the sail-operator.
```
luksa, line 38: > We will need to update the sail-operators mechanism to detect revisions that are being used. Today, we only look at the `istio.io/rev` label's value to check which revisions are in use.
This is not the case. The operator checks both the [namespace labels](https://github.com/istio-ecosystem/sail-operator/blob/e3c7f48fc22c81fef63c57301e323679d3088a5a/controllers/istiorevision/istiorevision_controller.go#L438) and the [pod labels and annotations](https://github.com/istio-ecosystem/sail-operator/blob/e3c7f48fc22c81fef63c57301e323679d3088a5a/controllers/istiorevision/istiorevision_controller.go#L442). We have to check both to account for the corner case where a namespace references a revision, but has no pods at that moment (maybe they were all deleted at the same time).

Anyway, updating the existing code to account for tags should fairly straightforward.

enhancements/SEP2-revision-tags.md
luksa: After thinking more about the `IstioRevisionTag` approach, I realized that both revisions and tags are referenced via the same `istio.io/rev` label, which means that their names must be unique across both revisions and tags (i.e. you can't have a revision called `foo` and a tag called `foo` at the same time). 

So, to ensure uniqueness, another approach would be to introduce a new `IstioRevision` type called `Tag` (in addition to the existing `Local` and `Remote` types). This ensures that users can look up all the active revision/tag names in a single list (`kubectl get istiorevisions`). The downside of this approach is that a tag is much different than a revision, since it doesn't have `values`, `version`, or `namespace`, but does have a reference to another revision. So, maybe merging revisions and tags into a single resource isn't a good approach, even though it would ensure name uniqueness. But it probably should be listed as an alternative that we have considered.

Bottom line seems to be that we can't really ensure name uniqueness through resource names, bringing into question whether it even makes sense to introduce `IstioRevisionTag`, as proposed in my previous comment.

enhancements/SEP2-revision-tags.md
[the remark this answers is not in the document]
luksa: No, this can't be it, because the `stable` tag appears in two revisions. The rendering of the charts would fail.

But it does look like you've shown that this approach can't work. The user directly controls only the `Istio` resources. And during an upgrade, the user will probably want to use the `stable` tag on the existing revision, and initially assign the `canary` tag to the new revision. Not sure how this can be done via a single `revisionTags` field in the `Istio` resource, unless the `revisionTags` in `Istio` only applies to the new revision. But then, the user is forced to update `revisionTags` at the same time as the `version` field.

luksa:
Can someone confirm that there are two separate approaches to creating tags in upstream istio?

1. `istioctl tag set` (this creates the MutatingWebhookConfiguration directly)
2. deploying the istiod chart via Helm and setting `values.revisionTags` (this creates the MutatingWebhookConfiguration along with all the other resources in this chart)

In the current proposal, we'll be creating the webhook through this second mechanism. But this doesn't feel right. A tag is not really part of a control plane; it's a separate thing that points to one. The lifecycle of the tag is (and should be) completely separate from the control plane. If we use the `values.revisionTags` approach, then the webhook is tied to the lifecycle of the control plane and we need to update the control plane every time the user changes the tag (I already explained this in one of my other comments where I pointed out that we'd have to reconcile both the old and the new IstioRevision).

enhancements/SEP2-revision-tags.md
nrfox: > What about introducing a new resource kind (e.g. IstioRevisionTag), which contains a reference to an IstioRevision?

I like this idea. What is the lifecycle of the tag tied to? The `Istio` resource? Maybe we could have a setting on the `Istio` resource that will create a default `IstioRevisionTag` for you when enabled when the `Istio` resource gets created?

luksa: @nrfox `istioctl` doesn't allow you to create a tag with the same name as an existing revision:
```console
$ istioctl tag set main-rev --revision=main-rev
Error: cannot create revision tag "main-rev": found existing control plane revision with same name
```

nrfox: Ah I was wrong then. Well it may be worth adding this discussion to the SEP as an alternative considered regardless. 

Wouldn't having tags as a field on the `Istio` resource run into the same problem? You could add a tag to `istio.spec.updateStrategy.revisionTags` that matched an existing `IstioRevision`.

dgn: I really like the `IstioRevisionTag` resource approach. I think the name uniqueness is not a show-stopper: we could always have a status condition that shows an error if the name isn't unique. Your tag simply never gets created

dgn:
> Can someone confirm that there are two separate approaches to creating tags in upstream istio?
> 
>     1. `istioctl tag set` (this creates the MutatingWebhookConfiguration directly)
> 
>     2. deploying the istiod chart via Helm and setting `values.revisionTags` (this creates the MutatingWebhookConfiguration along with all the other resources in this chart)
> 
> 
> In the current proposal, we'll be creating the webhook through this second mechanism. But this doesn't feel right. A tag is not really part of a control plane; it's a separate thing that points to one. The lifecycle of the tag is (and should be) completely separate from the control plane. If we use the `values.revisionTags` approach, then the webhook is tied to the lifecycle of the control plane and we need to update the control plane every time the user changes the tag (I already explained this in one of my other comments where I pointed out that we'd have to reconcile both the old and the new IstioRevision).

`istioctl tag set` just performs a helm render and filters for the MutatingWebhookConfiguration and then only deploys that. I had similar behavior in mind (in fact, I attempted importing the existing function but it didn't quite work because the charts are named differently in our repo). I'm all for keeping lifecycles separate, so I think I'll go with your suggestion of a `IstioRevisionTag` resource and put that into writing. The name uniqueness I would try and solve using status (and showing an error there if it's not unique). Does that work for you?

enhancements/SEP2-revision-tags.md
```diff
@@ -0,0 +1,69 @@
...
+
+If the `spec.targetRef` is used to reference an `Istio` resource, the Sail Operator will automatically update the revision tag when a new `IstioRevision` is created as part of a version update of the `Istio` resource. In this case, the `IstioRevisionTag` resource behaves like a floating tag, always referencing the latest `IstioRevision` of an `Istio` resource.
+
+When the very first `IstioRevision` is created in a cluster from a `RevisionBased` Istio resource, the Sail Operator will create a `IstioRevisionTag` with the name `default`, referencing that `IstioRevision`. This way, the standard namespace injection label `istio-injection=enabled` will work out of the box for RevisionBased deployments (see second paragraph of the [Overview](#overview)).
```
nrfox, line 39: nit:
```suggestion
When the very first `IstioRevision` is created in a cluster from a `RevisionBased` Istio resource, the Sail Operator will create a `IstioRevisionTag` with the name `default`, referencing that `IstioRevision`. This way, the standard namespace injection label `istio-injection=enabled` will work out of the box for `RevisionBased` deployments (see second paragraph of the [Overview](#overview)).
```

nrfox: This more or less ties the lifecycle of the default tag to the first `Istio` resource that gets created in the cluster. Creating an `Istio`, deleting the `Istio`, creating a new `Istio` will mean the second `Istio` doesn't get the default tag?

nrfox: Is it feasible to have the Sail Operator check if if no `Istio` resource exists in the cluster then create the default tag? That way deleting all `Istio`s and creating a new one will still get you an `Istio` with a default tag.

enhancements/SEP2-revision-tags.md
```diff
@@ -0,0 +1,69 @@
...
+
+We will also need to remove the `values.revisionTags` field from our API, which is how the upstream charts expose this feature.
```
nrfox, line 41: How does the `istioctl` implementation differ from this implementation?

enhancements/SEP2-revision-tags.md
```diff
@@ -0,0 +1,69 @@
...
+
+### Architecture
+We will need to update the sail-operators mechanism to detect revisions that are being used. Today, we only look at the `istio.io/rev` label's value to check which revisions are in use. But when revision tags are used, those values will be mere aliases, so we have to improve our detection mechanism. The most correct way is probably to look at the revision annotation on the pods that is set by Istio during injection. That requires inspecting every pod in the cluster, though. Another way could be to resolve the tags - the sail-operator knows which revisions the tags point to, after all - but only if tags are exclusively managed by the sail-operator.
```
nrfox, line 44: nit:
```suggestion
We will need to update the Sail-Operator's mechanism to detect in-use revisions. Today, we only look at the `istio.io/rev` label's value to check which revisions are in use. But when revision tags are used, those values will be mere aliases, so we have to improve our detection mechanism. The most correct way is probably to look at the revision annotation on the pods that is set by Istio during injection. That requires inspecting every pod in the cluster, though. Another way could be to resolve the tags - the sail-operator knows which revisions the tags point to, after all - but only if tags are exclusively managed by the sail-operator.
```

nrfox: Need to cover the case where both `istio-injection` and `istio.io/rev` are provided as well.

enhancements/SEP2-revision-tags.md
```diff
@@ -0,0 +1,69 @@
...
+* Compatibility with manual revision tag creation using istioctl. There might be a way to disable the operator functionality to avoid conflicts when creating revision tags manually, but that's it - you either do it yourself or let the operator do it
+
+## Design
+
+### User Stories
+1. As a user of sail-operator's RevisionBased update strategy, I want to be able to use the `istio-injection=enabled` label on my namespaces.
+1. As a platform engineer, I want my application teams to be able to use a fixed label for proxy injection without having to know which version of Istio is running in the cluster, so that I can perform upgrades in the background without the application teams having to make changes to use the new version.
+
+### API Changes
+We will add a new CRD called `IstioRevisionTag` that includes a `spec.targetRef` field. In this field, users can specify the `IstioRevision` or `Istio` resource that the `IstioRevisionTag` references.
```
nrfox, line 33: Include full spec for the `IstioRevisionTag`? Moar YAML!

enhancements/SEP2-revision-tags.md
```diff
@@ -0,0 +1,69 @@
...
+
+Revision tags and revision names can be used interchangably, so they must never overlap. Therefore, we'll need a `status` on the `IstioRevisionTag` resource that can show the user an error if the name they chose is already taken by a `IstioRevision`. Another case we'll need to cover is when an `IstioRevision` is being reconciled and it would be assigned the same name as an existing tag. In this case, I suggest the `IstioRevision` wins and the tag is decommissioned, also with an error in its status. As the revision names generated for `IstioRevision` are very specific and end with the Istio control plane version, it is unlikely for this to happen, but we'll need to define the behavior for when it does.
```
nrfox, line 46: What does "decommissioned" mean exactly? The underlying mutating webhook is deleted but the `IstioRevisionTag` resource remains with an error in its status?

dgn: pretty much. I'll add that to the doc

enhancements/SEP2-revision-tags.md
dgn: internally, it sets that field, renders the chart and then removes anything but the webhook iirc

enhancements/SEP2-revision-tags.md
dgn: moved to separate ticket https://github.com/istio-ecosystem/sail-operator/issues/439

enhancements/SEP2-revision-tags.md
```diff
@@ -0,0 +1,69 @@
...
+
+Revision tags can have any name, there is only one special case: revisions tagged `default` are treated as if they had an empty revision name, thereby allowing the use of the standard namespace injection label `istio-injection=enabled`.
+
+Each revision tag only ever points to exactly one Istio revision. Upstream, revision tags are created manually using `istioctl` and- as they only affect injection- are represented in the cluster by a MutatingWebhookConfiguration.
```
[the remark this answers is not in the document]
dgn: I think in Primary-Remote setups you'd need to create them in the Remote cluster and I'm not sure how they work in that case tbh. It's a good question

Issue #182, opened 2024-06-27
Explain API concepts


dgn:
We'll need some free-form text under `docs/` that explains our API concepts and how our CRDs are supposed to be used

PR #225, opened 2024-07-19
docs: explain API concepts

dgn:
Fixes https://github.com/istio-ecosystem/sail-operator/issues/182

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -40,7 +40,7 @@ sail-operator manages the lifecycle of your Istio control planes and aims to bec
 ## Concepts
 
 ### Istio resource
-The `Istio` resource is used to manage your Istio control planes. You can access all helm chart options through the `values` field in the `spec`:
+The `Istio` resource is used to manage your Istio control planes. It is a cluster-wide resource, as the Istio control plane operates in and requires access to the entire cluster. To select a namespace to run the control plane pods in, you can use the `spec.namespace` field. You can access all helm chart options through the `values` field in the `spec`:
 
 ```yaml
 apiVersion: operator.istio.io/v1alpha1
@@ -67,12 +67,12 @@ sail-operator supports two different update strategies for your control planes: 
 After creation of an `Istio` resource, the sail-operator will generate a revision name for it based on the updateStrategy that was chosen, and create a corresponding [`IstioRevision`](#istiorevision-resource).
 
 ### IstioRevision resource
-The `IstioRevision` is the lowest-level API the sail-operator provides, and it is usually not created by the user, but by the operator itself. It's schema closely resembles that of the `Istio` resource - but instead of representing the state of a control plane you want to be present in your cluster, it represents a *revision* of that control plane, which is an important distinction especially when you're working with Istio's canary update feature.
+The `IstioRevision` is the lowest-level API the sail-operator provides, and it is usually not created by the user, but by the operator itself. It's schema closely resembles that of the `Istio` resource - but instead of representing the state of a control plane you want to be present in your cluster, it represents a *revision* of that control plane, which is an important distinction especially when you're working with Istio's canary update feature. Therefore, it is also a cluster-wide resource.
 
 You can think of the relationship between the `Istio` and `IstioRevision` resource as similar to the one between Kubernetes' `ReplicaSet` and `Pod`: a `ReplicaSet` can be created by users and results in the automatic creation of `Pods`, which will trigger the instanciation of your containers. Similarly, users create an `Istio` resource which instructs the operator to create a matching `IstioRevision`, which then in turn triggers the creation of the Istio control plane. To do that, the sail-operator will copy all of your relevant configuration from the `Istio` resource to the `IstioRevision` resource.
 
 ### IstioCNI resource
-The lifecycle of Istio's CNI plugin is managed separately when using sail-operator. To install it, you can create a `IstioCNI` resource. It also has a `values` field that exposes all of the options provided in the `istio-cni` chart:
+The lifecycle of Istio's CNI plugin is managed separately when using sail-operator. To install it, you can create a `IstioCNI` resource. The `IstioCNI` resource is a cluster-wide resource as it will install a `DaemonSet` that will be operating on all nodes of your cluster. You can select a version by setting the `spec.version` field, as you can see in the sample below. To update the CNI plugin, just change the `version` field to the version you want to install. Just like the `Istio` resource, it also has a `values` field that exposes all of the options provided in the `istio-cni` chart:
 
 ```yaml
 apiVersion: operator.istio.io/v1alpha1
--- a/docs/README.md
+++ b/docs/README.md
@@ -35,10 +35,59 @@
 
 
 # User Documentation
-tbd
+sail-operator manages the lifecycle of your Istio control planes and aims to become a vehicle for automated integration of Istio with compatible third-party software. Instead of creating a new configuration schema, sail-operator APIs are built around Istio's helm chart APIs. All installation and configuration options that are exposed by Istio's helm charts are available through the sail-operator CRDs' `values` fields.
 
 ## Concepts
-tbd
+
+### Istio resource
+The `Istio` resource is used to manage your Istio control planes. You can access all helm chart options through the `values` field in the `spec`:
+
+```yaml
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -1,10 +1,59 @@
...
+
+You can think of the relationship between the `Istio` and `IstioRevision` resource as similar to the one between Kubernetes' `ReplicaSet` and `Pod`: a `ReplicaSet` can be created by users and results in the automatic creation of `Pods`, which will trigger the instanciation of your containers. Similarly, users create an `Istio` resource which instructs the operator to create a matching `IstioRevision`, which then in turn triggers the creation of the Istio control plane. To do that, the sail-operator will copy all of your relevant configuration from the `Istio` resource to the `IstioRevision` resource.
+
+### IstioCNI resource
+The lifecycle of Istio's CNI plugin is managed separately when using sail-operator. To install it, you can create a `IstioCNI` resource. It also has a `values` field that exposes all of the options provided in the `istio-cni` chart:
```
sridhargaddam, line 41: Can you add a note on how upgrades work for IstioCNI.

docs/README.md
```diff
@@ -1,10 +1,59 @@
...
 
 # User Documentation
-tbd
+sail-operator manages the lifecycle of your Istio control planes and aims to become a vehicle for automated integration of Istio with compatible third-party software. Instead of creating a new configuration schema, sail-operator APIs are built around Istio's helm chart APIs. All installation and configuration options that are exposed by Istio's helm charts are available through the sail-operator CRDs' `values` fields.
 
 ## Concepts
-tbd
+
+### Istio resource
+The `Istio` resource is used to manage your Istio control planes. You can access all helm chart options through the `values` field in the `spec`:
```
FilipB, line 9: Maybe adding a note that it's a cluster-wide resource?

docs/README.md
dgn: done! PTAL

docs/README.md
dgn: done! I also added a note to the IstioCNI section

docs/README.md
```diff
@@ -1,10 +1,59 @@
 [Return to Project Root](../)
 
 # User Documentation
-tbd
+sail-operator manages the lifecycle of your Istio control planes and aims to become a vehicle for automated integration of Istio with compatible third-party software. Instead of creating a new configuration schema, sail-operator APIs are built around Istio's helm chart APIs. All installation and configuration options that are exposed by Istio's helm charts are available through the sail-operator CRDs' `values` fields.
```
nrfox, line 4: > aims to become a vehicle for automated integration of Istio with compatible third-party software

I'm not really sure what this means.

FilipB (approved):
LGTM

docs/README.md
```diff
@@ -1,10 +1,59 @@
...
+    pilot:
+      resources:
+        requests:
+          cpu: 100m
+          memory: 1024Mi
+```
+
+To support canary updates of the control plane, sail-operator includes support for multiple Istio versions. You can select a version by setting the `version` field in the `spec` to the version you would like to install, prefixed with a `v`. You can then update to a new version just by changing this field.
+
+sail-operator supports two different update strategies for your control planes: `InPlace` and `RevisionBased`. When using `InPlace`, the operator will immediately replace your existing control plane resources with the ones for the new version, whereas `RevisionBased` uses Istio's canary update mechanism by creating a second control plane to which you can migrate your workloads to complete the update.
```
nrfox, line 31: Might be worth mentioning that the operator can also migrate the workloads for you if you specify `updateWorkloads=true`.

docs/README.md
```diff
@@ -1,10 +1,59 @@
...
+
+### IstioRevision resource
+The `IstioRevision` is the lowest-level API the sail-operator provides, and it is usually not created by the user, but by the operator itself. It's schema closely resembles that of the `Istio` resource - but instead of representing the state of a control plane you want to be present in your cluster, it represents a *revision* of that control plane, which is an important distinction especially when you're working with Istio's canary update feature. Therefore, it is also a cluster-wide resource.
+
+You can think of the relationship between the `Istio` and `IstioRevision` resource as similar to the one between Kubernetes' `ReplicaSet` and `Pod`: a `ReplicaSet` can be created by users and results in the automatic creation of `Pods`, which will trigger the instanciation of your containers. Similarly, users create an `Istio` resource which instructs the operator to create a matching `IstioRevision`, which then in turn triggers the creation of the Istio control plane. To do that, the sail-operator will copy all of your relevant configuration from the `Istio` resource to the `IstioRevision` resource.
```
nrfox, line 38:
```suggestion
You can think of the relationship between the `Istio` and `IstioRevision` resource as similar to the one between Kubernetes' `ReplicaSet` and `Pod`: a `ReplicaSet` can be created by users and results in the automatic creation of `Pods`, which will trigger the instantiation of your containers. Similarly, users create an `Istio` resource which instructs the operator to create a matching `IstioRevision`, which then in turn triggers the creation of the Istio control plane. To do that, the sail-operator will copy all of your relevant configuration from the `Istio` resource to the `IstioRevision` resource.
```

docs/README.md
```diff
@@ -1,10 +1,59 @@
...
+
+sail-operator supports two different update strategies for your control planes: `InPlace` and `RevisionBased`. When using `InPlace`, the operator will immediately replace your existing control plane resources with the ones for the new version, whereas `RevisionBased` uses Istio's canary update mechanism by creating a second control plane to which you can migrate your workloads to complete the update.
+
+After creation of an `Istio` resource, the sail-operator will generate a revision name for it based on the updateStrategy that was chosen, and create a corresponding [`IstioRevision`](#istiorevision-resource).
+
+### IstioRevision resource
+The `IstioRevision` is the lowest-level API the sail-operator provides, and it is usually not created by the user, but by the operator itself. It's schema closely resembles that of the `Istio` resource - but instead of representing the state of a control plane you want to be present in your cluster, it represents a *revision* of that control plane, which is an important distinction especially when you're working with Istio's canary update feature. Therefore, it is also a cluster-wide resource.
```
nrfox, line 36: Maybe also define what a "revision" is. revision == instance of a controlplane? snapshot of a controlplane at a specific version?

docs/README.md
```diff
@@ -1,10 +1,59 @@
...
+
+### IstioCNI resource
+The lifecycle of Istio's CNI plugin is managed separately when using sail-operator. To install it, you can create a `IstioCNI` resource. The `IstioCNI` resource is a cluster-wide resource as it will install a `DaemonSet` that will be operating on all nodes of your cluster. You can select a version by setting the `spec.version` field, as you can see in the sample below. To update the CNI plugin, just change the `version` field to the version you want to install. Just like the `Istio` resource, it also has a `values` field that exposes all of the options provided in the `istio-cni` chart:
```
nrfox, line 41:
```suggestion
The lifecycle of Istio's CNI plugin is managed separately when using sail-operator. To install it, you can create an `IstioCNI` resource. The `IstioCNI` resource is a cluster-wide resource as it will install a `DaemonSet` that will be operating on all nodes of your cluster. You can select a version by setting the `spec.version` field, as you can see in the sample below. To update the CNI plugin, just change the `version` field to the version you want to install. Just like the `Istio` resource, it also has a `values` field that exposes all of the options provided in the `istio-cni` chart:
```

docs/README.md
sridhargaddam, line 4: In one of the other PRs, it was agreed that we should use the convention "Sail Operator" (or sail operator) instead of "sail-operator". Accordingly the Readme was updated. Hope we can use the same convention in this doc as well.

docs/README.md
dgn: That is not implemented though :)

docs/README.md
dgn: yeah I'm removing that. it's a bit too vague anyway and we don't actually do that right now

docs/README.md
```diff
@@ -1,10 +1,59 @@
 [Return to Project Root](../)
 
 # User Documentation
```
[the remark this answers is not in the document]
dgn: no - the operator should work on every k8s distro

docs/README.md
dgn: Oh, I didn't know - making the change

PR #229, opened 2024-07-23
Update Sail Operator cli instructions

sridhargaddam:
Some minor updates to the doc:

1. openshift-operators is usually present in any OCP deployment, so the instructions should create it only if its missing.
2. The instructions were only suggesting the yaml for `Subscription` but there were no instructions to apply it.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -69,31 +69,33 @@ is installed. `Succeeded` should appear in the **Status** column.
 1. Create the `openshift-operators` namespace (if it does not already exist).
 
     ```bash
-    $ kubectl create namespace openshift-operators
+    $ kubectl get ns openshift-operators || kubectl create ns openshift-operators
     ```
 
 1. Create the `Subscription` object with the desired `spec.channel`.
 
-    ```yaml
-    apiVersion: operators.coreos.com/v1alpha1
-    kind: Subscription
-    metadata:
-      name: sailoperator
-      namespace: openshift-operators
-    spec:
-      channel: "0.1-nightly"
-      installPlanApproval: Automatic
-      name: sailoperator
-      source: community-operators
-      sourceNamespace: openshift-marketplace
-    ```
+   ```bash
+   kubectl apply -f - <<EOF
+       apiVersion: operators.coreos.com/v1alpha1
+       kind: Subscription
+       metadata:
+         name: sailoperator
+         namespace: openshift-operators
+       spec:
+         channel: "3.0-nightly"
+         installPlanApproval: Automatic
+         name: sailoperator
+         source: community-operators
+         sourceNamespace: openshift-marketplace
+   EOF
+   ```
 
 1. Verify that the installation succeeded by inspecting the CSV status.
 
     ```bash
     $ kubectl get csv -n openshift-operators
     NAME                                     DISPLAY         VERSION                    REPLACES                                 PHASE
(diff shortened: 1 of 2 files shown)


docs/README.md
```diff
@@ -36,31 +36,33 @@ is installed. `Succeeded` should appear in the **Status** column.
 1. Create the `openshift-operators` namespace (if it does not already exist).
 
     ```bash
-    $ kubectl create namespace openshift-operators
+    $ kubectl get ns openshift-operators || kubectl create ns openshift-operators
```
luksa, line 39: Does it really make sense to use this long command instead of just creating the namespace? These docs are meant to be read and followed by users, so the shorter, the better. And if this command fails with "namespace already exists", the user will understand what's happening and will move on to the next step.

sridhargaddam: I understand your POV, but sometimes users just copy/paste these commands into the console or create small scripts with them. Personally, I think it's a good idea to include a command that works in both scenarios. However, I don't have strong opinions on this and am happy to remove the change if you prefer. Do let me know.

fjglira: +1 to not add a long command that can hide things to the user. In the step description is already saying: create the namespace if does not already exist.

fjglira (commented):
lgtm just did a comment in the discussion of the step 1. I prefer also to not have long commands and be a simple step by step guide for users

docs/README.md
luksa: @sridhargaddam Yeah, I see the benefits of both options. But I still think a shorter command is more useful for the majority of folks that will be reading this document. I assume that only a small number of readers will actually create scripts from this document, whereas a large majority will only read/scan through the document and possibly follow the steps manually; for those folks, shorter is better.

sridhargaddam: Thanks guys. I'll revert this change. 

Do you want me to make changes at the following places as well?
1. https://github.com/istio-ecosystem/sail-operator/blob/f0d810ee1c462bde6d022f809b1d92bda2c99187/README.md?plain=1#L106
2. https://github.com/istio-ecosystem/sail-operator/blob/f0d810ee1c462bde6d022f809b1d92bda2c99187/README.md?plain=1#L114
3. https://github.com/istio-ecosystem/sail-operator/blob/f0d810ee1c462bde6d022f809b1d92bda2c99187/README.md?plain=1#L122
4. https://github.com/istio-ecosystem/sail-operator/blob/f0d810ee1c462bde6d022f809b1d92bda2c99187/docs/README.md?plain=1#L402

fjglira: Yes, I think is better to do the change on those places also and add the comment: "Create namespace xyz if it does not already exist"

docs/README.md
```diff
@@ -74,26 +74,28 @@ is installed. `Succeeded` should appear in the **Status** column.
...
-    ```
+   ```bash
+   kubectl apply -f - <<EOF
+       apiVersion: operators.coreos.com/v1alpha1
+       kind: Subscription
+       metadata:
+         name: sailoperator
+         namespace: openshift-operators
+       spec:
+         channel: "3.0-nightly"
```
luksa, line 85: Is this OK? I noticed today that the last build in `3.0-nightly` is from May 13. The newer builds are in the `0.1-nightly` channel.

sridhargaddam: Ohh, i thought `3.0-nightly` would have the most latest ones. I'll update it.

luksa (approved):

PR #232, opened 2024-07-24
Fix flaky e2e error in reconciliation check

sridhargaddam:


luksa:
Hmm, no. This change makes the test useless, since it hides the fact that the operator is continuously reconciling the resource.

sridhargaddam:
> Hmm, no. This change makes the test useless, since it hides the fact that the operator is continuously reconciling the resource.

Just for record, we discussed this on slack and the next step is to identify why the [Reconcile](https://github.com/istio-ecosystem/sail-operator/blob/main/controllers/istiorevision/istiorevision_controller.go#L108) API is being invoked periodically (which is happening even after 30+ mins). Marking this PR as draft until we understand the core issue.

fjglira:
Can we close this PR? I think is already solved with changes from Marko

sridhargaddam:
Closing this as the original issue is fixed.

Issue #181, opened 2024-06-27
Generate API reference docs


dgn:

fjglira:
#237 is merged. Working now on fix and replace some comments and flags that make the reference documentation to have some errors. I'll submit a new PR tomorrow with this

PR #237, opened 2024-07-29
Add make target to generate api reference documentation

fjglira:
Adding a new make target to be used: `make gen-api-docs`. This uses https://github.com/elastic/crd-ref-docs project to generate from go code the api documentation references. We can use this make target in the generic make gen to update the documentation reference. Before (Or can be done after) merging this and adding the api reference documentation we need to discuss which fields we want to skip, if we want to skip files, and also fix some references that are displayed in the documentation wrong when we execute the target.

This is part of #181


docs/api-reference/operator.istio.io.md
```diff
@@ -0,0 +1,3242 @@
...
+```
+{{</tab>}}
+{{</tabset>}}
+
+
+The following rule configures a client to use TLS when talking to a
+foreign service whose domain matches *.foo.com.
+
+
+{{<tabset category-name="example">}}
```
fjglira, line 330: This brackets are shown in the page as brackets alone

docs/api-reference/operator.istio.io.md
```diff
@@ -0,0 +1,3242 @@
...
+| `hub` _string_ | Hub to pull the container image from. Image will be `Hub/Image:Tag-Variant`. |  |  |
+| `tag` _[IntOrString](https://kubernetes.io/docs/reference/generated/kubernetes-api/v/#intorstring-intstr-util)_ | The container image tag to pull. Image will be `Hub/Image:Tag-Variant`. |  | XIntOrString: \{\} <br /> |
+| `image` _string_ | Image name to pull from. Image will be `Hub/Image:Tag-Variant`.<br />If Image contains a "/", it will replace the entire `image` in the pod. |  |  |
+| `labelPods` _boolean_ | The Repair controller has 3 modes (labelPods, deletePods, and repairPods). Pick which one meets your use cases. Note only one may be used.<br />The mode defines the action the controller will take when a pod is detected as broken.<br />If labelPods is true, the controller will label all broken pods with <brokenPodLabelKey>=<brokenPodLabelValue>.<br />This is only capable of identifying broken pods; the user is responsible for fixing them (generally, by deleting them).<br />Note this gives the DaemonSet a relatively high privilege, as modifying pod metadata/status can have wider impacts. |  |  |
```
fjglira, line 174: The descriptions is not correct, is mentioning the description from reparPods

docs/api-reference/operator.istio.io.md
```diff
@@ -0,0 +1,3242 @@
...
+- [Istio](#istio)
+- [IstioCNI](#istiocni)
+- [IstioCNIList](#istiocnilist)
+- [IstioList](#istiolist)
+- [IstioRevision](#istiorevision)
+- [IstioRevisionList](#istiorevisionlist)
+
+
+
+#### ArchConfig
```
fjglira, line 21: @luksa This configuration came from the `values_type_gen.go` file.  And for some reason is being shown first. There is some ways to filter which fields we want to show, should be doing some filtering here?

docs/api-reference/operator.istio.io.md
fjglira: @luksa

fjglira: This will be easy to fix, we just need to update the values in the go files referenced. But we will need to check every field description to check that matches the correct description

fjglira:
We can fix those reference errors in a different PR.

docs/api-reference/operator.istio.io.md
fjglira: @luksa we copy this from here https://github.com/istio/istio/blob/master/operator/pkg/apis/istio/v1alpha1/values_types.pb.go? Upstream has the same issue with the description of the fields. I'll submit a PR to fix those descriptions upstream

fjglira: I just saw that the description has a lot of references that are the same but at the end have the exact definition for the field

fjglira:
We need to skip those  comments that have $hide_from_docs. This is not part from this documentation

docs/api-reference/operator.istio.io.md
luksa: The objects are listed in alphabetical order, that's why ArchConfig is first.

docs/api-reference/operator.istio.io.md
luksa: There are three fields (labelPods, deletePods, and repairPods), and only one of them can be set at a time. That's why all three field descriptions start with "The Repair controller has 3 modes...". MAybe this description could be improved, but this doesn't have anything to do with this PR.

docs/api-reference/operator.istio.io.md
fjglira: Do you know from where this information comes? Do we create this example or does it also come from some upstream file? Those example has a lot of errors when they are shown in md format

luksa: This `<tabset>` tag seems to be an upstream Istio documentation thing. Not sure what we can do about it. We can't change it in the upstream files, so maybe we need to transform them using our API transform tool.

luksa:
> We can fix those reference errors in a different PR.

Are you talking about the broken links to k8s types? I think we should fix them in this PR.

fjglira:
> > We can fix those reference errors in a different PR.
> 
> Are you talking about the broken links to k8s types? I think we should fix them in this PR.

I'm checking this and also others errors that came from `hide_from_docs` comments and examples

luksa:
> We need to skip those comments that have $hide_from_docs. This is not part from this documentation

This can be done in the API transformer tool. We just need to replace `$hide_from_docs` with `+hideFromDocs`

luksa:
Can you add a link to the API reference to the main README (or wherever it makes sense)?

fjglira:
> Can you add a link to the API reference to the main README (or wherever it makes sense)?

Done

luksa (approved):

PR #241, opened 2024-07-30
Hide some fields from the API reference docs

luksa:


docs/api-reference/operator.istio.io.md
fjglira, line 1: There is still some of the $hide_from_docs that are shown in the document. For example: https://github.com/istio-ecosystem/sail-operator/blob/b582b4fb171ba7f59608e1ddc25c37a86eb0bcd9/docs/api-reference/operator.istio.io.md#meshconfigservicesettings

fjglira (commented):
LGTM, but there is still some hide comments in the docs

docs/api-reference/operator.istio.io.md
fjglira: There are only 7  left

fjglira:
This PR is part of #181 , to fix the error in the generated documentation

fjglira (approved):

PR #242, opened 2024-07-30
API reference docs: replace tab tags for headings

### LINKED ISSUE #181: Generate API reference docs

fjglira:
Fixes #181 

Adding a new replace in the transform to replace the tab tags for headings because markdown does not handle properly those tags

### DIFF
--- a/hack/api_transformer/main.go
+++ b/hack/api_transformer/main.go
@@ -273,6 +273,7 @@ func (t *FileTransformer) processFile() (*ast.File, error) {
 	t.renameImports(file)
 	fixNames(file)
 	hideFromDocs(file)
+	replaceTabsWithHeadings(file)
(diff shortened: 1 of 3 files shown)


hack/api_transformer/main.go
```diff
@@ -349,6 +366,49 @@ func convertHideFromDocs(doc *ast.CommentGroup) {
...
+				newComments = append(newComments, &ast.Comment{Text: "//"})
+			} else {
+				newComments = append(newComments, comment)
+			}
+		} else {
+			newComments = append(newComments, comment)
+		}
+	}
+
+	doc.List = newComments
```
luksa, line 409: I think you can simplify this function a lot by replacing the individual items in `doc.List`  instead of building a new list and replacing the entire list.

luksa (approved):

Issue #216, opened 2024-07-18
Add relatedImages to CSV


dgn:
In order to support OLM-based container image mirroring like we do downstream, we should populate the relatedImages section of the CSV before pushing to OperatorHub. It's probably easiest to do that in the script that publishes the CSV.

PR #243, opened 2024-07-31
Add relatedImages to ClusterServiceVersion

bmangoen:
We want to add `relatedImages` to the Sail Operator `ClusterServiceVersion` ( cf. #216 )

### DIFF
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -483,6 +483,8 @@ $(GITLEAKS): $(LOCALBIN)
 bundle: gen helm operator-sdk ## Generate bundle manifests and metadata, then validate generated files.
 	$(HELM) template chart chart $(HELM_TEMPL_DEF_FLAGS) --set image='$(IMAGE)' --set platform=openshift --set bundleGeneration=true | $(OPERATOR_SDK) generate bundle $(BUNDLE_GEN_FLAGS)
 
+	@hack/patch-csv.sh bundle/manifests/$(OPERATOR_NAME).clusterserviceversion.yaml
+
 	# update CSV's spec.customresourcedefinitions.owned field. ideally we could do this straight in ./bundle, but
 	# sadly this is only possible if the file lives in a `bases` directory
 	mkdir -p _tmp/bases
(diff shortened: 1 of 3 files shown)


Makefile.core.mk
```diff
@@ -497,6 +497,8 @@ bundle: gen helm operator-sdk ## Generate bundle manifests and metadata, then va
 				git checkout "$$csvPath" || echo "failed to revert timestamp change. assuming we're in the middle of a merge"; \
 			fi \
 		fi
+
+	@hack/patch-csv.sh bundle/manifests/$(OPERATOR_NAME).clusterserviceversion.yaml
```
dgn, line 501: If you do this earlier (between `helm template` and `operator-sdk generate bundle`) you can probably avoid the reformatting

bmangoen: You're right! 
Even just between `helm template ... | operator-sdk generate bundle` and `operator-sdk generate kustomize manifests` (cf. [here](https://github.com/istio-ecosystem/sail-operator/blob/main/Makefile.core.mk#L486-L492)) should work

FilipB (changes requested):
I tried to run the script in our fork and it adds olm.relatedImage fields to already existing [images](https://github.com/openshift-service-mesh/sail-operator/blob/main/bundle/manifests/servicemeshoperator3.clusterserviceversion.yaml#L514) fields which is not correct. I'm not sure what is the best solution but we need to make sure this change will not break downstream builds

dgn:
> I tried to run the script in our fork and it adds olm.relatedImage fields to already existing [images](https://github.com/openshift-service-mesh/sail-operator/blob/main/bundle/manifests/servicemeshoperator3.clusterserviceversion.yaml#L514) fields which is not correct. I'm not sure what is the best solution but we need to make sure this change will not break downstream builds

maybe we can add an env var to control whether the related images are added? then we can disable it in the fork

bmangoen:
@FilipB @dgn `make -e GENERATE_RELATED_IMAGES=false gen` should not generate the annotations `olm.relatedImage.<version>.<component_name>` and the `relatedImages` part

dgn:
> @FilipB @dgn `make -e GENERATE_RELATED_IMAGES=false gen` should not generate the annotations `olm.relatedImage.<version>.<component_name>` and the `relatedImages` part

The annotations should stay - but it looks to me like they will?

dgn (approved):

FilipB (approved):

FilipB:
Added https://github.com/openshift-service-mesh/sail-operator/pull/112 to disable this in our fork.

PR #244, opened 2024-08-01
Use reflect.New instead of DeepCopy in StandardReconciler

luksa:


pkg/reconciler/reconciler.go
```diff
@@ -69,7 +68,7 @@ func NewStandardReconcilerWithFinalizer[T client.Object](
 func (r *StandardReconciler[T]) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	log := logf.FromContext(ctx)
 
-	obj := r.object.DeepCopyObject().(T) // TODO: create object using scheme.New() instead?
+	obj := reflect.New(reflect.TypeOf(*new(T)).Elem()).Interface().(T)
```
nrfox, line 71: Would simply `var obj T` here work?

luksa: Nope, because T is typically a pointer to something (e.g. `*IstioRevision`), so `obj` would be `nil`.

luksa: Yup.

nrfox (approved):

PR #253, opened 2024-08-08
Issue 251: Fix tabs in yaml example for the api reference docs

### LINKED ISSUE #251: [BUG] Format errors in yaml examples in the api reference doc

### Bug Description

in the yaml examples inside the API reference documentation, there is format errors because there are tabs instead of spaces in the yaml file. Example: https://github.com/istio-ecosystem/sail-operator/blob/main/docs/api-reference/operator.istio.io.md#v1alpha3

### Operator Version

master/tp-01

### Link to Gist with Logs

_No response_

### Additional Information

_No response_

fjglira:
Fixes #251 

Adding two simple replace after the generation of the api reference docs to clean the visual errors produced by the `github.com/elastic/crd-ref-docs` package. I did not add those replace in the `api transformer` because making the changes before to convert the api reference docs change also the way that the descriptions are displayed later in `charts/crd` and `bundle/manifest` files and adding the replace here allows to later leave again the tabs and br tags to generate html docs


Makefile.core.mk
```diff
@@ -398,6 +398,10 @@ gen-api-docs: ## Generate API documentation.
 		--output-path=$(OUTPUT_DOCS_PATH) \
 		--output-mode=group
 	@echo "API reference documentation generated at $(OUTPUT_DOCS_PATH)"
+	## The package used to convert CRDs to markdown files generates tabs in the output markdown files and br.
+	## This command replaces tabs with two spaces and removes br tags.
+	@find $(OUTPUT_DOCS_PATH) -type f -name "*.md" -exec sed -i 's/\t/  /g' {} \;
+	@find $(OUTPUT_DOCS_PATH) -type f -name "*.md" -exec sed -i 's/<br>/\n/g;s/<br \/>/\n/g' {} \;
```
luksa, line 404: Is this the right place to fix the formatting, given that you're only fixing the reference docs and not, for example, the `kubectl explain` output?

luksa: See description field when running `kubectl explain istio.spec.values.meshConfig.meshMTLS`. The example has tabs instead of spaces.

fjglira: I'll check to do the change in the transform API to see if we can fix this in all the places then

hack/api_transformer/main.go
```diff
@@ -328,15 +328,19 @@ func processDocs(file *ast.File) {
 		case *ast.GenDecl:
 			convertHideFromDocs(x.Doc)
 			convertTabsToHeadings(x.Doc)
+			cleanComments(x.Doc)
```
luksa, line 331: We can now introduce a new function that calls all these three functions, so we don't have to repeat calling all three for every `case` in this `switch`.

hack/api_transformer/main.go
```diff
@@ -381,6 +385,16 @@ func convertTabsToHeadings(doc *ast.CommentGroup) {
 	}
 }
 
+func cleanComments(doc *ast.CommentGroup) {
+	if doc == nil {
+		return
+	}
+	// Replace tabs with spaces
```
luksa, line 392: Can you also explain why?

fjglira:
@luksa I still need to add the deletion of the empty lines inside the yaml files (Need to check from where it comes) and for some reason there is a still some tabs that are not being replaced (Need to check also Why they are not being replaced)

fjglira:
As discussed with @luksa the changes are in the make target `make gen-api-docs` because we can't for now properly handle tabs replacement and new line deletion in the ast nodes because `go fmt` replaces the spaces by tabs by default and add those newlines. I'll check this behavior if is a bug or there is a way to avoid it

fjglira:
Closing in favor of https://github.com/istio-ecosystem/sail-operator/pull/262

PR #254, opened 2024-08-08
Fix issues identified during API review

luksa:


nrfox (commented):
LGTM all issues I found are fixed

chart/crds/operator.istio.io_istiocnis.yaml
```diff
@@ -1039,6 +1039,10 @@ spec:
                           enabled:
                             description: Controls whether ambient redirection is enabled
                             type: boolean
+                          ipv6:
```
dgn, line 1042: hmm. as we don't support ambient yet, maybe we want to remove this for now?

dgn (approved):
lgtm except for one question. Adding hold in case you want to make further changes, feel free to remove

/hold

dgn:
just noticed that this still contains the `autoscalingv2` field: https://github.com/istio-ecosystem/sail-operator/blob/ac99ae0c4612444ae7feaaea7c73ff31cb2d3bff/api/v1alpha1/values_types.gen.go#L410-L412

I can submit a patch to remove it upstream and then update the commit we use in a separate PR

chart/crds/operator.istio.io_istiocnis.yaml
luksa: Removed. Thanks.

luksa:
> just noticed that this still contains the `autoscalingv2` field:
> I can submit a patch to remove it upstream and then update the commit we use in a separate PR

Go ahead and remove it from upstream. But you don't need to update the commit, because I'm working on making the API transformer use the commits in go.mod, so your changes will automatically be pulled in when istio is updated.

PR #255, opened 2024-08-08
Disabling broken shellcheck

FilipB:
It's already done in the main branch


FilipB:
Not cherrypicking as the change in the main branch was part of a big commit

asmigala (approved):

PR #258, opened 2024-08-09
Generate API from versions defined in go.mod

luksa:


dgn (approved):
This is great, thank you Marko

PR #262, opened 2024-08-09
Fix errors in api reference doc

### LINKED ISSUE #251: [BUG] Format errors in yaml examples in the api reference doc

fjglira:
Fixes #251 

As discussed with @luksa the changes are in the make target `make gen-api-docs` because we can't for now properly handle tabs replacement and new line deletion in the `ast nodes` because `go fmt` replaces the spaces by tabs by default and add those newlines. I'll check this behavior to see if it is a bug or if there is a way to avoid it

### DIFF
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -385,6 +385,15 @@ gen-api-docs: ## Generate API documentation.
 		--renderer=markdown \
 		--output-path=$(OUTPUT_DOCS_PATH) \
 		--output-mode=group
+	@echo "The following replaces are applied to the generated markdown files as workaround"
+	@echo "because can't be applied in the ast nodes in the transform api"
+	@echo "go fmt does not properly handle tabs"
+	@echo "Remove br tags from the generated markdown files"
+	@find $(OUTPUT_DOCS_PATH) -type f -name "*.md" -exec sed -i 's/<br \/>/ /g' {} \;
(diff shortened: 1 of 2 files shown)


Makefile.core.mk
```diff
@@ -385,6 +385,15 @@ gen-api-docs: ## Generate API documentation.
 		--renderer=markdown \
 		--output-path=$(OUTPUT_DOCS_PATH) \
 		--output-mode=group
+	@echo "The following replaces are applied to the generated markdown files as workaround"
```
dgn, line 388: I would just add this and the following two lines as comments in the Makefile. It doesn't have to be shown every time we generate the docs IMO

dgn (approved):
LGTM, thanks @fjglira!

Adding a hold in case you want turn the explanation into a comment in the Makefile, feel free to remove.

/hold

fjglira:
> LGTM, thanks @fjglira!
> 
> Adding a hold in case you want turn the explanation into a comment in the Makefile, feel free to remove.
> 
> /hold

Updated :)

PR #291, opened 2024-08-16
Fix proxy related image

bmangoen:


dgn (approved):

hack/patch-csv.sh
```diff
@@ -52,22 +52,23 @@ function get_field() {
...
 
-  # Set if non null order .defaults.<field> 
-  #                  then .defaults.<component>.<field> 
-  #                  then .defaults.global.<field> 
-  #               finally .defaults.global.<component>.<field>
+  # Set if non null order from the component most specific to the most generic
+  # 1) .defaults.global.<component>.<field>
+  # 2) .defaults.<component>.<field> 
+  # 3) .defaults.global.<field> 
+  # 4) .defaults.<field>
```
asmigala, line 59: Shuold globals really be before non-globals? Isn't the non-global one more specific?

dgn: good point. it doesn't matter much as we're not looking at user-supplied values. but let's get the order straight anyway

PR #297, opened 2024-08-16
Issue 230: adding bookinfo deploy to e2e framework

### LINKED ISSUE #230: e2e: add deploy of bookinfo to e2e test
We need to add e2e validation to deploy bookinfo and at least to do basic check over the app deployed

fjglira:
Fixes #230 

Adding bookinfo deploy to validate proxy injection and proxy Istio version are successful. 

Tested on both kind and ocp

### DIFF
--- a/tests/e2e/controlplane/control_plane_test.go
+++ b/tests/e2e/controlplane/control_plane_test.go
@@ -18,8 +18,6 @@ package controlplane
 
 import (
 	"fmt"
-	"io"
-	"net/http"
 	"path/filepath"
 	"regexp"
 	"strings"
@@ -239,7 +237,7 @@ spec:
 						Success("Istio Operator stopped reconciling")
 					})
 
-					It("can be deployed bookinfo with sidecar inyection", func(ctx SpecContext) {
+					It("can be deployed bookinfo with sidecar injection", func(ctx SpecContext) {
 						By("Applying bookinfo YAML")
 						Expect(deployBookinfo(version)).To(Succeed(), "Error deploying bookinfo")
 						Success("Bookinfo deployed")
@@ -393,7 +391,7 @@ func forceDeleteIstioResources() error {
 	return nil
 }
 
-func getBookinfoYAML(version supportedversion.VersionInfo) (string, error) {
+func getBookinfoURL(version supportedversion.VersionInfo) string {
 	// Bookinfo YAML for the current version can be found from istio/istio repository
 	// If the version is latest, we need to get the latest version from the master branch
 	bookinfoURL := fmt.Sprintf("https://raw.githubusercontent.com/istio/istio/%s/samples/bookinfo/platform/kube/bookinfo.yaml", version.Version)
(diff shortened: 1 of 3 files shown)


tests/e2e/controlplane/control_plane_test.go
```diff
@@ -232,6 +238,37 @@ spec:
 							ShouldNot(ContainSubstring("Reconciliation done"), "Istio Operator is continuously reconciling")
 						Success("Istio Operator stopped reconciling")
 					})
+
+					It("can be deployed bookinfo with sidecar inyection", func(ctx SpecContext) {
```
asmigala, line 242: typo: `injection`, but also maybe change to something like `It("injects sidecars into deployed bookinfo")` ?

tests/e2e/controlplane/control_plane_test.go
```diff
@@ -232,6 +238,37 @@ spec:
...
+							Eventually(common.GetObject).WithArguments(ctx, cl, kube.Key(pod.Name, bookinfoNamespace), &corev1.Pod{}).
+								Should(HaveCondition(corev1.PodReady, metav1.ConditionTrue), "Pod is not ready")
+						}
+						Success("Bookinfo pods are ready")
+
+						By("Checking the sidecar injection version match the expected version")
+						for _, pod := range bookinfoPods.Items {
+							sidecarVersion, err := getProxyVersion(pod.Name, bookinfoNamespace)
+							Expect(err).To(Succeed(), "Error getting sidecar version")
+							Expect(sidecarVersion).To(ContainSubstring(version.Version), "Sidecar injection version does not match the expected version")
```
asmigala, line 262: Is checking for substring enough here?

tests/e2e/controlplane/control_plane_test.go
```diff
@@ -355,3 +392,47 @@ func forceDeleteIstioResources() error {
 
 	return nil
 }
+
+func getBookinfoYAML(version supportedversion.VersionInfo) (string, error) {
```
asmigala, line 396: Since our `kubectl` is just a thin wrapper around the real `kubectl` command, can't we just pass the url directly, as in `kubectl apply -f https://raw.githubusercontent.com/istio/istio/master/samples/bookinfo/platform/kube/bookinfo.yaml` ? That way we don't need to mess with http reponses and such

asmigala (changes requested):

tests/e2e/controlplane/control_plane_test.go
fjglira: There is an issue here, because for example for latest the version that we will get from the sidecar will be something like `1.24-alpha-xxxxxxx`, because of that I'm checking the substring. For released version will work with an `equal` but for latest will not work

tests/e2e/controlplane/control_plane_test.go
fjglira: Fixed

tests/e2e/controlplane/control_plane_test.go
```diff
@@ -232,6 +236,37 @@ spec:
 							ShouldNot(ContainSubstring("Reconciliation done"), "Istio Operator is continuously reconciling")
 						Success("Istio Operator stopped reconciling")
 					})
+
+					It("can be deployed bookinfo with sidecar injection", func(ctx SpecContext) {
```
asmigala, line 240: Hm, I still don't think "it can be deployed bookinfo with sidecar injection" is a proper english sentence.

Also, shouldn't this whole test be in its own separate `When("bookinfo is deployed) ...` ?

fjglira: Yeah, make sense to be in a different node. Let me check and do some changes

fjglira: Updated

asmigala (approved):
lgtm

PR #300, opened 2024-08-19
Don't watch Secrets and EnvoyFilters in istiorevision controller

luksa:
Previously, these two resource kinds had to be watched, because they were present in the charts. But, since they were later removed, the controller should no longer watch them.


dgn (approved):

dgn:
I wonder if there's a need for us to catch these things automatically

luksa:
> I wonder if there's a need for us to catch these things automatically

Yup, my thoughts exactly. Already working on it.

PR #304, opened 2024-08-19
[release-0.1] Removing latest and 1.21 versions

FilipB:
Not sure if we want to keep 1.21.z version on not


dgn:
Strange- where do the API changes come from?

FilipB:
> Strange- where do the API changes come from?
Isn't that expected when removing the 'latest' version?

dgn:
nevermind- I was looking at the changes in CRDs and thought those would be in our API. that's expected

PR #310, opened 2024-08-20
Add `operator-chart` target

dgn:
This makes sure our helm chart always points to the correct image.


Makefile.core.mk
```diff
@@ -403,6 +403,10 @@ endif
 operator-name:
 	sed -i "s/\(projectName:\).*/\1 ${OPERATOR_NAME}/g" PROJECT
 
+.PHONY: operator-chart
+operator-chart:
+	sed -i -e "s/\(image: .*:\).*/\1${TAG}/" chart/values.yaml
```
FilipB, line 408: Would it be safer to overwrite whole IMAGE and not just TAG? This will not work if HUB, IMAGE_BASE are changed.

Makefile.core.mk
```diff
@@ -403,6 +403,10 @@ endif
 operator-name:
 	sed -i "s/\(projectName:\).*/\1 ${OPERATOR_NAME}/g" PROJECT
 
+.PHONY: operator-chart
+operator-chart:
```
FilipB, line 407: Do we want to also overwrite version in chart/values.yaml? And maybe even version and appVersion in chart/Chart.yaml?

Makefile.core.mk
dgn: good point. done!

Makefile.core.mk
dgn: also done!

Makefile.core.mk
```diff
@@ -403,6 +403,12 @@ endif
...
+	sed -i -e "s/^\(version: \).*$$/\1${VERSION}/g" \
+	       -e "s/^\(appVersion: \).*$$/\1\"${VERSION}\"/g" chart/Chart.yaml
+	sed -i -e "s|^\(image: \).*$$|\1${IMAGE}|g" chart/values.yaml
```
FilipB, line 410:
```suggestion
	sed -i -e "s|^\(image: \).*$$|\1${IMAGE}|g" \
	       -e "s/\(version: \).*$$/\1${VERSION}/g" chart/values.yam
```

FilipB (approved):
Thanks

PR #313, opened 2024-08-21
Issue 296: Adding OLM deploy for operator e2e

### LINKED ISSUE #296: Install the operator via olm for e2e test
We need to add the installation of the operator via OLM to be able to validate that the images described in the [csv](https://github.com/istio-ecosystem/sail-operator/blob/main/bundle/manifests/sailoperator.clusterserviceversion.yaml#L604) are correct

fjglira:
Fixes #296 

Adding:

- `--skip-tls` to `deploy-olm` target because we can run this target pushing the images to an insecure registry like kind-registry
- `HELM_PLATFORM ?= openshift` to `bundle` make target because we had hardcoded openshift, and we can install the operator via OLM on the Kubernetes cluster also if OLM is available in the cluster
- Adding OLM operator deployment to run the e2e test against the operator installed using OLM

### DIFF
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -483,17 +483,15 @@ gitleaks: $(GITLEAKS) ## Download gitleaks to bin directory.
 $(GITLEAKS): $(LOCALBIN)
 	@test -s $(LOCALBIN)/gitleaks || GOBIN=$(LOCALBIN) go install github.com/zricethezav/gitleaks/v8@${GITLEAKS_VERSION}
 
-# HELM_PLATFORM is the platform to use when rendering the Helm chart
-# If platform is different to openshift, the --set platform=<platform> flag is not added to the helm template command
-HELM_PLATFORM ?= openshift
-
-# Only add the platform flag if HELM_PLATFORM is set to "openshift"
-ifeq ($(HELM_PLATFORM), openshift)
-	HELM_TEMPL_DEF_FLAGS := $(HELM_TEMPL_DEF_FLAGS) --set platform=$(HELM_PLATFORM)
-endif
+# Openshift Platform flag
+# If is set to true will add `--set platform=openshift` to the helm template command
+OPENSHIFT_PLATFORM ?= true
 
 .PHONY: bundle
 bundle: gen-all-except-bundle helm operator-sdk ## Generate bundle manifests and metadata, then validate generated files.
+ifeq ($(OPENSHIFT_PLATFORM), true)
+	HELM_TEMPL_DEF_FLAGS := $(HELM_TEMPL_DEF_FLAGS) --set platform=openshift
+endif
 	$(HELM) template chart chart $(HELM_TEMPL_DEF_FLAGS) --set image='$(IMAGE)' --set bundleGeneration=true | $(OPERATOR_SDK) generate bundle $(BUNDLE_GEN_FLAGS)
 ifeq ($(GENERATE_RELATED_IMAGES), true)
 	@hack/patch-csv.sh bundle/manifests/$(OPERATOR_NAME).clusterserviceversion.yaml
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -483,11 +483,18 @@ gitleaks: $(GITLEAKS) ## Download gitleaks to bin directory.
 $(GITLEAKS): $(LOCALBIN)
 	@test -s $(LOCALBIN)/gitleaks || GOBIN=$(LOCALBIN) go install github.com/zricethezav/gitleaks/v8@${GITLEAKS_VERSION}
 
+# HELM_PLATFORM is the platform to use when rendering the Helm chart
+# If platform is different to openshift, the --set platform=<platform> flag is not added to the helm template command
 HELM_PLATFORM ?= openshift
+
+# Only add the platform flag if HELM_PLATFORM is set to "openshift"
+ifeq ($(HELM_PLATFORM), openshift)
+	HELM_TEMPL_DEF_FLAGS := $(HELM_TEMPL_DEF_FLAGS) --set platform=$(HELM_PLATFORM)
+endif
+
 .PHONY: bundle
 bundle: gen-all-except-bundle helm operator-sdk ## Generate bundle manifests and metadata, then validate generated files.
-	$(HELM) template chart chart $(HELM_TEMPL_DEF_FLAGS) --set image='$(IMAGE)' --set platform=$(HELM_PLATFORM) --set bundleGeneration=true | $(OPERATOR_SDK) generate bundle $(BUNDLE_GEN_FLAGS)
-
+	$(HELM) template chart chart $(HELM_TEMPL_DEF_FLAGS) --set image='$(IMAGE)' --set bundleGeneration=true | $(OPERATOR_SDK) generate bundle $(BUNDLE_GEN_FLAGS)
 ifeq ($(GENERATE_RELATED_IMAGES), true)
 	@hack/patch-csv.sh bundle/manifests/$(OPERATOR_NAME).clusterserviceversion.yaml
 endif
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -299,7 +299,7 @@ deploy-yaml-openshift: verify-kubeconfig helm ## Output YAML manifests used by `
 .PHONY: deploy-olm
 deploy-olm: verify-kubeconfig bundle bundle-build bundle-push ## Build and push the operator OLM bundle and deploy the operator using OLM.
 	kubectl create ns ${NAMESPACE} || echo "namespace ${NAMESPACE} already exists"
-	$(OPERATOR_SDK) run bundle $(BUNDLE_IMG) -n ${NAMESPACE}
+	$(OPERATOR_SDK) run bundle $(BUNDLE_IMG) -n ${NAMESPACE} --skip-tls
 
 .PHONY: undeploy
 undeploy: verify-kubeconfig ## Undeploy controller from an existing cluster.
@@ -483,9 +483,10 @@ gitleaks: $(GITLEAKS) ## Download gitleaks to bin directory.
 $(GITLEAKS): $(LOCALBIN)
 	@test -s $(LOCALBIN)/gitleaks || GOBIN=$(LOCALBIN) go install github.com/zricethezav/gitleaks/v8@${GITLEAKS_VERSION}
 
+HELM_PLATFORM ?= openshift
 .PHONY: bundle
 bundle: gen-all-except-bundle helm operator-sdk ## Generate bundle manifests and metadata, then validate generated files.
-	$(HELM) template chart chart $(HELM_TEMPL_DEF_FLAGS) --set image='$(IMAGE)' --set platform=openshift --set bundleGeneration=true | $(OPERATOR_SDK) generate bundle $(BUNDLE_GEN_FLAGS)
+	$(HELM) template chart chart $(HELM_TEMPL_DEF_FLAGS) --set image='$(IMAGE)' --set platform=$(HELM_PLATFORM) --set bundleGeneration=true | $(OPERATOR_SDK) generate bundle $(BUNDLE_GEN_FLAGS)
 
 ifeq ($(GENERATE_RELATED_IMAGES), true)
 	@hack/patch-csv.sh bundle/manifests/$(OPERATOR_NAME).clusterserviceversion.yaml
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -489,10 +489,12 @@ OPENSHIFT_PLATFORM ?= true
 
 .PHONY: bundle
 bundle: gen-all-except-bundle helm operator-sdk ## Generate bundle manifests and metadata, then validate generated files.
-ifeq ($(OPENSHIFT_PLATFORM), true)
-	HELM_TEMPL_DEF_FLAGS := $(HELM_TEMPL_DEF_FLAGS) --set platform=openshift
-endif
-	$(HELM) template chart chart $(HELM_TEMPL_DEF_FLAGS) --set image='$(IMAGE)' --set bundleGeneration=true | $(OPERATOR_SDK) generate bundle $(BUNDLE_GEN_FLAGS)
+	@TEMPL_FLAGS="$(HELM_TEMPL_DEF_FLAGS)"; \
+	if [ "$(OPENSHIFT_PLATFORM)" = "true" ]; then \
+		TEMPL_FLAGS="$$TEMPL_FLAGS --set platform=openshift"; \
+	fi; \
+	$(HELM) template chart chart $$TEMPL_FLAGS --set image='$(IMAGE)' --set bundleGeneration=true | $(OPERATOR_SDK) generate bundle $(BUNDLE_GEN_FLAGS)
+
 ifeq ($(GENERATE_RELATED_IMAGES), true)
 	@hack/patch-csv.sh bundle/manifests/$(OPERATOR_NAME).clusterserviceversion.yaml
 endif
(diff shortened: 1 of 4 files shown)


fjglira:
@FilipB Do we need to add any additional validation after the operator is being deployed? Or is enough by validating with the current e2e coverage?

fjglira:
Adding the CI job on this [PR](https://github.com/istio/test-infra/pull/5474) (needs to be merged after this PR to have available the job)

tests/e2e/common-operator-integ-suite.sh
```diff
@@ -191,18 +214,50 @@ if [ "${SKIP_BUILD}" == "false" ]; then
     get_internal_registry
   fi
 
-  # BUILD AND PUSH IMAGE
+  # BUILD AND PUSH OPERATOR IMAGE
```
sridhargaddam, line 217: We can probably rename this function itself as `build_and_push_operator_image` and remove the comment.

tests/e2e/common-operator-integ-suite.sh
```diff
@@ -191,18 +214,50 @@ if [ "${SKIP_BUILD}" == "false" ]; then
...
-  if [ "${OCP}" == "true" ]; then
-    # This is a workaround
-    # To avoid errors of certificates meanwhile we are pulling the operator image from the internal registry
-    # We need to set image $HUB to a fixed known value after the push
-    # This value always will be equal to the svc url of the internal registry
-    HUB="image-registry.openshift-image-registry.svc:5000/sail-operator"
+  # If OLM is enabled, deploy the operator using OLM
+  if [ "${OLM}" == "true" ]; then
+    # Set the platform to Kubernetes by default.
+    # We are skipping the deploy via OLM test on OCP because the workaround to avoid the certificate issue is not working.
```
sridhargaddam, line 223: If we have an issue/Jira to track it, we can include it here.

tests/e2e/common-operator-integ-suite.sh
```diff
@@ -87,6 +108,8 @@ initialize_variables() {
   COMMAND="kubectl"
   ARTIFACTS="${ARTIFACTS:-$(mktemp -d)}"
   KUBECONFIG="${KUBECONFIG:-"${ARTIFACTS}/config"}"
+  LOCALBIN="${LOCALBIN:-"${HOME}/.local/bin"}"
```
sridhargaddam, line 111: In [Makefile.core.mk](https://github.com/istio-ecosystem/sail-operator/blob/c5485e96934a9873fe57844f558d847f980aca29/Makefile.core.mk#L432) file, we use operator-sdk from the $PWD/bin, can we re-use the same path (as defaults)?

tests/e2e/common-operator-integ-suite.sh
```diff
@@ -191,18 +214,50 @@ if [ "${SKIP_BUILD}" == "false" ]; then
...
+    IMAGE="${HUB}/${IMAGE_BASE}:${TAG}" \
+    IMAGE_TAG_BASE="${IMAGE_TAG_BASE}" \
+    BUNDLE_IMG="${BUNDLE_IMG}" \
+    HELM_PLATFORM="${HELM_PLATFORM}" \
+    make bundle bundle-build bundle-push
+
+    # Create operator namespace
+    ${COMMAND} create ns "${NAMESPACE}" || echo "namespace ${NAMESPACE} already exists"
+    # Deploy the operator using OLM
+	  ${OPERATOR_SDK} run bundle "${BUNDLE_IMG}" -n "${NAMESPACE}" --skip-tls
```
sridhargaddam, line 243:
```suggestion
${OPERATOR_SDK} run bundle "${BUNDLE_IMG}" -n "${NAMESPACE}" --skip-tls
```

tests/e2e/common-operator-integ-suite.sh
```diff
@@ -191,18 +214,50 @@ if [ "${SKIP_BUILD}" == "false" ]; then
...
 
-  if [ "${OCP}" == "true" ]; then
-    # This is a workaround
-    # To avoid errors of certificates meanwhile we are pulling the operator image from the internal registry
-    # We need to set image $HUB to a fixed known value after the push
-    # This value always will be equal to the svc url of the internal registry
-    HUB="image-registry.openshift-image-registry.svc:5000/sail-operator"
+  # If OLM is enabled, deploy the operator using OLM
+  if [ "${OLM}" == "true" ]; then
+    # Set the platform to Kubernetes by default.
```
sridhargaddam, line 222: Should we check the SKIP_DEPLOY flag here and skip deploying the Sail Operator if its set to true?

tests/e2e/common-operator-integ-suite.sh
```diff
@@ -87,6 +108,8 @@ initialize_variables() {
   COMMAND="kubectl"
   ARTIFACTS="${ARTIFACTS:-$(mktemp -d)}"
   KUBECONFIG="${KUBECONFIG:-"${ARTIFACTS}/config"}"
+  LOCALBIN="${LOCALBIN:-"${HOME}/.local/bin"}"
+  OPERATOR_SDK="${LOCALBIN}/operator-sdk"
```
FilipB, line 112: Who is installing `operator-sdk` bin there?

Makefile.core.mk
```diff
@@ -483,9 +483,10 @@ gitleaks: $(GITLEAKS) ## Download gitleaks to bin directory.
 $(GITLEAKS): $(LOCALBIN)
 	@test -s $(LOCALBIN)/gitleaks || GOBIN=$(LOCALBIN) go install github.com/zricethezav/gitleaks/v8@${GITLEAKS_VERSION}
 
+HELM_PLATFORM ?= openshift
```
FilipB, line 486: Could you please add a comment with more context and valid options?

tests/e2e/common-operator-integ-suite.sh
```diff
@@ -191,18 +214,50 @@ if [ "${SKIP_BUILD}" == "false" ]; then
...
+  if [ "${OLM}" == "true" ]; then
+    # Set the platform to Kubernetes by default.
+    # We are skipping the deploy via OLM test on OCP because the workaround to avoid the certificate issue is not working.
+    HELM_PLATFORM="kubernetes"
+    
+    # Install OLM in the cluster because it's not available by default in kind.
+    ${OPERATOR_SDK} olm install
+
+    # Set image-related variables
+    IMAGE_TAG_BASE="${HUB}/${IMAGE_BASE}"
```
FilipB, line 230: Why setting `IMAGE_TAG_BASE` here? Isn't that already defined from invocation of `make test.e2e.kind`

tests/e2e/common-operator-integ-suite.sh
```diff
@@ -191,18 +214,50 @@ if [ "${SKIP_BUILD}" == "false" ]; then
...
+
+    # Deploy the operator using OLM
+    IMAGE="${HUB}/${IMAGE_BASE}:${TAG}" \
+    IMAGE_TAG_BASE="${IMAGE_TAG_BASE}" \
+    BUNDLE_IMG="${BUNDLE_IMG}" \
+    HELM_PLATFORM="${HELM_PLATFORM}" \
+    make bundle bundle-build bundle-push
+
+    # Create operator namespace
+    ${COMMAND} create ns "${NAMESPACE}" || echo "namespace ${NAMESPACE} already exists"
```
FilipB, line 241: The msg can be misleading when the `oc create ns` actually fails because of something else.

FilipB:
> @FilipB Do we need to add any additional validation after the operator is being deployed? Or is it enough to validate with the current e2e coverage?
I believe it's sufficient after https://github.com/istio-ecosystem/sail-operator/pull/297 we will know that expected images are started.

Makefile.core.mk
fjglira: I Improved to make the set of the variable and for what is used more clear

tests/e2e/common-operator-integ-suite.sh
fjglira: Is already present in the builder image that we use to run everything

tests/e2e/common-operator-integ-suite.sh
fjglira: For some reason when we invoke inside the script is not sharing that previous variable declaration when the first target was called. That is why I added again the set of variable

tests/e2e/common-operator-integ-suite.sh
```diff
@@ -74,6 +79,22 @@ parse_flags() {
...
+  fi
+
+  if [ "${SKIP_DEPLOY}" == "true" ]; then
+    echo "Skipping deploy"
+  fi
+
+  if [ "${OLM}" == "true" ]; then
+    echo "OLM deployment enabled"
+    if [ "${OCP}" == "true" ]; then
+      # OLM deploy test is being skipped on OCP due errors with the workaround to avoid the certificate issue
```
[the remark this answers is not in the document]
fjglira: Added

tests/e2e/common-operator-integ-suite.sh
fjglira: Changed

tests/e2e/common-operator-integ-suite.sh
fjglira: Added the comment also

tests/e2e/common-operator-integ-suite.sh
fjglira: Added to check at the same time that `OLM` is true and `SKIP_DEPLOY` is false

tests/e2e/common-operator-integ-suite.sh
fjglira: So, if the user want to execute the target without a container will need to have operator-sdk installed previously

fjglira:
Changes from review done @FilipB @sridhargaddam

tests/e2e/common-operator-integ-suite.sh
```diff
@@ -74,6 +79,22 @@ parse_flags() {
...
+  fi
+
+  if [ "${SKIP_DEPLOY}" == "true" ]; then
+    echo "Skipping deploy"
+  fi
+
+  if [ "${OLM}" == "true" ]; then
+    echo "OLM deployment enabled"
+    if [ "${OCP}" == "true" ]; then
+      echo "OLM deploy test is being skipped on OCP due errors with the workaround to avoid the certificate issue"
```
sridhargaddam, line 94: What about the following sentence?

```suggestion
      echo "Skipping operator deployment using OLM on OCP clusters due to certificate issues with the internal registry."
```

sridhargaddam (commented):
thanks @fjglira 

One small comment, otherwise LGTM

tests/e2e/common-operator-integ-suite.sh
fjglira: Changed

sridhargaddam:
LGTM. Will let @FilipB to review the updated PR.

tests/e2e/common-operator-integ-suite.sh
FilipB: I'm not sure if I understand. Why creating `mkdir -p "${LOCALBIN}"` when `operator-sdk` should be already there? If it's not there, it does not make sense to create that dir anyway.

tests/e2e/common-operator-integ-suite.sh
```diff
@@ -191,18 +220,50 @@ if [ "${SKIP_BUILD}" == "false" ]; then
...
-    # To avoid errors of certificates meanwhile we are pulling the operator image from the internal registry
-    # We need to set image $HUB to a fixed known value after the push
-    # This value always will be equal to the svc url of the internal registry
-    HUB="image-registry.openshift-image-registry.svc:5000/sail-operator"
+  # If OLM is enabled, deploy the operator using OLM
+  if [ "${OLM}" == "true" ] && [ "${SKIP_DEPLOY}" == "false" ]; then
+    # Set the platform to Kubernetes by default.
+    # We are skipping the deploy via OLM test on OCP because the workaround to avoid the certificate issue is not working.
+    # Jora ticket related to the limitation: https://issues.redhat.com/browse/OSSM-7993
+    HELM_PLATFORM="kubernetes"
```
FilipB, line 230: Here is the case I mentioned above, HELM_PLATFORM is ignored unless it's 'openshift'

tests/e2e/common-operator-integ-suite.sh
FilipB: But you are setting `IMAGE_TAG_BASE` which is not available from parent Makefile invocation but at the same time you expect that `TAG` is available. Or at least I don't see you define `TAG` anywhere.

tests/e2e/common-operator-integ-suite.sh
fjglira: I added some comments to clarify why we are setting it to `Kubernetes` another option is to change the variable name to OCP and fill it empty in the call of the target to not set the helm flag. But I think is more explicit to set it to kubernetes

Makefile.core.mk
fjglira: Ok, I changed the variable name to OPENSHIFT_PLATFORM with expected values true or false, if set to true with set the flag if not it will not add the flag

tests/e2e/common-operator-integ-suite.sh
fjglira: Changed. It should be already there as you mention

tests/e2e/common-operator-integ-suite.sh
fjglira: If I do this instead of setting the `IMAGE_TAG_BASE` to set later `BUNDLE_IMG` we get errors because `BUNDLE_IMG` is not correctly filled:

```
    HUB="${HUB}" \
    IMAGE_BASE="${IMAGE_BASE}" \
    OPENSHIFT_PLATFORM="false" \
    make bundle bundle-build bundle-push
```
When bundle-push is run we got that the `BUNDLE_IMG` is being filled with the default value:
```
docker build -f bundle.Dockerfile -t quay.io/maistra-dev/sail-operator-bundle:v0.1.0 .
[+] Building 2.1s (7/7) FINISHED                                                                                                                            docker:default
 => [internal] load build definition from bundle.Dockerfile                                                                                                           
```

This is because `IMAGE_TAG_BASE` and `BUNDLE_IMG` previously existed when the first target is getting invoked. So we need to set those values  manually inside the script

fjglira: `TAG` has the default value from the first make target invocation also, that is Why this value is filled correctly

fjglira: This type of issue we already have it before and is because of the circle invocation of the make targets. These variables are not going to be filled again in the make target because in the first invocation, they already filled:
```
# For example, running 'make bundle-build bundle-push catalog-build catalog-push' will build and push both
# quay.io/sail-dev/sail-operator-bundle:$VERSION and quay.io/sail-dev/sail-operator-catalog:$VERSION.
IMAGE_TAG_BASE ?= ${HUB}/${IMAGE_BASE}

BUNDLE_MANIFEST_DATE := $(shell cat bundle/manifests/${OPERATOR_NAME}.clusterserviceversion.yaml 2>/dev/null | grep createdAt | awk '{print $$2}')

# BUNDLE_IMG defines the image:tag used for the bundle.
# You can use it as an arg. (E.g make bundle-build BUNDLE_IMG=<some-registry>/<project-name-bundle>:<tag>)
BUNDLE_IMG ?= $(IMAGE_TAG_BASE)-bundle:v$(VERSION)
```

Makefile.core.mk
fjglira: Like this it will be more easy to understand and make more sense according what we are setting or not setting in the helm command

FilipB (approved):

PR #318, opened 2024-08-22
Fetch Istio CRDs from Istio version specified in go.mod

luksa:
Previously, the version of Istio CRDs was determined by the `crdSourceVersion` field in `versions.yaml`. This change ensures that Istio CRDs are now pulled directly from the istio.io/istio version specified in the go.mod file.


dgn (approved):
lgtm. should we cherry-pick it to 0.1?

PR #325, opened 2024-08-28
docs: add section on migrating from in-cluster operator

dgn:
Fixes https://github.com/istio-ecosystem/sail-operator/issues/217

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -14,6 +14,7 @@
     - [Installing through the web console](#installing-through-the-web-console)
     - [Installing using the CLI](#installing-using-the-cli)
   - [Installation from Source](#installation-from-source)
+- [Migrating from Istio in-cluster Operator](#migrating-from-istio-in-cluster-operator)
 - [Gateways](#gateways)
 - [Update Strategy](#update-strategy)
   - [InPlace](#inplace)
@@ -63,6 +64,8 @@ spec:
           cpu: 100m
           memory: 1024Mi
 ```
+
+Istio uses a ConfigMap for its global configuration, called the MeshConfig. All of its settings are available through `spec.meshConfig`.
 
 To support canary updates of the control plane, Sail Operator includes support for multiple Istio versions. You can select a version by setting the `version` field in the `spec` to the version you would like to install, prefixed with a `v`. You can then update to a new version just by changing this field.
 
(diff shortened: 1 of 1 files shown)


fjglira:
```
make[1]: Leaving directory '/home/prow/go/src/github.com/istio-ecosystem/sail-operator'
./tests/e2e/common-operator-integ-suite.sh: line 224: /home/prow/go/src/github.com/istio-ecosystem/sail-operator/bin/operator-sdk: No such file or directory
+ cleanup_kind_cluster operator-integration-tests
```
I need to verify Why is not finding operator-sdk binary, we have everything set in the make file to get the binary

docs/README.md
```diff
@@ -187,6 +190,57 @@ is installed. `Succeeded` should appear in the **Status** column.
 
 If you're not using OpenShift or simply want to install from source, follow the [instructions in the Contributor Documentation](../README.md#deploying-the-operator).
 
+## Migrating from Istio in-cluster Operator
```
fjglira, line 193: Should we be adding also an example of the entire process? Or is too much information for this kind of docs?

dgn: I think we don't want to offer a complete migration path at this stage... You can of course do a revision-based update from your existing control plane to the one deployed by Sail Operator, but I'd say we let people figure it out. If we can add a more detailed guide down the line that'll be great, but the important part for now I think is that we point out how to recreate the same instance, people can take it from there

docs/README.md
```diff
@@ -187,6 +190,57 @@ is installed. `Succeeded` should appear in the **Status** column.
...
+
+If you're planning to migrate from the [now-deprecated Istio in-cluster operator](https://istio.io/latest/blog/2024/in-cluster-operator-deprecation-announcement/) to the Sail Operator, you will have to make some adjustments to your Kubernetes Resources. While direct usage of the IstioOperator resource is not possible with the Sail Operator, you can very easily transfer all your settings to the respective Sail Operator APIs. As shown in the [Concepts](#concepts) section, every API resource has a `spec.values` field which accepts the same input as the `IstioOperator`'s `spec.values` field. Also, the [Istio resource](#istio-resource) provides a `spec.meshConfig` field, just like IstioOperator does.
+
+Another important distinction between the two operators is that Sail Operator can manage and install different versions of Istio and its components, whereas the in-cluster operator always installs the version of Istio that it was released with. This makes managing control plane upgrades much easier, as the operator update is disconnected from the control plane update.
+
+So for a simple Istio deployment, the transition will be very easy:
```
FilipB, line 199: Should we test if this transition will work in-place? Or if it's necessary to first remove IstioOperator and only then create Istio CR?

dgn: I'd leave it up to the users to figure out for now. The important bit is how they can deploy the same control plane using Sail Operator - the exact migration steps for a no-downtime control plane migration are a bigger topic. We can add that later if there's demand, maybe somebody from the community will contribute their steps

FilipB (approved):

PR #330, opened 2024-09-05
Issue #91 Create multicluster e2e test

### LINKED ISSUE #91: e2e multicluster tests
There needs to be e2e tests to ensure you can use the Sail operator and APIs to setup a multicluster istio deployment. The tests need to cover setting up:

1. multi-primary
2. primary-remote
3. external controlplane

These e2e tests should:
- Create two separate kind clusters
- Install the sail operator/crds into each
- Follow the upstream documentation for setting up the above multicluster deployments but use the Sail CRs rather than `istioctl install ...`: https://istio.io/latest/docs/setup/install/multicluster/multi-primary/
  - A shared trust will need to be created for both clusters: https://istio.io/latest/docs/setup/install/multicluster/before-you-begin/
- Ensure that istio is setup correctly after install

fjglira:
Fixes #91 

Description of the changes and steps added to this PR:

- Updated the common script to be able to run multicluster setup using the upstream setup kind script
- Updated both setup scripts to add the capability to create multicluster setup on kind
- Add multicluster test cases
- Modify the k8s client init to be able to set from different kubeconfig files (This will be improved in a consecutive PR)
- Moved GetVersionFromIstiod() to common utils
- Added istioctl util to be able to run istioctl commands in the e2e framework
- Modify kubectl to add the possibility to use different kubeconfig files (Pending the refactor of the kubectl util to be more flexible)
- Add download istioctl binary and set variable to the make file
- Created the util `certs` to avoid having to create the certificates by using the Make file from upstream


tests/e2e/util/kubectl/kubectl.go
```diff
@@ -41,8 +49,8 @@ func kubectl(format string, args ...interface{}) string {
 }
 
 // CreateFromString creates a resource from the given yaml string
-func CreateFromString(yamlString string) error {
-	cmd := kubectl("create -f -")
+func CreateFromString(yamlString string, kubeconfig ...string) error {
```
luksa, line 52: Hmm, not sure adding `kubeconfig` to all the functions is a good approach. Instead, we should be able to create  a new kubectl object and pass the kubeconfig to it.

luksa: But now I'm also thinking that we should maybe get rid of this `kubectl` thing altogether and just invoke `shell.ExecuteCommand` directly. Wouldn't that make the code simpler? Currently, `kubectl.go` represents another abstraction layer that is starting to become hard to understand due to all the different usage options.

luksa: Or maybe use the builder pattern to construct the command.

fjglira: > But now I'm also thinking that we should maybe get rid of this `kubectl` thing altogether and just invoke `shell.ExecuteCommand` directly. Wouldn't that make the code simpler? Currently, `kubectl.go` represents another abstraction layer that is starting to become hard to understand due to all the different usage options.

Yes, using `shell.ExecuteCommand` will be simpler to get rid of the kubeconfig param, but I think it will not be so useful because we will need to set everything in the steps of the test, and that will make the code less readable as is now. I can try the builder pattern, which I think can be a better

fjglira: Hey @luksa I marked the PR as ready for review, but I did not address the `kubectl` handle kubeconfig on this PR. I'll do it in a consecutive PR, I think the PR is complex enough to add more complexity by refactoring the `kubectl` util

tests/e2e/integ-suite-kind.sh
```diff
@@ -52,16 +62,67 @@ function setup_kind_registry() {
...
   done
 }
 
-KUBECONFIG="${ARTIFACTS}/config" setup_kind_cluster "${KIND_CLUSTER_NAME}" "" "" "true" "true"
-setup_kind_registry
+# Create a self signed root CA and intermediate CAs and push them to each cluster.
+function setup_ca() {
+    # Create a self signed root CA and intermediate CAs.
+    mkdir -p certs
+    pushd certs
```
nrfox, line 83: These should probably write to a `/tmp` folder and not to the current folder.

tests/e2e/integ-suite-kind.sh
```diff
@@ -52,16 +62,67 @@ function setup_kind_registry() {
...
+    for node in $(kind get nodes --name="${cluster}"); do
+      kubectl annotate node "${node}" "kind.x-k8s.io/registry=localhost:${KIND_REGISTRY_PORT}" --overwrite;
+    done
   done
 }
 
-KUBECONFIG="${ARTIFACTS}/config" setup_kind_cluster "${KIND_CLUSTER_NAME}" "" "" "true" "true"
-setup_kind_registry
+# Create a self signed root CA and intermediate CAs and push them to each cluster.
+function setup_ca() {
```
nrfox, line 80: I'm wondering if we want to do this here at all or if it should be done inside `pushIntermediateCA` in the test? Then the tests aren't reliant on connectivity to `https://raw.githubusercontent.com` or the makefile changing.

tests/e2e/multicluster/multicluster_multiprimary_test.go
```diff
@@ -0,0 +1,479 @@
...
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ = Describe("Multicluster deployment models", Ordered, func() {
+	SetDefaultEventuallyTimeout(180 * time.Second)
+	SetDefaultEventuallyPollingInterval(time.Second)
+
+	// debugInfoLogged := false
```
nrfox, line 51: Can this be removed?
```suggestion
```

tests/e2e/multicluster/multicluster_multiprimary_test.go
```diff
@@ -0,0 +1,479 @@
...
+
+	BeforeAll(func(ctx SpecContext) {
+		if !skipDeploy {
+			// Deploy the Sail Operator on both clusters
+			Expect(kubectl.CreateNamespace(namespace, kubeconfig)).To(Succeed(), "Namespace failed to be created on Primary Cluster")
+			Expect(kubectl.CreateNamespace(namespace, kubeconfig2)).To(Succeed(), "Namespace failed to be created on Remote Cluster")
```
nrfox, line 57: nit: the second cluster isn't always a "Remote" cluster like in the case of multi-primary. Maybe it's possible to expose the name of the cluster here? If `kubeconfig` we're an object for example. Not a blocker though so if it's too much trouble then don't worry about it.

tests/e2e/integ-suite-kind.sh
```diff
@@ -52,16 +62,67 @@ function setup_kind_registry() {
   fi
 
   # https://docs.tilt.dev/choosing_clusters.html#discovering-the-registry
-  # TODO get context/config from existing variables
-  kind export kubeconfig --name="${KIND_CLUSTER_NAME}"
-  for node in $(kind get nodes --name="${KIND_CLUSTER_NAME}"); do
-    kubectl annotate node "${node}" "kind.x-k8s.io/registry=localhost:${KIND_REGISTRY_PORT}" --overwrite;
+  for cluster in $(kind get clusters); do
```
nrfox, line 65: If you have other kind clusters prior to running the tests, won't this pickup those too?

tests/e2e/multicluster/multicluster_multiprimary_test.go
```diff
@@ -0,0 +1,479 @@
...
+  name: default
+spec:
+  version: %s
+  namespace: %s
+  values:
+    global:
+      meshID: %s
+      multiCluster:
+        clusterName: %s
+      network: %s`
```
nrfox, line 98: Non blocking question: is there not a generated sail-operator go client we could use for the e2e tests?

tests/e2e/multicluster/multicluster_multiprimary_test.go
```diff
@@ -0,0 +1,479 @@
...
+						multiclusterPrimaryYAML := fmt.Sprintf(multiclusterYAML, version.Name, controlPlaneNamespace, "mesh1", "cluster1", "network1")
+						Log("Istio CR Primary: ", multiclusterPrimaryYAML)
+						Expect(kubectl.CreateFromString(multiclusterPrimaryYAML, kubeconfig)).To(Succeed(), "Istio Resource creation failed on Primary Cluster")
+
+						multiclusterRemoteYAML := fmt.Sprintf(multiclusterYAML, version.Name, controlPlaneNamespace, "mesh1", "cluster2", "network2")
+						Log("Istio CR Remote: ", multiclusterRemoteYAML)
+						Expect(kubectl.CreateFromString(multiclusterRemoteYAML, kubeconfig2)).To(Succeed(), "Istio Resource creation failed on Remote Cluster")
```
nrfox, line 105: nit: again since this is a multi-primary deployment, this is not a `Remote` cluster.

tests/e2e/multicluster/multicluster_multiprimary_test.go
```diff
@@ -0,0 +1,479 @@
...
+	Expect(kubectl.ApplyWithLabels(ns, helloWorldURL, "version=v1", kubeconfig)).To(Succeed(), "Sample service deploy failed on Primary Cluster")
+	Expect(kubectl.ApplyWithLabels(ns, helloWorldURL, "version=v2", kubeconfig2)).To(Succeed(), "Sample service deploy failed on Remote Cluster")
+
+	sleepURL := fmt.Sprintf("https://raw.githubusercontent.com/istio/istio/%s/samples/sleep/sleep.yaml", version)
+	Expect(kubectl.Apply(ns, sleepURL, kubeconfig)).To(Succeed(), "Sample sleep deploy failed on Primary Cluster")
+	Expect(kubectl.Apply(ns, sleepURL, kubeconfig2)).To(Succeed(), "Sample sleep deploy failed on Remote Cluster")
+}
+
+// getPodName returns the pod name of the given deployment in the given namespace
+func getSleepPodName(ctx context.Context, cl client.Client, ns, deploymentName string) string {
```
nrfox, line 335: does `getSleepPodName` need to be its own function? Can just add the name of the pod as an additional argument to make it a little more generic. From the comment it looks like this used to be `getPodName`?

tests/e2e/multicluster/multicluster_multiprimary_test.go
```diff
@@ -0,0 +1,479 @@
...
+		Expect(err).NotTo(HaveOccurred())
+		responses = append(responses, response)
+	}
+	return responses
+}
+
+// getSVCAddress returns the address of the istiod service in the given cluster
+func getSVCAddress(ctx context.Context, cl client.Client, ns string) string {
+	svc := &corev1.Service{}
+	err := cl.Get(ctx, kube.Key("istio-eastwestgateway", ns), svc)
```
nrfox, line 363: Should this be hardcoded to `istio-eastwestgateway`? Similar to `getPodName` you can add the svc name as an argument here to make it a little more generic.

tests/e2e/multicluster/multicluster_multiprimary_test.go
```diff
@@ -0,0 +1,479 @@
...
+	Expect(err).NotTo(HaveOccurred())
+	return svc.Status.LoadBalancer.Ingress[0].IP
```
nrfox, line 365: This will panic if Ingress isn't ready which could make this failure more confusing. Can you first add a check to ensure that ingress has at least one element?

tests/e2e/multicluster/multicluster_multiprimary_test.go
```diff
@@ -0,0 +1,479 @@
...
+}
+
+// pushIntermediateCA pushes the intermediate CA to the cluster
+func pushIntermediateCA() {
+	// Create the namespace
+	Expect(kubectl.CreateNamespace(controlPlaneNamespace, kubeconfig)).To(Succeed(), "Namespace failed to be created")
+	Expect(kubectl.CreateNamespace(controlPlaneNamespace, kubeconfig2)).To(Succeed(), "Namespace failed to be created")
+
+	// Set cert dir
+	certDir := filepath.Join(project.RootDir, "certs")
```
nrfox, line 375: See earlier comments about using a `/tmp` folder for this.

tests/e2e/multicluster/multicluster_multiprimary_test.go
```diff
@@ -0,0 +1,479 @@
...
+
+	AfterAll(func(ctx SpecContext) {
+		// Delete the Sail Operator from both clusters
+		Expect(kubectl.DeleteNamespace(namespace, kubeconfig)).To(Succeed(), "Namespace failed to be deleted on Primary Cluster")
+		Expect(kubectl.DeleteNamespace(namespace, kubeconfig2)).To(Succeed(), "Namespace failed to be deleted on Remote Cluster")
+
+		// Delete the intermediate CA from both clusters
+		common.CheckNamespaceEmpty(ctx, clPrimary, namespace)
+		common.CheckNamespaceEmpty(ctx, clRemote, namespace)
+	})
```
nrfox, line 302: Do we need to also delete all the resources that we're created? `Istio`, `sample`, gateways, etc.

tests/e2e/multicluster/multicluster_multiprimary_test.go
```diff
@@ -0,0 +1,479 @@
...
+							Should(HaveCondition(appsv1.DeploymentAvailable, metav1.ConditionTrue), "Istiod is not Available on Remote; unexpected Condition")
+						Expect(common.GetVersionFromIstiod()).To(Equal(version.Version), "Unexpected istiod version")
+						Success("Istiod is deployed in the namespace and Running on Remote Cluster")
+					})
+
+				})
+
+				When("Gateway is created in both clusters", func() {
+					BeforeAll(func(ctx SpecContext) {
+						eastGatewayURL := "https://raw.githubusercontent.com/istio-ecosystem/sail-operator/main/docs/multicluster/east-west-gateway-net1.yaml"
```
nrfox, line 138: This makes the tests reliant on connectivity to `http://raw.githubusercontent.com`. Maybe we can copy this yaml into a `testdata` folder or construct the yaml inside the tests?

common/scripts/kind_provisioner.sh
```diff
@@ -34,6 +34,9 @@ set -x
 # DEFAULT_KIND_IMAGE is used to set the Kubernetes version for KinD unless overridden in params to setup_kind_cluster(s)
 DEFAULT_KIND_IMAGE="gcr.io/istio-testing/kind-node:v1.28.4"
 
+# the default kind cluster should be ipv4 if not otherwise specified
+IP_FAMILY="${IP_FAMILY:-ipv4}"
```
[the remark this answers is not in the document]
fjglira: This script change came directly from the upstream one, we copy this files automatically

tests/e2e/integ-suite-kind.sh
```diff
@@ -52,16 +62,67 @@ function setup_kind_registry() {
...
 
-KUBECONFIG="${ARTIFACTS}/config" setup_kind_cluster "${KIND_CLUSTER_NAME}" "" "" "true" "true"
-setup_kind_registry
+# Create a self signed root CA and intermediate CAs and push them to each cluster.
+function setup_ca() {
+    # Create a self signed root CA and intermediate CAs.
+    mkdir -p certs
+    pushd certs
+    curl -fsL -o common.mk "https://raw.githubusercontent.com/istio/istio/1.23.0/tools/certs/common.mk"
+    curl -fsL -o Makefile.selfsigned.mk "https://raw.githubusercontent.com/istio/istio/1.23.0/tools/certs/Makefile.selfsigned.mk"
```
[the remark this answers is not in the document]
fjglira: Yes, I'll add a way to properly install the correct files from the current istio version that we are testing

tests/e2e/config/multicluster.json
```diff
@@ -0,0 +1,14 @@
...
+      "pod_subnet": "10.10.0.0/16",
+      "svc_subnet": "10.255.10.0/24",
+      "network_id": "0"
+    },
+    {
+      "cluster_name": "remote",
+      "pod_subnet": "10.20.0.0/16",
+      "svc_subnet": "10.255.20.0/24",
+      "network_id": "0"
+    }
```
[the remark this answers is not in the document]
fjglira: Ok. I take it the multicluster setup from Istio upstream; I'll test with different networks because it made more sense to do it like that to simulate a more real scenario

tests/e2e/multicluster/multicluster_multiprimary_test.go
fjglira: Yes, I leave there by mistake. I'll add this debug information in a consecutive PR

tests/e2e/multicluster/multicluster_multiprimary_test.go
fjglira: I opened today an [issue](https://github.com/istio-ecosystem/sail-operator/issues/337) to do a refactor of the `kubectl` util because, yes, I need to find a better way to do this and simply use the way of managing that kubeconfig

common/scripts/kind_provisioner.sh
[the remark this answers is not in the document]
fjglira: So, I don't know if there is any issue with IPv6. We can test it

tests/e2e/integ-suite-kind.sh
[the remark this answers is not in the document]
fjglira: I moved the logic to a new func, now we are not relying into the make file from upstream

tests/e2e/multicluster/multicluster_multiprimary_test.go
fjglira: Changed

tests/e2e/multicluster/multicluster_multiprimary_test.go
fjglira: Done, created the generic func `GetPodNameByLabel`

tests/e2e/multicluster/multicluster_multiprimary_test.go
fjglira: Added a generic func to getsvc svc address

tests/e2e/multicluster/multicluster_multiprimary_test.go
fjglira: Added a validation to wait until the Loadbalancer to be ready

tests/e2e/multicluster/multicluster_multiprimary_test.go
fjglira: Sample apps are being deleted a step before finishing the node of the test:
```
When("sample apps are deleted in both clusters", func() {
					BeforeAll(func(ctx SpecContext) {
						// Delete the entire sample namespace in both clusters
						Expect(kubectl.DeleteNamespace("sample", kubeconfig)).To(Succeed(), "Namespace failed to be deleted on Primary Cluster")
						Expect(kubectl.DeleteNamespace("sample", kubeconfig2)).To(Succeed(), "Namespace failed to be deleted on Remote Cluster")
					})

					It("sample app is deleted in both clusters and the namespace is empty", func(ctx SpecContext) {
						common.CheckNamespaceEmpty(ctx, clPrimary, "sample")
						common.CheckNamespaceEmpty(ctx, clRemote, "sample")
						Success("Sample app is deleted in both clusters")
					})
				})
```

The same for Istio resources, gateways, etc

tests/e2e/multicluster/multicluster_multiprimary_test.go
fjglira: I'll check for a better way to do this. Copying this into the config folder can be an option. Construct the yaml also, but I do not want to make the test too difficult to read

tests/e2e/integ-suite-kind.sh
fjglira: Moved

fjglira:
Closing this PR after the mess and conflict with the rebase. I'll open a new one

PR #335, opened 2024-09-13
Add External Controlplane to user docs

nrfox:
Adds instructions for the [External Controlplane](https://istio.io/latest/docs/setup/install/external-controlplane/) setup using the Sail Operator.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -22,10 +22,11 @@
   - [RevisionBased](#revisionbased)
     - [Example using the RevisionBased strategy](#example-using-the-revisionbased-strategy)
 - [Multi-cluster](#multi-cluster)
+  - [Prerequisites](#prerequisites)
   - [Common Setup](#common-setup)
   - [Multi-Primary](#multi-primary---multi-network)
   - [Primary-Remote](#primary-remote---multi-network)
-  - [External Controlplane](#external-controlplane)
+  - [External Control Plane](#external-controlplane)
 - [Addons](#addons)
   - [Deploy Prometheus and Jaeger addons](#deploy-prometheus-and-jaeger-addons)
   - [Deploy Kiali addon](#deploy-kiali-addon)
@@ -476,13 +477,12 @@ Steps:
 
 You can use the Sail Operator and the Sail CRDs to manage a multi-cluster Istio deployment. The following instructions are adapted from the [Istio multi-cluster documentation](https://istio.io/latest/docs/setup/install/multicluster/) to demonstrate how you can setup the various deployment models with Sail. Please familiarize yourself with the different [deployment models](https://istio.io/latest/docs/ops/deployment/deployment-models/) before starting.
 
-*Prerequisites*
-
-Each deployment model requires you to install the Sail Operator and the Sail CRDs to every cluster that is part of the mesh.
+### Prerequisites
 
 - Install [istioctl](https://istio.io/latest/docs/setup/install/istioctl) and have it included in your `$PATH`.
 - Two kubernetes clusters with external lb support. (If using kind, `cloud-provider-kind` is running in the background)
 - kubeconfig file with a context for each cluster.
+- Install the Sail Operator and the Sail CRDs to every cluster.
 
 ### Common Setup
 
@@ -872,15 +872,15 @@ In this setup there is a Primary cluster (`cluster1`) and a Remote cluster (`clu
     kubectl delete ns sample --context="${CTX_CLUSTER2}"
     ```
 
-### External-Controlplane
+### External Control Plane
 
-These instructions install an [external-controlplane](https://istio.io/latest/docs/setup/install/external-controlplane/) Istio deployment using the Sail Operator and Sail CRDs. **Before you begin**, ensure meet the requirements of the [common setup](#common-setup) and complete **only** the "Setup env vars" step. Unlike the other Multi-Cluster deployments, you won't be creating a common CA in this setup.
+These instructions install an [external control plane](https://istio.io/latest/docs/setup/install/external-controlplane/) Istio deployment using the Sail Operator and Sail CRDs. **Before you begin**, ensure meet the requirements of the [common setup](#common-setup) and complete **only** the "Setup env vars" step. Unlike the other Multi-Cluster deployments, you won't be creating a common CA in this setup.
 
 These installation instructions are adapted from: https://istio.io/latest/docs/setup/install/external-controlplane/ and are intended to be run in a development environment such as `kind` and not in production.
 
-In this setup there is an External Controlplane cluster (`cluster1`) and a Remote cluster (`cluster2`) which are on separate networks.
+In this setup there is an external control plane cluster (`cluster1`) and a remote cluster (`cluster2`) which are on separate networks.
 
-1. Create an `Istio` resource on `cluster1` to manage the ingress gateways for the External Controlplane.
+1. Create an `Istio` resource on `cluster1` to manage the ingress gateways for the external control plane.
 
     ```sh
     kubectl create namespace istio-system --context "${CTX_CLUSTER1}"
@@ -898,7 +898,7 @@ In this setup there is an External Controlplane cluster (`cluster1`) and a Remot
     kubectl wait --context "${CTX_CLUSTER1}" --for=condition=Ready istios/default --timeout=3m
     ```
 
-2. Create the ingress gateway for the External Controlplane.
+2. Create the ingress gateway for the external control plane.
 
     ```sh
     kubectl --context "${CTX_CLUSTER1}" apply -f https://raw.githubusercontent.com/istio-ecosystem/sail-operator/main/docs/multicluster/controlplane-gateway.yaml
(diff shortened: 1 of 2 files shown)


docs/README.md
```diff
@@ -868,6 +872,274 @@ In this setup there is a Primary cluster (`cluster1`) and a Remote cluster (`clu
...
+    { kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd?ref=v1.1.0" | kubectl apply -f - --context="${CTX_CLUSTER2}"; }
+    ```
+
+    Expose `helloworld` through the ingress gateway
+    ```sh
+    kubectl apply -f https://raw.githubusercontent.com/istio/istio/${ISTIO_VERSION}/samples/helloworld/gateway-api/helloworld-gateway.yaml -n sample --context="${CTX_CLUSTER2}"
+    kubectl -n sample --context="${CTX_CLUSTER2}" wait --for=condition=programmed gtw helloworld-gateway
+    ```
+
+    Confirm you can access the `helloworld` application through the ingress gateway:
```
fjglira, line 1126: Can we add here something similar to what we have in the previous examples? I mean, to give the user more information about what to expect. Something like:
`Confirm you can access the `helloworldV2` application through the ingress gateway created in the Remote cluster`

docs/README.md
```diff
@@ -868,6 +872,274 @@ In this setup there is a Primary cluster (`cluster1`) and a Remote cluster (`clu
...
+              port:
+                number: 443
+    EOF
+    ```
+
+9. Create the `sample` namespace on the Remote cluster and label it to enable injection.
+
+    ```sh
+    kubectl create --context="${CTX_CLUSTER2}" namespace sample
+    kubectl label --context="${CTX_CLUSTER2}" namespace sample istio.io/rev=external-istiod
```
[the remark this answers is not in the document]
nrfox: This won't work with sail today unless you create an `Istio` with the name `default` since sail does not support the default tag. Although let me try playing with some of the helm settings to try and get this working.

docs/README.md
```diff
@@ -868,6 +872,274 @@ In this setup there is a Primary cluster (`cluster1`) and a Remote cluster (`clu
...
+      name: default
+    spec:
+      version: v${ISTIO_VERSION}
+      namespace: istio-system
+      values:
+        pilot:
+          resources:
+            requests:
+              cpu: 100m
+              memory: 1024Mi
```
luksa, line 900: I'm thinking we should introduce a new profile called `kind` instead of repeating this in all the examples. Better yet, check if this is really (still?) needed when deploying kind with defaults and if not, simplify remove this.

docs/README.md
```diff
@@ -868,6 +872,274 @@ In this setup there is a Primary cluster (`cluster1`) and a Remote cluster (`clu
...
+        global:
+          network: network1
+    EOF
+    kubectl wait --context "${CTX_CLUSTER1}" --for=jsonpath='{.status.revisions.ready}'=1 istios/default --timeout=3m
```
luksa, line 904: We simplified this in the other multi-cluster examples:
```suggestion
    kubectl wait --context "${CTX_CLUSTER1}" --for=condition=Ready istios/default --timeout=3m
```

docs/README.md
nrfox: I don't think it's needed. Maybe in CI where resources are more constrained? But IMO we don't need this in the examples. I probably just copy/pasta'd or branched off an old example that included this. I will remove it.

docs/README.md
nrfox: I'll go back through and compare the two examples because, similar to specifying the resource requirements, I probably just branched off an older example.

docs/README.md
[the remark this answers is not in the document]
nrfox: I wanted to try setting `RemoteIstio.spec.values.defaultRevision: external-istiod` but this only affects the validating webhook: https://github.com/istio/istio/blob/b957e7bd33d4db0023865d3f3f00b639f14665e5/manifests/charts/istiod-remote/templates/defaultrevisionvalidatingadmissionpolicy.yaml#L10. There is still no mutating webhook, which enables the `istio-injection` label, for the default tag deployed in the `Remote` cluster because sail does not support the default tag yet.

docs/README.md
```diff
@@ -868,6 +872,305 @@ In this setup there is a Primary cluster (`cluster1`) and a Remote cluster (`clu
...
+
+    ```
+11. Verify that you see a response from both v1 and v2 on `cluster1`.
+
+    `cluster1` responds with v1 and v2
+    ```sh
+    kubectl exec --context="${CTX_CLUSTER1}" -n sample -c sleep \
+        "$(kubectl get pod --context="${CTX_CLUSTER1}" -n sample -l \
+        app=sleep -o jsonpath='{.items[0].metadata.name}')" \
+        -- curl -sS helloworld.sample:5000/hello
```
sridhargaddam, line 1106: In the documented steps, we did not create any `sample` namespace on cluster1.

nrfox: Agh I think this was yet another bad copy/paste error.

docs/README.md
```diff
@@ -868,6 +872,305 @@ In this setup there is a Primary cluster (`cluster1`) and a Remote cluster (`clu
...
+    kubectl apply --context "${CTX_CLUSTER1}" -f - <<EOF
+    apiVersion: sailoperator.io/v1alpha1
+    kind: Istio
+    metadata:
+      name: default
+    spec:
+      version: v${ISTIO_VERSION}
+      namespace: istio-system
+      global:
+        network: network1
```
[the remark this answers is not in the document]
nrfox: They are in separate networks but we're not actually labeling `cluster2` with `network2` anywhere. This was only relevant if you added a third cluster as a remote of the remote cluster but I did not include those steps here: https://istio.io/latest/docs/setup/install/external-controlplane/#setup-east-west-gateways. I can include those steps but it would be largely the same as the primary-remote setup.

docs/README.md
```diff
@@ -868,6 +872,274 @@ In this setup there is a Primary cluster (`cluster1`) and a Remote cluster (`clu
     kubectl delete ns sample --context="${CTX_CLUSTER2}"
     ```
 
+### External-Controlplane
+
+These instructions install an [external-controlplane](https://istio.io/latest/docs/setup/install/external-controlplane/) Istio deployment using the Sail Operator and Sail CRDs. **Before you begin**, ensure you complete the [common setup](#common-setup).
```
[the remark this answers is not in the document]
nrfox: This might not be obvious but it is part of the pre-req instructions: https://github.com/istio-ecosystem/sail-operator/tree/main/docs#multi-cluster:

> Each deployment model requires you to install the Sail Operator and the Sail CRDs to every cluster that is part of the mesh.

I think that should just be another bullet point in the list.

docs/README.md
```diff
@@ -868,6 +872,274 @@ In this setup there is a Primary cluster (`cluster1`) and a Remote cluster (`clu
...
+    kubectl --context "${CTX_CLUSTER1}" wait '--for=jsonpath={.status.loadBalancer.ingress[].ip}' --timeout=30s svc istio-ingressgateway -n istio-system
+    ```
+
+3. Configure your environment to expose the ingress gateway.
+
+    **Note:** these instructions are intended to be executed in a test environment. For production environments, please refer to: https://istio.io/latest/docs/setup/install/external-controlplane/#set-up-a-gateway-in-the-external-cluster and https://istio.io/latest/docs/tasks/traffic-management/ingress/secure-ingress/#configure-a-tls-ingress-gateway-for-a-single-host for setting up a secure ingress gateway.
+
+    ```sh
+    export EXTERNAL_ISTIOD_ADDR=$(kubectl -n istio-system --context="${CTX_CLUSTER1}" get svc istio-ingressgateway -o jsonpath='{.status.loadBalancer.ingress[0].ip}')
+    export SSL_SECRET_NAME=NONE
```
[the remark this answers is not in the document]
nrfox: Yep good catch. This was part of another step that I consolidated but forgot to remove this env var.

docs/README.md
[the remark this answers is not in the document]
sridhargaddam: Yeah, no need to add the third cluster use-case. I was just trying to confirm that networks of cluster1 and  2 can be different.

docs/README.md
```diff
@@ -868,6 +872,284 @@ In this setup there is a Primary cluster (`cluster1`) and a Remote cluster (`clu
     kubectl delete ns sample --context="${CTX_CLUSTER2}"
     ```
 
+### External Control Plane
+
+These instructions install an [external control plane](https://istio.io/latest/docs/setup/install/external-controlplane/) Istio deployment using the Sail Operator and Sail CRDs. **Before you begin**, ensure meet the requirements of the [common setup](#common-setup) and complete **only** the "Setup env vars" step. Unlike the other Multi-Cluster deployments, you won't be creating a common CA in this setup.
```
sridhargaddam, line 877:
```suggestion
These instructions install an [external control plane](https://istio.io/latest/docs/setup/install/external-controlplane/) Istio deployment using the Sail Operator and Sail CRDs. **Before you begin**, ensure you meet the requirements of the [common setup](#common-setup) and complete **only** the "Setup env vars" step. Unlike other Multi-Cluster deployments, you won't be creating a common CA in this setup.
```

docs/README.md
```diff
@@ -868,6 +872,284 @@ In this setup there is a Primary cluster (`cluster1`) and a Remote cluster (`clu
...
+
+These installation instructions are adapted from: https://istio.io/latest/docs/setup/install/external-controlplane/ and are intended to be run in a development environment such as `kind` and not in production.
```
sridhargaddam, line 879:
```suggestion
These installation instructions are adapted from [Istio's external control plane documentation](https://istio.io/latest/docs/setup/install/external-controlplane/) and are intended to be run in a development environment, such as `kind`, rather than in production.
```

sridhargaddam (approved):
Thanks Nick. Added hold, incase anyone wants to review it. Feel free to remove the hold if you want to merge it.

Issue #336, opened 2024-09-13
[FEATURE] Helm override for kube-rbac-proxy image


twhite0:
### Is this the right place to submit this?

- [X] This is not a question about how to use the sail-operator

### Feature Description

The [operator image](https://github.com/istio-ecosystem/sail-operator/blob/main/chart/templates/deployment.yaml#L82) provides for an override however the [kube-rbac-proxy](https://github.com/istio-ecosystem/sail-operator/blob/main/chart/templates/deployment.yaml#L56) container does not.  For folks like myself working in an environment with proxied image repositories, we need a means of overriding the image with our internal repo path.

Happy to contribute the changes, just following the "Way of Working" guide asking for a discussion up front.

### Additional Information

something like the following values.yaml enhancement.
```yaml
proxy:
   image:  gcr.io/kubebuilder/kube-rbac-proxy:v0.16.0
   imagePullPolicy: Always
```

and enhancement to deployment.yaml
```yaml
containers:
- args:
   name: kube-rbac-proxy
   image: {{ .Values.proxy.image }}
   imagePullPolicy: {{ .values.proxy.imagePullPolicy }}
```

PR #339, opened 2024-09-16
added helm overrides for proxy image as is available for operator

twhite0:
#### default
```bash
$ helm template helm-proxy-default chart  | grep image
        image: gcr.io/kubebuilder/kube-rbac-proxy:v0.16.0
        imagePullPolicy: Always
        image: quay.io/maistra-dev/sail-operator:0.2-latest
        imagePullPolicy: Always
```

#### override
```bash
helm template helm-proxy-override chart --set proxy.image=my/custom/image:tag --set proxy.imagePullPolicy=IfNotPresent | grep image
        image: my/custom/image:tag
        imagePullPolicy: IfNotPresent
        image: quay.io/maistra-dev/sail-operator:0.2-latest
        imagePullPolicy: Always

```

closes https://github.com/istio-ecosystem/sail-operator/issues/336


chart/values.yaml
```diff
@@ -44,6 +44,9 @@ csv:
     features.operators.openshift.io/csi: "false"
 image: quay.io/maistra-dev/sail-operator:0.2-latest
 imagePullPolicy: Always
+proxy:
+  image: gcr.io/kubebuilder/kube-rbac-proxy:v0.16.0
+  imagePullPolicy: Always
```
luksa, line 49: Do we want the default to be `Always`? Previously, we didn't set `imagePullPolicy`, which means that the policy was effectively `IfNotPresent`, given that the image tag was not `latest`.

twhite0: @luksa:  Good point that the behavior will be changed based on the existing absence of a value.  I followed the existing pattern but do understand that its using an explicit tag value rather than some form of _latest_.  Happy to change to IfNotPresent if you'd prefer.

luksa: Actually, I feel like it should be left unset. But since the same should also be true for the `imagePullPolicy` used in the main container, I'm okay with merging this as is. We should then set both to empty by default.

dgn:
You'll need to run `make gen` and add the email you used to commit to your GitHub account, then sign the CNCF CLA (as per the bot comment above)

dgn (approved):
Thank you @twhite0, this LGTM. Just needs a rebase

dgn:
Nevermind, it was just the timestamp on the CSV which caused a conflict. Fixed it myself

PR #349, opened 2024-09-19
Update/Fix Istioctl links and docs

longmuir:
A few minor updates to the user doc to update/fix istioctl links to point to the install-istioctl-tool.md, which provides istioctl installation instructions with the assumption that the sail-operator has installed Istio (and not istioctl).

### DIFF
--- a/docs/common/install-istioctl-tool.md
+++ b/docs/common/install-istioctl-tool.md
@@ -8,7 +8,9 @@ operators to debug and diagnose Istio service mesh deployments.
 
 Use an `istioctl` version that is the same version as the Istio control plane 
 for the Service Mesh deployment. See [Istio Releases](https://github.com/istio/istio/releases) for a list of valid 
-releases, including Beta releases.
+releases, including Beta releases. 
+
+These instructions assume that the Sail Operator has installed `Istio` in the `istio-system` namespace.
 
 
 ### Procedure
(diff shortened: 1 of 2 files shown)


docs/common/install-istioctl-tool.md
```diff
@@ -8,7 +8,9 @@ operators to debug and diagnose Istio service mesh deployments.
 
 Use an `istioctl` version that is the same version as the Istio control plane 
 for the Service Mesh deployment. See [Istio Releases](https://github.com/istio/istio/releases) for a list of valid 
-releases, including Beta releases.
+releases, including Beta releases. 
+
+These instructions assume that the Sail Operator has installed `Istio` in the `istio-system` namespace.
```
nrfox, line 13: Sorry if this is overly pedantic but `Istio` is a cluster-wide resource that is not namespace scoped. When I see `Istio` with backticks I assume you are talking about the `Istio` resource and if it's Istio without backticks I assume you're talking about Istio the project.

Maybe say something like the "Sail Operator has installed an `Istio` in your cluster? I don't think the namespace matters in this context. In fact, you can remove the namespace from step 2. below e.g.
```sh
$ oc get istio
```

longmuir: Good catch - thank! I had forgotten that `Istio` is a cluster wide resource, and the `-n istio-system` below implied to me that the resource must exist in the "istio-system" namespace, which is not guaranteed if it were namespace scoped. Based on what you said, I think it's probably best to just remove the `-n istio-system` part of the command below and I'll delete this addition. WDYT?

nrfox: @longmuir yep that sounds good! Sorry I missed your question until just now.

nrfox (approved):

PR #353, opened 2024-09-22
release workflow: Automatically adjust default release channel

dgn:
This will make sure that the channel field is pre-populated with a good default when we run the release workflow.


bundle.Dockerfile
```diff
@@ -5,7 +5,8 @@ LABEL operators.operatorframework.io.bundle.mediatype.v1=registry+v1
 LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=sailoperator
-LABEL operators.operatorframework.io.bundle.channels.v1="0.2"
+LABEL operators.operatorframework.io.bundle.channels.v1="0.2-dev"
+LABEL operators.operatorframework.io.bundle.channel.default.v1=0.2-dev
```
FilipB, line 9: Is this needed when we have just one channel? I believe this is only useful when having multiple channels in `operators.operatorframework.io.bundle.channel.default.v1`

Makefile.core.mk
```diff
@@ -79,7 +79,12 @@ GINKGO_FLAGS := $(if $(VERBOSE),-v) $(if $(CI),--no-color)
 # To re-generate a bundle for other specific channels without changing the standard setup, you can:
 # - use the CHANNELS as arg of the bundle target (e.g make bundle CHANNELS=candidate,fast,stable)
 # - use environment variables to overwrite this value (e.g export CHANNELS="candidate,fast,stable")
-CHANNELS ?= ${MINOR_VERSION}
+CHANNEL_SUFFIX := dev
```
FilipB, line 82: Looking at channel naming best [practice](https://olm.operatorframework.io/docs/best-practices/channel-naming/#example-2), this should be probably a prefix.

Makefile.core.mk
```diff
@@ -425,6 +432,9 @@ operator-chart:
 	sed -i -e "s|^\(image: \).*$$|\1${IMAGE}|g" \
 	       -e "s/^\(  version: \).*$$/\1${VERSION}/g" chart/values.yaml
 
+github-workflow:
+	sed -i -e '1,/default:/ s/^\(.*default:\).*$$/\1 ${DEFAULT_CHANNEL}/' .github/workflows/release.yaml
```
FilipB, line 436: IIUC the `default` field in the workflow actually means a default value for channels which can be list of channels and not a DEFAULT_CHANNEL

bundle.Dockerfile
dgn: good point. removed

Makefile.core.mk
dgn: fixed!

Makefile.core.mk
dgn: fixed! thank you

FilipB (approved):

Issue #355, opened 2024-09-23
Upload Helm artifacts for existing release


MaxBab:
### Is this the right place to submit this?

- [X] This is not a question about how to use the sail-operator

### Feature Description

Add helm artifact reference for release version of 0.1.0
The helm artifacts published during the release, stored as release
assets.
The "gh-pages" branch will store the "index.yaml" file, which will
contain the metadata for the artifacts and poing to the artifacts assets
location.

PR #356, opened 2024-09-23
Add helm artifact version sail-operator-0.2.0

MaxBab:
Add helm artifact for the released version of 0.2.0.
Helm artifacts will be added to `gh-pages` branch.

Fixes https://github.com/istio-ecosystem/sail-operator/issues/355


dgn:
Hey @MaxBab, latest release is 0.1.0 :)

MaxBab:
@dgn 

Oh, you're right.
My bad.
Fixing.

MaxBab:
Fixed.
Thanks a lot for noticing.

index.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
+entries:
+  sail-operator:
+  - apiVersion: v2
+    appVersion: 0.1.0
+    created: "2024-09-23T12:51:47.932616888+03:00"
+    description: A Helm chart for Kubernetes
+    digest: 94e0237a3476ef3f4784b299ec976bab05fdc73ccf4d978037ad996f0aa73c9f
```
sridhargaddam, line 8: When I quickly verified the sha256sum of the [release tar](https://github.com/istio-ecosystem/sail-operator/releases/download/0.1.0/sail-operator-0.1.0.tgz), I'm seeing a different digest. Please check the digest value once.

MaxBab: That is because I generated a new tar by using the `helm package ...` command and not took the artifact attached to the release.
I switched to the `release-0.1` branch.
I did it mostly because as part of the process I also created an `index.yaml` file which references the existing artifacts when the `helm repo add` command is used.

If you think I should rather add an artifact attached to the release, I could do it.

sridhargaddam: Got it. I'd let @dgn to comment if we want to use release artifact or the latest release-0.1 branch.

dgn: IMO it should be the same artifact from the Releases section, otherwise people might be confused/alarmed

MaxBab: No problem.
Will modify it.

MaxBab: Done

index.yaml
```diff
@@ -0,0 +1,14 @@
...
+  sail-operator:
+  - apiVersion: v2
+    appVersion: 0.1.0
+    created: "2024-09-23T19:48:26.553639312+03:00"
+    description: A Helm chart for Kubernetes
+    digest: ba77933659946e929da51baecd453a42b731a8a993563543efff275958ca2b98
+    name: sail-operator
+    type: application
+    urls:
+    - https://istio-ecosystem.github.io/sail-operator/sail-operator-0.1.0.tgz
```
dgn, line 12: maybe it's a stupid question, but would it be possible to just link to the release (https://github.com/istio-ecosystem/sail-operator/releases/download/0.1.0/sail-operator-0.1.0.tgz)? then we wouldn't have to pollute our repo with binaries

MaxBab: @dgn 

Wow, that's some really interesting approach.

I searched over the official documentation for separate store of `index.yaml` file, which holds the helm artifact metadata and the actual helm artifact, but haven't find a thing.

So, I did some simulation and it actually works.
If I'm placing the path of the release asset artifact under the `url` argument of `index.yaml` file, it's able to pull and install or upgrade helm deployment by using the artifacts from the gh release assets.

But there is one single thing to notice.
I found the following question and answer regarding the helm `urls` argument in the index.yaml file.
https://github.com/helm/helm/issues/7314#issuecomment-569790204

So, turns out that `urls` argument was designed to support a mirroring approach.
But later it was never requested an improvement or any change.
So, they left it in the API for backwards compatibility purposes.

Now, my concern comes from that it could be removed (maybe?) at some point.
So, I'm not sure it's good to implement this solution based on it.
From the other side, this thread is from 2019 and nothing changed since then.

What do you think?

Thanks.

dgn: it looks to me as though that is more about having >1 url. The downloader will always follow the link. Seems safe enough to me :) I'd say let's go ahead and add the link to the GitHub releases, keeps our git repo small. Thank you for investigating!

sridhargaddam: As @dgn mentioned, the helm charts downloader seems to support [multiple URL formats](https://github.com/helm/helm/tree/main/pkg/downloader/testdata/repository) and for fully-qualified URLs, it simply follows the url.
https://github.com/helm/helm/blob/main/pkg/downloader/chart_downloader_test.go

MaxBab: @dgn @sridhargaddam 

Done.

sridhargaddam (approved):

Issue #372, opened 2024-09-29
[Enhancement] Sail Operator dualStack support


sridhargaddam:
### Epic Description

Add necessary support in the Sail Operator to deploy Istio in dualStack mode and validate both control plane and dataplane tests.

- [x] [Deploy dualStack Kind clusters](https://github.com/istio-ecosystem/sail-operator/pull/373)
- [x] [Implement e2e dualStack control plane tests](https://github.com/istio-ecosystem/sail-operator/pull/373)
- [x] [Implement e2e dualStack data plane tests](https://github.com/istio-ecosystem/sail-operator/pull/383)
- [x] [Update the documentation](https://github.com/istio-ecosystem/sail-operator/pull/407)
- [x] [Add CI automation to run dualStack tests on Kind clusters](https://github.com/istio/test-infra/pull/5500).
- [x] [Support Sail Operator metrics on dualStack clusters](https://github.com/istio-ecosystem/sail-operator/pull/426)

PR #373, opened 2024-09-29
Implement e2e dualStack control plane tests

sridhargaddam:
This PR includes the necessary support
1. to deploy dualStack/IPv6 KIND clusters
2. to deploy the supported Istio versions in a dualStack mode

and verifies that Istiod is properly configured to support dualStack use-cases.

To run e2e dualStack tests, use the following command. 
$ export IP_FAMILY=dual
$ make test.e2e.kind

### DIFF
--- a/tests/e2e/dualstack/dualstack_suite_test.go
+++ b/tests/e2e/dualstack/dualstack_suite_test.go
@@ -1,0 +1,65 @@
+//go:build e2e
+
+// Copyright Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package dualstack
+
+import (
+	"testing"
+
+	k8sclient "github.com/istio-ecosystem/sail-operator/tests/e2e/util/client"
+	env "github.com/istio-ecosystem/sail-operator/tests/e2e/util/env"
+	"github.com/istio-ecosystem/sail-operator/tests/e2e/util/kubectl"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	cl                    client.Client
+	err                   error
+	ocp                   = env.GetBool("OCP", false)
+	namespace             = env.Get("NAMESPACE", "sail-operator")
+	deploymentName        = env.Get("DEPLOYMENT_NAME", "sail-operator")
+	controlPlaneNamespace = env.Get("CONTROL_PLANE_NS", "istio-system")
+	istioName             = env.Get("ISTIO_NAME", "default")
+	image                 = env.Get("IMAGE", "quay.io/maistra-dev/sail-operator:latest")
+	skipDeploy            = env.GetBool("SKIP_DEPLOY", false)
+	expectedRegistry      = env.Get("EXPECTED_REGISTRY", "^docker\\.io|^gcr\\.io")
+	multicluster          = env.GetBool("MULTICLUSTER", false)
+	ipFamily              = env.Get("IP_FAMILY", "ipv4")
+
+	k *kubectl.KubectlBuilder
+)
+
+func TestDualStack(t *testing.T) {
+	if ipFamily != "dual" || multicluster {
+		t.Skip("Skipping the dualStack tests")
+	}
+
+	RegisterFailHandler(Fail)
+	setup()
+	RunSpecs(t, "DualStack test suite")
+}
+
+func setup() {
+	GinkgoWriter.Println("************ Running Setup ************")
+
+	GinkgoWriter.Println("Initializing k8s client")
+	cl, err = k8sclient.InitK8sClient("")
+	Expect(err).NotTo(HaveOccurred())
+
+	k = kubectl.NewKubectlBuilder()
+}
--- a/tests/e2e/dualstack/dualstack_test.go
+++ b/tests/e2e/dualstack/dualstack_test.go
@@ -37,7 +37,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-var _ = Describe("Installation on a dualStack cluster", Ordered, func() {
+var _ = Describe("DualStack configuration ", Ordered, func() {
 	SetDefaultEventuallyTimeout(180 * time.Second)
 	SetDefaultEventuallyPollingInterval(time.Second)
 
@@ -63,7 +63,7 @@ var _ = Describe("Installation on a dualStack cluster", Ordered, func() {
 		Success("Operator is deployed in the namespace and Running")
 	})
 
-	Describe("using supported Istio version", func() {
+	Describe("for supported versions", func() {
 		for _, version := range supportedversion.List {
 			// Note: This var version is needed to avoid the closure of the loop
 			version := version
@@ -73,12 +73,12 @@ var _ = Describe("Installation on a dualStack cluster", Ordered, func() {
 				continue
 			}
 
-			Context(version.Name, func() {
+			Context("Istio version is: "+version.Version, func() {
 				BeforeAll(func() {
 					Expect(k.CreateNamespace(controlPlaneNamespace)).To(Succeed(), "Istio namespace failed to be created")
 				})
 
-				When("the Istio CR is created", func() {
+				When("the Istio CR is created with DualStack configuration", func() {
 					BeforeAll(func() {
 						istioYAML := `
 apiVersion: sailoperator.io/v1alpha1
--- a/tests/e2e/dualstack/dualstack_test.go
+++ b/tests/e2e/dualstack/dualstack_test.go
@@ -1,0 +1,213 @@
+//go:build e2e
+
+// Copyright Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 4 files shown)


tests/e2e/dualstack/dualstack_test.go
```diff
@@ -0,0 +1,213 @@
...
+			Expect(helm.Install("sail-operator", filepath.Join(project.RootDir, "chart"), "--namespace "+namespace, "--set=image="+image, extraArg)).
+				To(Succeed(), "Operator failed to be deployed")
+		}
+
+		Eventually(common.GetObject).WithArguments(ctx, cl, kube.Key(deploymentName, namespace), &appsv1.Deployment{}).
+			Should(HaveCondition(appsv1.DeploymentAvailable, metav1.ConditionTrue), "Error getting Istio CRD")
+		Success("Operator is deployed in the namespace and Running")
+	})
+
+	Describe("using supported Istio version", func() {
```
fjglira, line 66: `using supported Istio version`  we should be putting information in the Describe, Context, etc that gives clear information about what will happen in that node or step of the test. So, this Describe should be something like: `configuration for supported versions`

tests/e2e/dualstack/dualstack_test.go
```diff
@@ -0,0 +1,213 @@
...
+		for _, version := range supportedversion.List {
+			// Note: This var version is needed to avoid the closure of the loop
+			version := version
+
+			// The minimum supported version is 1.23 (and above)
+			if version.Major == 1 && version.Minor < 23 {
+				continue
+			}
+
+			Context(version.Name, func() {
```
fjglira, line 76: And then here: `"Istio version is: "+version.Version`

tests/e2e/dualstack/dualstack_test.go
```diff
@@ -0,0 +1,213 @@
...
+				BeforeAll(func() {
+					Expect(k.CreateNamespace(controlPlaneNamespace)).To(Succeed(), "Istio namespace failed to be created")
+				})
+
+				When("the Istio CR is created", func() {
```
fjglira, line 81: "the Istio CR is created with DualStack configuration"

tests/e2e/dualstack/dualstack_test.go
```diff
@@ -0,0 +1,213 @@
...
+						common.CheckNamespaceEmpty(ctx, cl, controlPlaneNamespace)
+						Success("Namespace is empty")
+					})
+				})
+			})
+		}
+
+		AfterAll(func(ctx SpecContext) {
+			if CurrentSpecReport().Failed() {
+				common.LogDebugInfo()
```
fjglira, line 169: We will need to add more debug information for when DualStack is set. If the answer is yes, we can add some specific information there

tests/e2e/dualstack/dualstack_suite_test.go
```diff
@@ -0,0 +1,65 @@
...
+	skipDeploy            = env.GetBool("SKIP_DEPLOY", false)
+	expectedRegistry      = env.Get("EXPECTED_REGISTRY", "^docker\\.io|^gcr\\.io")
+	multicluster          = env.GetBool("MULTICLUSTER", false)
+	ipFamily              = env.Get("IP_FAMILY", "ipv4")
+
+	k *kubectl.KubectlBuilder
+)
+
+func TestDualStack(t *testing.T) {
+	if ipFamily != "dual" || multicluster {
```
fjglira, line 48: Ok, this is only to be executed when dualStack is being set. But we will need to skip the other tests when ipFamily is set to dualStack also. Maybe it is worth starting to use the labels in our test?

fjglira (commented):
I put some comments. But checking deeply I'm think if worth to have a complete separated test for dualstack? Checking the entire test run the steps are the same one than the regular control plane installation besides two or three differents steps, so, maybe worth to add those steps inside the current test and skip those steps when dualstack is not being set? wdyt?

tests/e2e/dualstack/dualstack_suite_test.go
sridhargaddam: IMO, we can continue to run the other test suites even when ipFamily is set to dualStack. Besides validating dualStack on a dualStack cluster, one of our use cases was to ensure that Istio runs successfully in IPv4 mode, even on a dualStack cluster. By running the other test suites, we can regularly get feedback on the status.

tests/e2e/dualstack/dualstack_test.go
sridhargaddam: The current information that is logged is good for troubleshooting any issues.

tests/e2e/dualstack/dualstack_suite_test.go
fjglira: Ok, that sounds good

tests/e2e/dualstack/dualstack_test.go
```diff
@@ -0,0 +1,213 @@
...
+	"github.com/istio-ecosystem/sail-operator/tests/e2e/util/helm"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/types"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ = Describe("Installation on a dualStack cluster", Ordered, func() {
```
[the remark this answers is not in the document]
sridhargaddam: The output of the test (i.e., report.xml) would report the following with the existing code.

```
<testcase name="[It] Installation on a dualStack cluster using supported Istio version v1.23.0 when the Istio CR is created updates the Istio CR status to Reconciled" classname="DualStack test suite" status="passed" time="13.197685198">
```

If I update the Describe("DualStack configuration"), it will be as follows.

```
<testcase name="[It] DualStack configuration configuration for supported versions Istio version is: 1.23.0 when the Istio CR is created with DualStack configuration updates the Istio CR status to Reconciled" classname="DualStack test suite" status="passed" time="23.52515575">
```
Even if I remove the duplicated "configuration" word, I feel the existing text is more readable. Please let me know what you think?

tests/e2e/dualstack/dualstack_test.go
sridhargaddam: please see my previous comment.

tests/e2e/dualstack/dualstack_test.go
sridhargaddam: sure, will update it.

tests/e2e/dualstack/dualstack_test.go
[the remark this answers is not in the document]
fjglira: I agree that xml looks better like this, but the idea also is to be able to run `make test.e2e.describe` and see the BDD way test description like this
```
Describe,Multicluster deployment models,1626,16408,false,false,false,""
    Describe,Multi-Primary Multi-Network configuration,3194,15925,false,false,false,""
            When,Istio resources are created in both clusters with multicluster configuration,3458,7240,false,false,false,""
                It,updates both Istio CR status to Ready,5574,6253,true,false,false,""
                It,deploys istiod,6260,7233,true,false,false,""
            When,Gateway is created in both clusters,7246,8741,false,false,false,""
                It,updates both Gateway status to Available,7999,8734,true,false,false,""
            When,are installed remote secrets on each cluster,8747,10722,false,false,false,""
                It,remote secrets are created,10086,10715,true,false,false,""
            When,sample apps are deployed in both clusters,10728,13884,false,false,false,""
                It,updates the pods status to Ready,10986,12174,true,false,false,""
                It,can access the sample app from both clusters,12181,13877,true,false,false,""
            When,istio CR is deleted in both clusters,13890,14883,false,false,false,""
                BeforeEach,,13949,14354,false,false,false,""
                It,removes istiod pod,14361,14876,true,false,false,""
```

The Describe, Context, When and it nodes needs to make sense to be able to share this information like this also

sridhargaddam:
> I put some comments. But checking deeply I'm think if worth to have a complete separated test for dualstack? Checking the entire test run the steps are the same one than the regular control plane installation besides two or three differents steps, so, maybe worth to add those steps inside the current test and skip those steps when dualstack is not being set? wdyt?

Thanks @fjglira for reviewing. Yes, currently this looks very similar to the regular control_plane tests (with some minor differences), but I'm also working on adding data-path tests to this test-suite where we try to validate datapath use-cases like a client is able to connect to IPv4/IPv6/dualStack services. So, having a separate dualStack testsuite would be readable and will allow adding any dualStack specific tests without cluttering the existing test suite. Also, as I mentioned in my [comment](https://github.com/istio-ecosystem/sail-operator/pull/373#discussion_r1781323531), we want the existing control_plane suite to be validated (i.e., Istio deployed with defaults) along with dualStack suite (where Istio will be deployed in dualStack mode).

fjglira:
> > I put some comments. But checking deeply I'm think if worth to have a complete separated test for dualstack? Checking the entire test run the steps are the same one than the regular control plane installation besides two or three differents steps, so, maybe worth to add those steps inside the current test and skip those steps when dualstack is not being set? wdyt?
> 
> Thanks @fjglira for reviewing. Yes, currently this looks very similar to the regular control_plane tests (with some minor differences), but I'm also working on adding data-path tests to this test-suite where we try to validate datapath use-cases like a client is able to connect to IPv4/IPv6/dualStack services. So, having a separate dualStack testsuite would be readable and will allow adding any dualStack specific tests without cluttering the existing test suite. Also, as I mentioned in my [comment](https://github.com/istio-ecosystem/sail-operator/pull/373#discussion_r1781323531), we want the existing control_plane suite to be validated (i.e., Istio deployed with defaults) along with dualStack suite (where Istio will be deployed in dualStack mode).

Ok, this make a lot of sense also. So, ok for me then

tests/e2e/dualstack/dualstack_test.go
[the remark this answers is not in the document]
sridhargaddam: Okay, let me update it.

tests/e2e/dualstack/dualstack_test.go
sridhargaddam: updated and made some minor change.

fjglira (approved):
LGTM

PR #383, opened 2024-10-04
Implement e2e dualStack data plane tests

sridhargaddam:
This PR validates the following use-cases

It spawns a
1. dualStack service in dual-stack namespace which listens on both IPv4 and IPv6 addresses
2. ipv4 only service in ipv4 namespace which listens only on IPv4 address
3. ipv6 only service in ipv6 namespace which listens only on IPv6 address

After spawning the above pods, it uses a sleep pod and tries to connect to all the three services to verify that connectivity is successful for all the services.

To run the tests, use the following command.
$ export IP_FAMILY=dual
$ make test.e2e.kind

### DIFF
--- a/tests/e2e/dualstack/dualstack_test.go
+++ b/tests/e2e/dualstack/dualstack_test.go
@@ -35,6 +35,14 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	DualStackNamespace = "dual-stack"
+	IPv4Namespace      = "ipv4"
+	IPv6Namespace      = "ipv6"
+	SleepNamespace     = "sleep"
 )
 
 var _ = Describe("DualStack configuration ", Ordered, func() {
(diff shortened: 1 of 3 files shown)


tests/e2e/dualstack/dualstack_test.go
```diff
@@ -148,6 +184,68 @@ spec:
...
+						Expect(k.Patch("namespace", DualStackNamespace, "merge", `{"metadata":{"labels":{"istio-injection":"enabled"}}}`)).
+							To(Succeed(), "Error patching dual-stack namespace")
+						Expect(k.Patch("namespace", IPv4Namespace, "merge", `{"metadata":{"labels":{"istio-injection":"enabled"}}}`)).
+							To(Succeed(), "Error patching ipv4 namespace")
+						Expect(k.Patch("namespace", IPv6Namespace, "merge", `{"metadata":{"labels":{"istio-injection":"enabled"}}}`)).
+							To(Succeed(), "Error patching ipv6 namespace")
+						Expect(k.Patch("namespace", SleepNamespace, "merge", `{"metadata":{"labels":{"istio-injection":"enabled"}}}`)).
+							To(Succeed(), "Error patching sleep namespace")
+
+						deployDualStackValidationPods(version)
```
fjglira, line 208: I'm not a big fan of hiding this; in the end, it is only deploying the pods, It is easy to read if the creation of the pods is here directly. But if you want to leave it like this, I'm ok with that

tests/e2e/dualstack/dualstack_test.go
```diff
@@ -211,3 +324,59 @@ func ImageFromRegistry(regexp string) types.GomegaMatcher {
 func getEnvVars(container corev1.Container) []corev1.EnvVar {
 	return container.Env
 }
+
+func getPodURL(version supportedversion.VersionInfo, namespace string) string {
```
fjglira, line 328: Maybe change the func name to getYAMLPodURL? Calling this func getPodURL may be confusing

tests/e2e/dualstack/dualstack_test.go
```diff
@@ -211,3 +324,59 @@ func ImageFromRegistry(regexp string) types.GomegaMatcher {
...
+}
+
+func deployDualStackValidationPods(version supportedversion.VersionInfo) {
+	Expect(k.SetNamespace(DualStackNamespace).Apply(getPodURL(version, DualStackNamespace))).To(Succeed(), "error deploying tcpDualStack pod")
+	Expect(k.SetNamespace(IPv4Namespace).Apply(getPodURL(version, IPv4Namespace))).To(Succeed(), "error deploying ipv4 pod")
+	Expect(k.SetNamespace(IPv6Namespace).Apply(getPodURL(version, IPv6Namespace))).To(Succeed(), "error deploying ipv6 pod")
+	Expect(k.SetNamespace(SleepNamespace).Apply(getPodURL(version, SleepNamespace))).To(Succeed(), "error deploying sleep pod")
+}
+
+func checkPodsReady(ctx SpecContext, namespace string) (*corev1.PodList, error) {
```
fjglira, line 358: I think this function is very useful; I think we can move it to the kubectl util and use it in all the tests because in other tests, we also validate that all the pods in the ns are Ready

fjglira (commented):
LGTM, I added some minor comments.
Besides those comments, will you create also the job to run this test here in the sail repo? Are you planning to add more test related to dualStack?

FilipB:
Do you know if there are any dual stack tests in core istio/istio?

fjglira:
> Do you know if there are any dual stack tests in core istio/istio?

Yes, some of the ambient and integration tests set IP_FAMILY to dual, for example: https://github.com/istio/test-infra/blob/638d34a5abd39033166230f10f1a973a832ded95/prow/config/jobs/istio.yaml#L123

sridhargaddam:
> LGTM, I added some minor comments. Besides those comments, will you create also the job to run this test here in the sail repo? 

Yes @fjglira. I'm working on it. 

> Are you planning to add more test related to dualStack?

No, we are good. I basically automated the [following validation](https://istio.io/latest/docs/setup/additional-setup/dual-stack/) in the e2e tests.

tests/e2e/dualstack/dualstack_test.go
fjglira: Sure, moving into `common` sounds better

fjglira (approved):

PR #385, opened 2024-10-07
Add Helm artifacts publish automation

### LINKED ISSUE #367: Automate helm chart artifacts reference publish on release

### Feature Description

As per now, Sail Operator helm charts could be installed by using `helm repo add helm-sail-operator https://istio-ecosystem.github.io/sail-operator/` flow without a need to clone the repository.
In `gh-pages` branch, an "index.yaml" file holds a reference to the helm artifact attached to a release asset.

The next step is to automate "index.yaml" file update during each release, so it will always contain the reference to the new helm artifact version.

### Additional Information

_No response_

MaxBab:
#### What type of PR is this?

- [X] Enhancement / New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [ ] Test
- [ ] Documentation Update

When a release being created, the helm artifacts stored as an assets of the release.
In order to make those artifacts available for "helm repo add" flow, an "index.yaml" file under "gh-pages" branch needs to be updated.

Once updated, it will serve the helm deployment method with the new artifacts version.

Add an automation flow to github actions that could be triggered and will make the required updates.
At the end, a pull request with the changes will be created.

Fixes #367

### DIFF
--- a/.github/workflows/helm.yaml
+++ b/.github/workflows/helm.yaml
@@ -1,0 +1,32 @@
+name: Publish Helm artifact
+
+on:
+  workflow_dispatch:
+    inputs:
+      release_version:
+        description: "Release version"
+        required: true
+
+run-name: Publish Helm artifact ${{ inputs.release_version }}
+
+env:
+  GIT_USER: ${{ secrets.GIT_USER }}
+  GITHUB_TOKEN: ${{ secrets.GIT_TOKEN }}
+  VERSION: ${{ inputs.release_version }}
+
+jobs:
+  helm:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Publish Helm artifact to "gh-pages" branch
+        run: |
+          make helm-artifacts-publish \
+            -e GIT_CONFIG_USER_NAME="$GIT_CONFIG_USER_NAME" \
+            -e GIT_CONFIG_USER_EMAIL="$GIT_CONFIG_USER_EMAIL" \
+            -e OPERATOR_VERSION=$VERSION
+        env:
+          GIT_CONFIG_USER_NAME: "${{ github.actor }}"
+          GIT_CONFIG_USER_EMAIL: "${{ github.actor_id }}+${{ github.actor }}@users.noreply.github.com"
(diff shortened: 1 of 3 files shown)


MaxBab:
Currently, the github action job has been created with `workflow_dispatch`, which means it has the manual trigger.

I did it as per now, because I need to deeply check the way to automation trigger of the job, when a release is being published.
The key point to check is that a workflow could not be triggered automatically by a release trigger [1], in case a release job has a manual trigger `workflow_dispatch` [2].

I'll check and test it separately, and update the configuration to add a support for automation trigger by release.

[1] - https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#release
[2] - https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/triggering-a-workflow#triggering-a-workflow-from-a-workflow

fjglira:
There is any way of testing this before merged?

MaxBab:
Yes.
I've tested it.
I've cloned the sail-operator repo and uploaded under my account with a different name.

There I've done all the testing.
This flow works.

.github/workflows/helm.yaml
```diff
@@ -0,0 +1,32 @@
+name: Publish Helm artifact
+
+on:
+  workflow_dispatch:
```
dgn, line 4: we could add a `[on.release](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#release)`, right? then it would be triggered automatically.

```
on:
  release:
    types: [published]
```

MaxBab: @dgn 
Not exactly.

As I mentioned in the comment above, it depends on the type of GITHUB TOKEN being used.
I've added link, which explains that if the workflow in using the default GITHUB TOKEN, which is being generated by the workflow for each execution and the workflow itself is using a `workflow_dispatch` jobs type, the other workflow job will not be triggered even if it set to check the published release.

I've tested it and it indeed does not work.

In the same github docs, there is an explanation that it could be workarounded by using a custom create GITHUB TOKEN.
And I'm going to test it.
But I prefer to add the existing workflow with the manual trigger as per now, to provide the option to publish the helm artifacts to helm repo.

Will check the complete automation flow as a next step.

dgn: ah. we already have a GIT_TOKEN secret that you could use. it's the token of our bot account, openshift-service-mesh-bot

dgn (approved):

.github/workflows/helm.yaml
MaxBab: Yes, I know that we're using a custom token.
I saw it.

I just need to catch Brian and ask him regarding exact config of this token.
So I could simulate and properly test its flow.

PR #386, opened 2024-10-07
Update Istio versions

luksa:


dgn:
This one's interesting. Do we hard-code the version somewhere or are we failing to parse the trailing . version?
```
  [FAILED] Unexpected istiod version
  Expected
      <string>: 1.24-alpha.6
  to equal
      <string>: 1.24-alpha 
```

PR #391, opened 2024-10-08
Simplify supportedversion.go and add unit tests

luksa:
These tests will not just ensure that the functions in supportedversion.go are okay, but also that the versions.yaml file is valid and is parsed successfully.


dgn (approved):

dgn:
Maybe we'll need to update the update version script as well to include the full version (including .commit_hash suffix)?

PR #394, opened 2024-10-08
Improve kubectl utility in e2e tests

luksa:
The main purpose of these changes is to prevent bugs when kubectl.Builder (now called kubectl.Kubectl) is used in multiple places. Previously, if you called `SetNamespace()`, you had to also call `ResetNamespace()` afterwards so that the next invocation of the tool defaults back to no namespace. This also means that the tool wasn't thread-safe.

### DIFF
--- a/tests/e2e/util/certs/certs.go
+++ b/tests/e2e/util/certs/certs.go
@@ -228,7 +228,7 @@ func PushIntermediateCA(ns, kubeconfig, zone, network, basePath string, cl clien
 	if err != nil {
 		// Label the namespace with the network
 		k := kubectl.NewBuilder()
-		k.SetKubeconfig(kubeconfig)
+		k.WithKubeconfig(kubeconfig)
 		err = k.Patch("namespace", ns, "merge", `{"metadata":{"labels":{"topology.istio.io/network":"`+network+`"}}}`)
 		if err != nil {
 			return fmt.Errorf("failed to label namespace: %w", err)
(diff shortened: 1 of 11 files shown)


fjglira (approved):

tests/e2e/util/certs/certs.go
```diff
@@ -227,8 +227,8 @@ func PushIntermediateCA(ns, kubeconfig, zone, network, basePath string, cl clien
 	_, err := common.GetObject(context.Background(), cl, kube.Key("cacerts", ns), &corev1.Secret{})
 	if err != nil {
 		// Label the namespace with the network
-		k := kubectl.NewKubectlBuilder()
-		k.SetKubeconfig(kubeconfig)
+		k := kubectl.New()
+		k.WithKubeconfig(kubeconfig)
```
sridhargaddam, line 231: I think we have an issue here. This should be
```
k = k.WithKubeconfig(kubeconfig)
```
or even better, we can remove that statement itself and modify the next line to
```
err = k.WithKubeconfig(kubeconfig).Patch(....)
```

sridhargaddam (commented):
Thanks @luksa. This looks much cleaner.

tests/e2e/util/certs/certs.go
luksa: AAAaaaa, you're right. This is what's causing the test to fail. I've been trying to figure this out for the past two hours. 

Remind me to buy you a drink at the next F2F :)

sridhargaddam: `WithKubeconfig()` is not only returning `k` but also updating the `k.kubeconfig` which is finally used in the actual commands, so we are good. Sure, you can update it in a different PR.

sridhargaddam (approved):

PR #397, opened 2024-10-09
Generate asciidoc API reference

bmangoen:
Add the possibility to generate API reference in asciidoc format (example of generated [sailoperator.io.asciidoc](https://gist.github.com/bmangoen/f744019fd12cad940d37b1b6be2bd377#file-sailoperato-io-asciidoc) file)

Related Issue [OSSM-6856](https://issues.redhat.com/browse/OSSM-6856)

### DIFF
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -395,7 +395,8 @@ gen-check: gen restore-manifest-dates check-clean-repo ## Verify that changes in
 CRD_PATH := ./api
 OUTPUT_DOCS_PATH := ./docs/api-reference
 CONFIG_API_DOCS_GEN_PATH := ./hack/api-docs/config.yaml
-TEMPLATES_DIR := ./hack/api-docs/templates/markdown
+DOCS_RENDERER := markdown
+TEMPLATES_DIR := ./hack/api-docs/templates/$(DOCS_RENDERER)
 
 gen-api-docs: ## Generate API documentation. Known issues: go fmt does not properly handle tabs and add new line empty. Workaround is applied to the generated markdown files. The crd-ref-docs tool add br tags to the generated markdown files. Workaround is applied to the generated markdown files.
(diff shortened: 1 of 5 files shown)


fjglira:
Hey @bmangoen , do you know if this will be rendered Ok later in the docs? I think it is better to tune the replacement section to get rid of this also to avoid wrong rendering later
image

bmangoen:
> Hey @bmangoen , do you know if this will be rendered Ok later in the docs? I think it is better to tune the replacement section to get rid of this also to avoid wrong rendering later image

Good question. 
I see different rendering on gist.github.com and on the asciidoc preview on my IDE (which seems correct).

Maybe @rh-tokeefe @gwynnemonahan can help us with?

Makefile.core.mk
```diff
@@ -405,12 +406,12 @@ gen-api-docs: ## Generate API documentation. Known issues: go fmt does not prope
 		--source-path=$(CRD_PATH) \
 		--templates-dir=$(TEMPLATES_DIR) \
 		--config=$(CONFIG_API_DOCS_GEN_PATH) \
-		--renderer=markdown \
+		--renderer=$(DOCS_RENDERER) \
 		--output-path=$(OUTPUT_DOCS_PATH) \
 		--output-mode=group
 	@find $(OUTPUT_DOCS_PATH) -type f -name "*.md" -exec sed -i 's/<br \/>/ /g' {} \;
 	@find $(OUTPUT_DOCS_PATH) -type f -name "*.md" -exec sed -i 's/\t/  /g' {} \;
```
dgn, line 413: it looks like we will also need to replace the `\t`'s in the asciidoc, right?

bmangoen: I just checked and you're right ! Fixing it

dgn (approved):

PR #399, opened 2024-10-09
Use upstream builder image in update-deps workflow

### LINKED ISSUE #390: Move automator jobs to GitHub Actions
istio-testing bot does not have permissions for changing workflows, which breaks us. If we can't get the token updated, we'll need a short-term replacement, best would probably be to run the jobs on GH actions

dgn:
Turns out we need the upstream image in order for the pr-creator tool to be present.

Sorry about this one. This time I tested end-to-end on my fork.

Fixes #390

### DIFF
--- a/.github/workflows/update-deps.yaml
+++ b/.github/workflows/update-deps.yaml
@@ -22,12 +22,18 @@ env:
 jobs:
   update-deps:
     runs-on: ubuntu-latest
+    container:
+      image: gcr.io/istio-testing/build-tools:master-8eb42e9551b9a67c330aeef783f2498647d91289
+      options: --entrypoint ''
 
     steps:
     - uses: actions/checkout@v4
       with:
         repository: istio/test-infra
         ref: master
+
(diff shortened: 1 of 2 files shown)


fjglira (approved):

.github/workflows/update-deps.yaml
```diff
@@ -22,13 +22,19 @@ env:
...
+      options: --entrypoint ''
 
     steps:
     - uses: actions/checkout@v4
       with:
         repository: istio/test-infra
         ref: master
 
+    # this is a workaround for a permissions issue when using the istio build container
+    - run: git config --system --add safe.directory /__w/sail-operator/sail-operator
```
jwendell, line 36: is this directory correct?

jwendell (approved):

.github/workflows/update-deps.yaml
dgn: yeah, not sure why it's sail-operator/sail-operator, but it's correct - and without it, the action fails

PR #413, opened 2024-10-11
[DRAFT] Revision Tag Support

### LINKED ISSUE #380: Revision Tag support
Design currently in review in https://github.com/istio-ecosystem/sail-operator/pull/212

Next steps:
- [ ] Finalize SEP
- [ ] alpha implementation

dgn:
Fixes #380


controllers/istiorevision/istiorevision_controller.go
```diff
@@ -225,6 +226,10 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 	// The handler triggers the reconciliation of the referenced IstioRevision CR so that its InUse condition is updated.
 	podHandler := wrapEventHandler(logger, handler.EnqueueRequestsFromMapFunc(r.mapPodToReconcileRequest))
 
+	// revisionTagHandler handles IstioRevisionTags that reference the IstioRevision CR via their targetRef.
+	// The handler triggers the reconciliation of the referenced IstioRevision CR so that its InUse condition is updated.
+	revisionTagHandler := enqueuelogger.WrapIfNecessary(v1alpha1.IstioRevisionKind, logger, handler.EnqueueRequestsFromMapFunc(r.mapRevisionTagToReconcileRequest))
```
luksa, line 231:
```suggestion
	revisionTagHandler := wrapEventHandler(logger, handler.EnqueueRequestsFromMapFunc(r.mapRevisionTagToReconcileRequest))
```

(the wrapEventHandler function was added just recently)

controllers/istiorevision/istiorevision_controller.go
```diff
@@ -409,6 +417,18 @@ func (r *Reconciler) isRevisionReferencedByWorkloads(ctx context.Context, rev *v
 	log := logf.FromContext(ctx)
 	nsList := corev1.NamespaceList{}
 	nsMap := map[string]corev1.Namespace{}
+	// if we're referenced by an in-use revisionTag, we're done. we can outsource the checking to the IstioRevisionTagController
+	revisionTagList := v1alpha1.IstioRevisionTagList{}
+	if err := r.Client.List(ctx, &revisionTagList); err != nil {
+		return false, fmt.Errorf("failed to list IstioRevisionTags: %w", err)
+	}
+	for _, tag := range revisionTagList.Items {
+		if tag.Status.IstioRevision == rev.Name && tag.Status.GetCondition(v1alpha1.IstioRevisionTagConditionInUse).Status == metav1.ConditionTrue {
```
luksa, line 426: Is it correct to check whether the tag is in use? As a user, if I create a tag and point it to a revision, I don't want that revision to be deleted by the operator. It doesn't matter if there are no pods/namespaces referencing that tag.

Also, is it correct to check `status.istioRevision` instead of `spec.targetRef`? What if the spec was just updated to point to the revision, but the controller hasn't yet updated the status?

luksa:
You also need to add `IstioRevisionTag` to the `PROJECT` file. I believe some make target(s) only update certain files if the CRD is listed in the PROJECT file.

controllers/istiorevision/istiorevision_controller.go
dgn: > Is it correct to check whether the tag is in use? As a user, if I create a tag and point it to a revision, I don't want that revision to be deleted by the operator. It doesn't matter if there are no pods/namespaces referencing that tag.

I thought about that, too. Yeah, maybe referencing a revision in a tag should be treated just like a reference from a pod or namespace and prevent deletion. I can implement that.

> Also, is it correct to check status.istioRevision instead of spec.targetRef? What if the spec was just updated to point to the revision, but the controller hasn't yet updated the status?

Well, the status shows what is actually referenced ie what the webhooks will inject. So I would consider that to take precedent over what is in the spec. Apart from that, it's not obvious from the spec which IstioRevision is actually referenced, eg if the targetRef points to an Istio resource. I think adding this logic here would mean the IstioRevision controller knowing too many technical details about IstioRevisionTag. But I'm happy to be convinced otherwise.

controllers/istiorevision/istiorevision_controller.go
```diff
@@ -409,6 +417,18 @@ func (r *Reconciler) isRevisionReferencedByWorkloads(ctx context.Context, rev *v
 	log := logf.FromContext(ctx)
 	nsList := corev1.NamespaceList{}
 	nsMap := map[string]corev1.Namespace{}
+	// if we're referenced by an in-use revisionTag, we're done. we can outsource the checking to the IstioRevisionTagController
```
nrfox, line 420: Is the revision tag controller also checking pods? What if you have created a tag but are still using the rev label and not the tag name to label your namespace?

nrfox: The only thing the tag adds, if it exists, is an additional condition for the namespace. Now it could be either `istio.io/rev: <rev>` or `istio.io/rev: <tag>`.

api/v1alpha1/istiorevisiontags_types.go
```diff
@@ -0,0 +1,199 @@
...
+	//
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	Kind string `json:"kind"`
+
+	// Name is the name of the target resource.
+	//
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	Name string `json:"name"`
```
nrfox, line 46: Should this be `required` along with `Kind`?

controllers/istiorevisiontag/istiorevisiontag_controller.go
```diff
@@ -0,0 +1,522 @@
...
+			log.V(2).Info("Revision is referenced by Namespace", "Namespace", ns.Name)
+			return true, nil
+		}
+		nsMap[ns.Name] = ns
+	}
+
+	podList := corev1.PodList{}
+	if err := r.Client.List(ctx, &podList); err != nil { // TODO: can we optimize this by specifying a label selector
+		return false, fmt.Errorf("failed to list pods: %w", err)
+	}
```
nrfox, line 373: I wonder if this controller should be checking for pods at all or if it can defer to the `istiorevision` controller for that which is already checking for the pod annotation that is independent of tag? Basically if this controller's istiorev ref is in use then this tag should be considered in use?

nrfox: Main question I have is this ^^. Rest is basically nits.

bundle/manifests/sailoperator.clusterserviceversion.yaml
```diff
@@ -217,6 +217,12 @@ spec:
         displayName: Helm Values
         path: values
       version: v1alpha1
+    - description: IstioRevisionTag references a Istio or IstioRevision object and
```
FilipB, line 220: Why it is possible to reference Istio resource? Shouldn't the tag only reference the IstioRevision? What happens when there are more revisions for given istio? e.g. during the upgrade?

resources/v1.23.2/charts/revisiontags/templates/revision-tags.yaml
```diff
@@ -0,0 +1,141 @@
...
+  objectSelector:
+    matchExpressions:
+    - key: sidecar.istio.io/inject
+      operator: In
+      values:
+      - "true"
+    - key: istio.io/rev
+      operator: DoesNotExist
+
+{{- if $.Values.sidecarInjectorWebhook.enableNamespacesByDefault }}
```
FilipB, line 119: Do we want to support this case? It is one more case to test and it might never be used by customers.

tests/integration/api/istiorevisiontag_test.go
```diff
@@ -0,0 +1,217 @@
...
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"istio.io/istio/pkg/ptr"
+)
+
+var _ = Describe("IstioRevisionTag resource", Ordered, func() {
```
FilipB, line 35: Do we want to cover also other injection cases like pod level labeling and other combinations defined in the MutatingWebhookConfiguration? Also some negative cases e.g.:
- user creates `default` Istio CR (InPlace strategy) and then creates `default` IstioRevisionTag -> it should fail with reasonable error
- what about cases when the user is using `istioctl` to create tags?  Do we want to prevent that?

bundle/manifests/sailoperator.clusterserviceversion.yaml
FilipB: Ah, ok. Answer is in the SEP document

api/v1alpha1/istiorevisiontags_types.go
```diff
@@ -0,0 +1,199 @@
...
+	Conditions []IstioRevisionTagCondition `json:"conditions,omitempty"`
+
+	// Reports the current state of the object.
+	State IstioRevisionTagConditionReason `json:"state,omitempty"`
+
+	// IstiodNamespace stores the namespace of the corresponding Istiod instance
+	IstiodNamespace string `json:"istiodNamespace"`
+
+	// IstioRevision stores the name of the referenced IstioRevision
+	IstioRevision string `json:"istioRevision"`
```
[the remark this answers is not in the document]
dgn: no, actually the reasoning behind this field is exactly the fact that we could be referencing an Istio resource. The IstioRevision controller needs some easy way to know which IstioRevision is being referenced (even if technically it is referencing an Istio, it will always resolve to an IstioRevision), so instead of duplicating the resolution logic behind that, I'm just doing it at reconcile time in the IstioRevisionTag controller and write the result into this field

tests/integration/api/istiorevisiontag_test.go
dgn: * I'll add a case for pod labelling
* also good point about default revision vs revisiontag, I'll cover that
* I'm not sure we can prevent istioctl usage

resources/v1.23.2/charts/revisiontags/templates/revision-tags.yaml
dgn: I guess? I don't know what the support level for it is upstream tbh

controllers/istiorevisiontag/istiorevisiontag_controller.go
dgn: hmm. yes that is an optimization, at least for the pods. for the namespaces we'd still need to check

api/v1alpha1/istiorevisiontags_types.go
dgn: done! thanks

controllers/istiorevision/istiorevision_controller.go
dgn: actually... I was previously checking the pods but that doesn't make sense, as their annotation will always have the actual revision name, not the tag. so we don't need to check pods in the revisiontag controller

dgn: okay I removed the annotation checking, but I'm now checking the label to identify injection intent

dgn:
Okay I refactored (and simplified) the InUse detection quite a bit and I'm going to outline it here (will also update the design doc accordingly):

An IstioRevisionTag is considered InUse when
- there's a pod or namespace that explicitly references the IstioRevisionTag in an `istio.io/rev` label
- the name of the IstioRevisionTag name is 'default' and there's a pod or namespace with the `istio-injection: enabled` label

Note that a pod's `istio.io/rev` __annotation__ will not be considered as that will always have the name of the referenced IstioRevision rather than the name of the IstioRevisionTag! The labels however are added by users and reflect __usage intent__, ie the user will use the name of the IstioRevisionTag.

Even if the referenced IstioRevision of an IstioRevisionTag is considered InUse, that does not suffice to make the IstioRevisionTag considered InUse by the operator! It is considered an unused alias for an InUse IstioRevision.

An IstioRevision is considered InUse when (new parts in __bold__)
- there's a pod that explicitly references the IstioRevision in an `istio.io/rev` annotation __or label__
- there's a namespace that explicitly references the IstioRevision in an `istio.io/rev` label
- the name of the IstioRevision name is 'default' and there's a pod or namespace with the `istio-injection: enabled` label
- __there's an IstioRevisionTag referencing this IstioRevision__

Note specifically the third item here: being referenced by an IstioRevisionTag will now always lead to being considered InUse! __It does not matter if the IstioRevisionTag is itself considered InUse!__

I think this is the cleanest approach with the least amount of circular dependencies.

nrfox (commented):
When I first create the sample this is what I get:
```
kubectl get istios
NAME      REVISIONS   READY   IN USE   ACTIVE REVISION   STATUS    VERSION   AGE
default   1           1       1        default-v1-23-2   Healthy   v1.23.2   113s

kubectl get istiorevisions.sailoperator.io
NAME              TYPE    READY   STATUS    IN USE   VERSION   AGE
default-v1-23-2   Local   True    Healthy   True     v1.23.2   2m5s

kubectl get istiorevisiontags.sailoperator.io
NAME      STATUS                    IN USE   REVISION          AGE
default   NotReferencedByAnything   False    default-v1-23-2   114s
```

Notice that the `Istio` and `IstioRevision` are both `IN USE = true` but the tag is not? That rightfully changes when I deploy a sample workload that uses the tag but the istio/rev both seem to have a different definition of `IN USE` than the tag does.

edit: Re-reading the above comment outlining when each resource is considered `IN USE`, this is actually the expected behavior and I think makes sense.

edit2: I wonder if it would help to have short human readable message for the istiorevtag printer columns like "Istio Tag is not referenced by a pod or a namespace." to make it more clear why the tag is considered not `IN USE`?

docs/api-reference/sailoperator.io.md
```diff
@@ -877,6 +879,163 @@ _Appears in:_
...
+
+| Field | Description |
+| --- | --- |
+| `NameAlreadyExists` | IstioRevisionTagNameAlreadyExists indicates that the a revision with the same name as the IstioRevisionTag already exists.  |
+| `RefNotFound` | IstioRevisionTagReasonReferenceNotFound indicates that the resource referenced by the tag's TargetRef was not found  |
+| `ReconcileError` | IstioRevisionReasonReconcileError indicates that the reconciliation of the resource has failed, but will be retried.  |
+| `ReferencedByWorkloads` | IstioRevisionReasonReferencedByWorkloads indicates that the revision is referenced by at least one pod or namespace.  |
+| `NotReferencedByAnything` | IstioRevisionReasonNotReferenced indicates that the revision is not referenced by any pod or namespace.  |
+| `UsageCheckFailed` | IstioRevisionReasonUsageCheckFailed indicates that the operator could not check whether any workloads use the revision.  |
+| `Healthy` | IstioRevisionReasonUsageCheckFailed indicates that the operator could not check whether any workloads use the revision.  |
```
nrfox, line 945: Description for `Healthy` looks like it's the same for `UsageCheckFailed`.

tests/integration/api/istiorevisiontag_test.go
```diff
@@ -0,0 +1,355 @@
...
+	const (
+		defaultTagName            = "default"
+		prodTagName               = "prod"
+		istioName                 = "test-istio"
+		istioRevisionTagNamespace = "istiorevisiontag-test"
+		workloadNamespace         = "istiorevisiontag-test-workloads"
+
+		gracePeriod = 5 * time.Second
+	)
+	// istioKey := client.ObjectKey{Name: istioName}
```
nrfox, line 46: Can remove this?
```suggestion
```

tests/integration/api/istiorevisiontag_test.go
```diff
@@ -0,0 +1,355 @@
...
+	istio := &v1alpha1.Istio{}
+	istioKey := client.ObjectKey{Name: istioName}
+	defaultTagKey := client.ObjectKey{Name: defaultTagName}
+	workloadNamespaceKey := client.ObjectKey{Name: workloadNamespace}
+	// prodTagKey := client.ObjectKey{Name: prodTagName}
```
nrfox, line 51: Can remove this?
```suggestion
```

api/v1alpha1/istiorevisiontags_types.go
```diff
@@ -0,0 +1,204 @@
...
+
+	// IstioRevisionReasonNotReferenced indicates that the revision is not referenced by any pod or namespace.
+	IstioRevisionTagReasonNotReferenced IstioRevisionTagConditionReason = "NotReferencedByAnything"
+
+	// IstioRevisionReasonUsageCheckFailed indicates that the operator could not check whether any workloads use the revision.
+	IstioRevisionTagReasonUsageCheckFailed IstioRevisionTagConditionReason = "UsageCheckFailed"
+)
+
+const (
+	// IstioRevisionReasonUsageCheckFailed indicates that the operator could not check whether any workloads use the revision.
```
sridhargaddam, line 172:
```suggestion
	// IstioRevisionTagReasonHealthy indicates that the operator could not check whether any workloads use the revision.
```

controllers/istiorevisiontag/istiorevisiontag_controller.go
```diff
@@ -0,0 +1,541 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"istio.io/istio/pkg/ptr"
+)
+
+const (
+	IstioInjectionLabel        = "istio-injection"
+	IstioInjectionEnabledValue = "enabled"
+	IstioRevLabel              = "istio.io/rev"
+	IstioSidecarInjectLabel    = "sidecar.istio.io/inject"
```
sridhargaddam, line 57: We can probably move these to some common constants package as we have similar constants in istiorevistion_controller.go file.

controllers/istiorevisiontag/istiorevisiontag_controller.go
```diff
@@ -0,0 +1,541 @@
...
+
+func getReferencedRevisionFromNamespace(labels map[string]string) string {
+	if labels[IstioInjectionLabel] == IstioInjectionEnabledValue {
+		return v1alpha1.DefaultRevision
+	}
+	revision := labels[IstioRevLabel]
+	if revision != "" {
+		return revision
+	}
+	// TODO: if .Values.sidecarInjectorWebhook.enableNamespacesByDefault is true, then all namespaces except system namespaces should use the "default" revision
```
sridhargaddam, line 421: Do we need this TODO here as we are checking for the enableNamespacesByDefault in Line 396?

sridhargaddam:
@dgn, does this PR implement the following from the [SEP](https://github.com/istio-ecosystem/sail-operator/pull/212/files)?

```
When the very first `IstioRevision` is created in a cluster from a `RevisionBased` Istio resource, the Sail Operator will create a `IstioRevisionTag` with the name `default`, referencing that `IstioRevision`. This way, the standard namespace injection label `istio-injection=enabled` will work out of the box for RevisionBased deployments
```

chart/samples/istio-sample-revisionbased.yaml
```diff
@@ -0,0 +1,18 @@
...
+kind: Istio
+metadata:
+  name: default
+spec:
+  version: v1.23.2
+  namespace: istio-system
+  updateStrategy:
+    type: RevisionBased
+---
+apiVersion: sailoperator.io/v1alpha1
```
sridhargaddam, line 11: @dgn JFYI, I am seeing the following reconcile error in the Sail Operator logs when applying this yaml.

```
sail-operator 2024-10-29T12:40:01Z    ERROR    ctrlr.revtag    Reconciler error    {"IstioRevisionTag": "default", "reconcileID": "1ff96be2-388e-41e6-a437-fed29156cc84", "error": "failed to retrieve IstioRevision for IstioRevisionTag \"default\": referenced Istio has no active revision"}
sail-operator sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler
sail-operator     sigs.k8s.io/controller-runtime@v0.19.0/pkg/internal/controller/controller.go:316
sail-operator sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem
sail-operator     sigs.k8s.io/controller-runtime@v0.19.0/pkg/internal/controller/controller.go:263
sail-operator sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func2.2
sail-operator     sigs.k8s.io/controller-runtime@v0.19.0/pkg/internal/controller/controller.go:224
sail-operator 2024-10-29T12:40:01Z    INFO    ctrlr.istio    Reconciling    {"Istio": "default", "reconcileID": "0580456b-b418-4041-a6f8-cd15d7d8d6a3"}
```

docs/api-reference/sailoperator.io.md
dgn: fixed

chart/samples/istio-sample-revisionbased.yaml
dgn: Yeah that is a problem. I'll check if I can make it look less like an error

dgn: should be fixed now!

dgn:
> @dgn, does this PR implement the following from the [SEP](https://github.com/istio-ecosystem/sail-operator/pull/212/files)?
> 
> ```
> When the very first `IstioRevision` is created in a cluster from a `RevisionBased` Istio resource, the Sail Operator will create a `IstioRevisionTag` with the name `default`, referencing that `IstioRevision`. This way, the standard namespace injection label `istio-injection=enabled` will work out of the box for RevisionBased deployments
> ```

As mentioned in Slack, that part will be left out (I moved it to Alternatives Considered) and might be picked up again as part of #439.

controllers/istiorevisiontag/istiorevisiontag_controller.go
dgn: good catch. done!

controllers/istiorevisiontag/istiorevisiontag_controller.go
```diff
@@ -0,0 +1,571 @@
...
+
+	// for other resources, comparing the metadata.generation suffices
+	return oldObject.GetGeneration() != newObject.GetGeneration()
+}
+
+func wrapEventHandler(logger logr.Logger, handler handler.EventHandler) handler.EventHandler {
+	return enqueuelogger.WrapIfNecessary(v1alpha1.IstioRevisionTagKind, logger, handler)
+}
+
+func UnpackReconcilerError(err error) error {
```
nrfox, line 513: This is basically what [Unwrap](https://pkg.go.dev/errors#Unwrap) does?

fjglira:
Do you think we should cover the use of the revision tag also in the e2e test? By adding some test cases related to revision based update

api/v1alpha1/istiorevisiontags_types.go
```diff
@@ -0,0 +1,204 @@
...
+
+const (
+	IstioRevisionTagKind = "IstioRevisionTag"
+	DefaultRevisionTag   = "default"
+)
+
+// IstioRevisionTagSpec defines the desired state of IstioRevisionTag
+type IstioRevisionTagSpec struct {
+	// +kubebuilder:validation:Required
+	TargetRef IstioRevisionTagTargetReference `json:"targetRef"`
```
luksa, line 31: If an `IstioRevisionTag` can reference an `Istio`, doesn't that make the name "revision tag" confusing? Should it be called `IstioTag`?

Also, why exactly do we need the option to reference `Istio`? How do the two reference options fit into the fully-automatic updates? The plan was for the operator to automatically change the `istio.io/rev` label from the old to the new revision, so it should probably also change the `targetRef` in `IstioRevisionTag`. If we decide against the latter, then we should also not implement the former, but then the automatic update wouldn't actually do anything. 

We should discuss this and go over all possible scenarios.

dgn: Well, revision tag is the upstream term. So I'd prefer to keep it around.

The reason it can reference an Istio is exactly to make the upgrade use case easier. The operator doesn't need to change any labels for an automatic upgrade anymore, it will just update the tag to point to the new IstioRevision (it does that already). I honestly think that's the more elegant solution that changing labels on workloads and namespaces - IMO the operator shouldn't touch those user resources.

controllers/istiorevisiontag/istiorevisiontag_controller.go
dgn: Good point. Using `Unwrap()` now

controllers/istiorevisiontag/istiorevisiontag_controller.go
```diff
@@ -0,0 +1,565 @@
...
+	}
+
+	log.Info("Retrieving referenced IstioRevision for IstioRevisionTag")
+	rev, err := r.getIstioRevision(ctx, tag.Spec.TargetRef)
+	if rev == nil || err != nil {
+		return nil, err
+	}
+
+	log.Info("Installing Helm chart")
+	return rev, r.installHelmCharts(ctx, tag, rev)
```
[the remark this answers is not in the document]
dgn: That was indeed my first design, but that would leave us having to expose the `values.revisionTags` field in the IstioRevision spec, which I wanted to avoid because it'll be confusing and conflicting with the IstioRevisionTag CRD. Also, I figured that if we're going to have side-effects (ie 3rd party resources influencing the reconcile loop of a controller) it's probably better to keep them in a low-impact resource/chart like revision tags

controllers/istiorevisiontag/istiorevisiontag_controller.go
```diff
@@ -0,0 +1,565 @@
...
+//
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.14.1/pkg/reconcile
+func (r *Reconciler) Reconcile(ctx context.Context, tag *v1alpha1.IstioRevisionTag) (ctrl.Result, error) {
+	log := logf.FromContext(ctx).WithValues("IstioRevisionTag", tag.Name)
+
+	rev, reconcileErr := r.doReconcile(ctx, tag)
+
+	log.Info("Reconciliation done. Updating labels and status.")
+	labelsErr := r.updateLabels(ctx, tag, rev)
```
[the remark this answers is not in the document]
dgn: I did it to improve the List() operation in the IstioRevision controller, as that can now filter based on a label

pkg/reconciler/errors.go
```diff
@@ -32,3 +32,21 @@ func IsValidationError(err error) bool {
 	e := &ValidationError{}
 	return errors.As(err, &e)
 }
+
+// TransitoryError is an error returned by a Reconciler that will usually resolve itself when retrying, e.g. some resource not yet reconciled
+type TransitoryError struct {
```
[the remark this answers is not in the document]
dgn: Honestly I'm not sure which is the better word but it seemed to be that transitory was more appropriate, compare https://www.difference.wiki/transient-vs-transitory/

No strong preference either way though

controllers/istiorevision/istiorevision_controller.go
```diff
@@ -213,6 +210,14 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 	// The handler triggers the reconciliation of the referenced IstioRevision CR so that its InUse condition is updated.
 	podHandler := wrapEventHandler(logger, handler.EnqueueRequestsFromMapFunc(r.mapPodToReconcileRequest))
 
+	// revisionTagHandler handles Istios that reference an IstioRevision CR as ActiveRevision
+	// The handler triggers the reconciliation of the active IstioRevision CR so that its InUse condition is updated.
+	istioHandler := wrapEventHandler(logger, handler.EnqueueRequestsFromMapFunc(r.mapIstioToReconcileRequest))
```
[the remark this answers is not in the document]
dgn: This fixes a bug I found while testing. It can only occur if no workload or namespace is labelled with the revision tag and/or there are no pod events coming in. The scenario is as follows:

You have a RevisionBased Istio I, a Revision R and a tag T. T references I, and therefore R, R's InUse condition is `true`. Now when you update the version in I, a new revision R* is created, and the active revision of I is updated. This creates a reconcile of T, which is now pointing to R*, which in turn leads to a reconciliation of R*, to mark it InUse=true. Without this watch, R is never reconciled again, which means its InUse condition never flips to `false` - until a Pod event causes its reconciliation.

controllers/istiorevisiontag/istiorevisiontag_controller.go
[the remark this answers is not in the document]
dgn: > Wouldn't that produce a simpler system that's closer to what upstream does?

I'm not sure - what upstream does is fetch the values of the existing control plane, re-render the entire istiod chart and hand-pick the MutatingWebhook YAML and apply it in isolation. So we're pretty close to what upstream does.

pkg/reconciler/errors.go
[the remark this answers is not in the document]
luksa: Interesting. Those descriptions to point more towards "transitory" being the correct term. I've never heard the term "transitory error" before, only "transient error". A google search does confirm that "transient error" is much more commonly used (70.000 results vs 3700 results).

controllers/istiorevision/istiorevision_controller.go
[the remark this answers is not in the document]
luksa: The current approach may also be fragile, since it's possible that it triggers the revision reconciliation before the tag's `status.istioRevision` is updated. I believe there is a chance for a race condition here.

controllers/istiorevisiontag/istiorevisiontag_controller.go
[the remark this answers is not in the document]
luksa: Ah. I'm not sure if you broke the first rule of optimization ("Don't do it") or the second ("Don't do it yet"), but this optimization isn't necessary. The List operation uses the local cache, and there will only be a few IstioRevisionTags in the cluster, so iterating over all of them is perfectly fine.

controllers/istiorevisiontag/istiorevisiontag_controller.go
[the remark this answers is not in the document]
luksa: > That was indeed my first design, but that would leave us having to expose the values.revisionTags field in the IstioRevision spec...

No, I didn't say "add `revisionTags` to `IstioRevision.spec.values`", but for the IstioRevision controller to set the `revisionTags` value only in the map that gets passed to the Helm render function.

> ...what upstream does is fetch the values of the existing control plane, re-render the entire istiod chart and hand-pick the MutatingWebhook YAML and apply it in isolation

OK, then this approach is indeed closer to upstream's, but I feel like it's a hack. The code that cherry-picks the MutatingWebhook YAML must know that the `revisionTags` value is only used in that YAML and in no other YAML produced by the istiod chart. So, it's somewhat fragile. 

I'm not against this approach, but I still feel that the alternative would be much simpler and more robust. It would also automatically cause both the old and the new revision to be reconciled whenever they become or stop being the active revision.

controllers/istiorevision/istiorevision_controller.go
```diff
@@ -398,6 +408,18 @@ func (r *Reconciler) isRevisionReferencedByWorkloads(ctx context.Context, rev *v
 	log := logf.FromContext(ctx)
 	nsList := corev1.NamespaceList{}
 	nsMap := map[string]corev1.Namespace{}
+	// if an IstioRevision is referenced by a revisionTag, it's considered as InUse
```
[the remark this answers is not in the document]
dgn: Done!

pkg/reconciler/errors.go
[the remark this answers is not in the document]
dgn: Going with transient then

luksa (approved):

luksa:
Approved. Also added a hold so you can confirm that all the others' comments were addressed.

PR #417, opened 2024-10-13
Add debug info to multicluster e2e tests

### LINKED ISSUE #381: [Enhancement] Add debug information for e2e multicluster test
When we run the e2e test we need to log more debug information for when the test fails to easily detect what is the root cause

MaxBab:
In case multicluster e2e test fails, we should be debug information regarding the failure.

 

#### What type of PR is this?

- [ ] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [X] Test
- [ ] Documentation Update

Add debug info to multicluster e2e tests in case of test failure.

Fixes #381


luksa (approved):

fjglira:
Hi @MaxBab, I think we need to make changes also in the debug func because we need to pass both clusters or check both clusters configurations, this will need to include certs and others

MaxBab:
@fjglira 
Sure, will look into it.

PR #418, opened 2024-10-14
Make spec.namespace fields immutable

dgn:
Moving a control plane around has consequences beyond just the templates installed by helm, as the control plane namespace has its own semantics within Istio: it is often the place where you can put Istio configs that serve as defaults. By making the field immutable, we avoid strange side- effects that might occur when moving a control plane to another namespace.

### DIFF
--- a/api/v1alpha1/istio_types.go
+++ b/api/v1alpha1/istio_types.go
@@ -57,7 +57,7 @@ type IstioSpec struct {
 	// +kubebuilder:validation:Enum=ambient;default;demo;empty;external;openshift-ambient;openshift;preview;stable
 	Profile string `json:"profile,omitempty"`
 
-	// Namespace to which the Istio components should be installed.
+	// Namespace to which the Istio components should be installed. Note that this field is immutable.
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:io.kubernetes:Namespace"}
 	// +kubebuilder:default=istio-system
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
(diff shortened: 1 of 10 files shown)


dgn:
> So, what's the recommended way to move the control plane to another namespace? I think this deserve a small paragraph in the docs.

I'm not sure. Istio docs don't state anything about moving control planes. Personally, I'd suggest copying the CA root cert over and creating a second control plane, then slowly migrating workloads over. But that's just a guess on my part.

I think this is something the Istio community needs to document. We can then follow their best practices.

> 
> Also, a test for this scenario would be great too.

A test that verifies the namespace change is denied? Sure, I can add one.

FilipB:
> So, what's the recommended way to move the control plane to another namespace? I think this deserve a small paragraph in the docs.
> 
> Also, a test for this scenario would be great too.

IIUC this change would mean it's not possible to move. Only remove/create. Is there any scenario where remove/create is more complicated than the move where the user would have to make sure root cert is moved correctly anyway plus some other unknowns?

jwendell:
> > So, what's the recommended way to move the control plane to another namespace? I think this deserve a small paragraph in the docs.
> 
> I'm not sure. Istio docs don't state anything about moving control planes. Personally, I'd suggest copying the CA root cert over and creating a second control plane, then slowly migrating workloads over. But that's just a guess on my part.

This could be another benefit of using our operator. The operator could handle all of this, no?

dgn:
I added a test!

jwendell (approved):
LGTM.

Should we add a note on the docs saying the namespace cannot be changed?

/hold

dgn:
> LGTM.
> 
> Should we add a note on the docs saying the namespace cannot be changed?
> 
> /hold

Done!

PR #427, opened 2024-10-16
Include integration tests in code coverage calculation

dgn:
This makes sure we're also measuring code coverage for our integration tests. I hope codecov is smart enough to merge the files.

### DIFF
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -10,12 +10,12 @@ jobs:
 
     - name: Run coverage
       run: |
-        make test.unit \
+        make test \
           -e COVERAGE=true
 
     - name: Upload to Codecov
       uses: codecov/codecov-action@v4
       with:
-        file: out/coverage.out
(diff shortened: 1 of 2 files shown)


dgn:
Hmm this takes way too long and is going to eat up our GH actions budget. Will check if we can just upload after running the tests in prow.

/hold

luksa:
So this re-runs all the tests even though we already run them in the other jobs? Isn't there an option to generate a coverage file there and then use that file to generate this report?

dgn:
> So this re-runs all the tests even though we already run them in the other jobs? Isn't there an option to generate a coverage file there and then use that file to generate this report?

I'm looking at removing the prow jobs if we have enough GitHub action budget, and then only running the github actions for unit and integration tests. The problem with generating the profiles in the prow jobs is the following:
- we cannot upload the profiles from the prowjobs because we cannot add secrets to the prow instance
- we could push the profiles to the repo and then just use github actions to upload but it's ~0.5mb to add to the repo

luksa (approved):

PR #432, opened 2024-10-17
Honor the `--log-enqueue-events` option in all controllers

luksa:
Previously, this was only honored in the IstioRevision controller.


dgn (approved):
lgtm.

codecov is getting a bit aggressive here

PR #446, opened 2024-10-23
Ignore istio version suffix used in the downstream builds

FilipB:
#### What type of PR is this?

- [ ] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [X] Test
- [ ] Documentation Update

The test is expecting to see the same version parsed from versions.yaml and a version string returned by `pilot-discovery version`. This is not the same in downstream builds as we use the `_ossm` suffix.


pbajjuri20 (approved):
LGTM..!

eg: In Version:"1.23.0-snapshot", it will capture 1.23.0 and stop before the underscore.

dgn (approved):

PR #450, opened 2024-10-25
Pin the version of OLM we use for e2e tests

dgn:
The latest release does not work, but 0.28.0 does (see https://github.com/operator-framework/operator-lifecycle-manager/issues/3419). I also added some automation to update-deps to update the version.


bmangoen (dismissed):

tests/e2e/common-operator-integ-suite.sh
```diff
@@ -241,9 +241,12 @@ if [ "${SKIP_BUILD}" == "false" ]; then
...
     OPENSHIFT_PLATFORM=false \
     make bundle bundle-build bundle-push
-     
+
     # Install OLM in the cluster because it's not available by default in kind.
-    ${OPERATOR_SDK} olm install
+    if [ "${OLM_VERSION}" != "" ]; then
+      OLM_INSTALL_ARGS="--version ${OLM_VERSION}"
+    fi
+    ${OPERATOR_SDK} olm install "${OLM_INSTALL_ARGS}"
```
FilipB, line 249: Isn't this failing when OLM_INSTALL_ARGS is undefined?

tools/update_deps.sh
```diff
@@ -58,6 +58,10 @@ sed -i "s|CONTROLLER_TOOLS_VERSION ?= .*|CONTROLLER_TOOLS_VERSION ?= ${CONTROLLE
 OPM_LATEST_VERSION=$(getLatestVersion operator-framework/operator-registry)
 sed -i "s|OPM_VERSION ?= .*|OPM_VERSION ?= ${OPM_LATEST_VERSION}|" "${ROOTDIR}/Makefile.core.mk"
 
+# Update olm
+OLM_LATEST_VERSION=$(getLatestVersion operator-framework/operator-registry)
```
bmangoen, line 62: Typo here `operator-framework/operator-lifecycle-manager`

tests/e2e/common-operator-integ-suite.sh
FilipB: It will fail if undefined as the script is using `set -eu -o pipefail`

dgn: good point. fixed

tools/update_deps.sh
dgn: Thanks, fixed!

FilipB (approved):
Thank you

jwendell (approved):

PR #451, opened 2024-10-25
Set global.platform when not set by the user

luksa:
This change ensures that the `openshift` platform is set correctly even when the `openshift` profile isn't selected via `spec.profile`.

Users can now select a profile (e.g. `remote`) and still have the correct platform selected automatically.

### DIFF
--- a/pkg/istiovalues/profiles.go
+++ b/pkg/istiovalues/profiles.go
@@ -19,6 +19,7 @@ import (
 	"os"
 	"path"
 
+	"github.com/istio-ecosystem/sail-operator/pkg/config"
 	"github.com/istio-ecosystem/sail-operator/pkg/helm"
 	"github.com/istio-ecosystem/sail-operator/pkg/reconciler"
 	"gopkg.in/yaml.v3"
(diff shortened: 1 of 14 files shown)


pkg/istiovalues/profiles.go
```diff
@@ -27,13 +28,22 @@ import (
...
+) (helm.Values, error) {
 	profile := resolve(defaultProfile, userProfile)
 	defaultValues, err := getValuesFromProfiles(path.Join(resourceDir, version, "profiles"), profile)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get values from profile %q: %w", profile, err)
 	}
-	return mergeOverwrite(defaultValues, userValues), nil
+	values := helm.Values(mergeOverwrite(defaultValues, userValues))
+
+	if platform != config.PlatformKubernetes && platform != config.PlatformUndefined {
```
sridhargaddam, line 41: Apart from "openshift" (and kubernetes), are there any valid platforms used in Istio?

luksa: Yes, there are also:
- k3d
- k3s
- microk8s
- minikube

See `profile-platform-*.yaml` files [here](https://github.com/istio/istio/tree/master/manifests/charts/istio-control/istio-discovery/files).

sridhargaddam (approved):

PR #453, opened 2024-10-28
Consolidate reconciler config fields into ReconcilerConfig

luksa:
Most reconcilers need the same three config fields: `ResourceDirectory`, `Platform`, and `DefaultProfile`. Instead of passing these fields separately, we now pass them via a `ReconcilerConfig` struct.


dgn (approved):
LGTM. Thanks Marko

PR #461, opened 2024-10-31
Remove RemoteIstio resource

luksa:
In latest versions of Istio, the istiod-remote chart has been removed. Instead, users deploy the regular istiod chart, but set the profile to `remote`.


FilipB:
Just referencing related PR which might not be needed when this is merged? Not sure. https://github.com/istio-ecosystem/sail-operator/pull/464

dgn (approved):

PR #483, opened 2024-11-20
Centralize the way the sample apps yaml are being set to be deployed

### LINKED ISSUE #467: [FEATURE] Avoid the use of bookinfo as testing app or enable a way to set custom images

### Feature Description

We can test the configuration in the e2e framework by using an alternative containers or adding the possibility to set custom images that can be set by the user to avoid pulling always upstream images to run the test

### Additional Information

_No response_

fjglira:
#### What type of PR is this?

- [X] Enhancement / New Feature
- [ ] Bug Fix
- [X] Refactor
- [ ] Optimization
- [ ] Test
- [ ] Documentation Update

This PR addresses these needs:
* Centralize the way that the pod yaml URL is being used across all the e2e tests to simplify the deployment of the sample application
* Stop using bookinfo app to validate sidecar injection
* Adding POD_SAMPLE_YAML_BASE_URL to enable a way to set custom yaml string that will allow user to use their own yaml with custom images

Fixes #467

### DIFF
--- a/tests/e2e/util/common/e2e_utils.go
+++ b/tests/e2e/util/common/e2e_utils.go
@@ -278,15 +278,13 @@ func GetYAMLPodURL(version supportedversion.VersionInfo, appName string) string 
 	switch appName {
 	case "tcp-echo-dual-stack":
 		url = "samples/tcp-echo/tcp-echo-dual-stack.yaml"
-	case "tcp-echo-ipv4":
-	case "tcp-echo":
+	case "tcp-echo-ipv4", "tcp-echo":
 		url = "samples/tcp-echo/tcp-echo-ipv4.yaml"
 	case "tcp-echo-ipv6":
 		url = "samples/tcp-echo/tcp-echo-ipv6.yaml"
 	case "sleep":
 		url = "samples/sleep/sleep.yaml"
-	case "helloworld":
-	case "sample":
+	case "helloworld", "sample":
 		url = "samples/helloworld/helloworld.yaml"
 	default:
 		return ""
--- a/tests/e2e/util/common/e2e_utils.go
+++ b/tests/e2e/util/common/e2e_utils.go
@@ -19,6 +19,7 @@ package common
 import (
 	"context"
 	"fmt"
+	"os"
 	"path/filepath"
 	"regexp"
 	"strings"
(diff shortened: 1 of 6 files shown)


fjglira:
I'll check tomorrow morning the failure in the suites: dual-stack and multicluster. For sure, I left some errors in the new way of setting the sample apps

tests/e2e/util/common/e2e_utils.go
```diff
@@ -261,3 +263,46 @@ func InstallOperatorViaHelm(extraArgs ...string) error {
...
+		url = "samples/sleep/sleep.yaml"
+	case "helloworld":
+	case "sample":
+		url = "samples/helloworld/helloworld.yaml"
+	default:
+		return ""
+	}
+
+	// Get the base URL from the environment variable or use the default
+	baseURL := os.Getenv("POD_SAMPLE_YAML_BASE_URL")
```
MaxBab, line 296: Does it mean, that if I'm setting a custom "baseURL" based on the env var, the path for the yaml files should be identical to the mentioned in lines 279-290?

fjglira: Yes, I'll improve this to be able to set the full path

fjglira: @MaxBab take a look, I think the best way is to set each yaml individually. With this, wdyt?

MaxBab: @fjglira 

Cool!
I like this approach.

MaxBab (approved):

tests/e2e/util/common/e2e_utils.go
```diff
@@ -261,3 +263,60 @@ func InstallOperatorViaHelm(extraArgs ...string) error {
 func UninstallOperator() error {
 	return helm.Uninstall("sail-operator", "--namespace", OperatorNamespace)
 }
+
+// GetYAMLPodURL returns the URL of the yaml file for the testing app.
+// args:
+// version: the version of the Istio to get the yaml file from.
+// appName: the name of the testing app. Example: helloworld, sleep, tcp-echo.
+func GetYAMLPodURL(version supportedversion.VersionInfo, appName string) string {
```
dgn, line 271: I know it was called that before, but `GetYAMLPodURL` is a strange name for this func. Isn't  `GetSampleYAML` more adequate? Not sure what the pod has to do with this?

tests/e2e/util/common/e2e_utils.go
```diff
@@ -261,3 +263,60 @@ func InstallOperatorViaHelm(extraArgs ...string) error {
...
+	case "sleep":
+		url = "samples/sleep/sleep.yaml"
+	case "helloworld", "sample":
+		url = "samples/helloworld/helloworld.yaml"
+	default:
+		return ""
+	}
+
+	// Base URL logic
+	baseURL := os.Getenv("POD_SAMPLE_YAML_BASE_URL")
```
dgn, line 312: why the `POD` prefix?

dgn (commented):
LGTM to me, good improvement! Only two nits

dgn (approved):

PR #496, opened 2024-11-27
Remove validation when generating YAML manifests

sridhargaddam:


dgn (approved):
Good catch, thank you!

PR #506, opened 2024-12-03
Add documentation for running multiple meshes on a single cluster

luksa:


docs/README.md
```diff
@@ -468,6 +476,250 @@ Steps:
...
+   metadata:
+     name: default
+     namespace: istio-system1
+   spec:
+     mtls:
+       mode: STRICT
+   EOF
+   
+   $ kubectl apply -f - <<EOF
+   apiVersion: security.istio.io/v1beta1
```
MaxBab, line 562: I guess the apiVersion here should be `security.istio.io/v1` and not `security.istio.io/v1beta`.
The example in line 551 does not match line 562.

luksa: Thanks. Fixed.

MaxBab (approved):

PR #507, opened 2024-12-04
Add config option to set maximum Istio version

dgn:


dgn:
It looks like the pre-release versions are not accepted by the semver regex. Which is strange, I think they should be fine

controllers/istio/istio_controller.go
```diff
@@ -98,6 +99,13 @@ func validate(istio *v1alpha1.Istio) error {
 	if istio.Spec.Version == "" {
 		return reconciler.NewValidationError("spec.version not set")
 	}
+	istioVersion, err := semver.NewVersion(istio.Spec.Version)
+	if err != nil {
+		return reconciler.NewValidationError("spec.version is not a valid semver: " + err.Error())
+	}
+	if config.Config.MaximumIstioVersion != nil && istioVersion.GreaterThan(config.Config.MaximumIstioVersion) {
+		return reconciler.NewValidationError("spec.version is not supported")
```
luksa, line 107: Can you add the version to the error message. This makes it read as though setting the field itself is not supported.

luksa (commented):
Hmm, this approach won't work, because `spec.version` is currently not really an actual version number. It's just a string (for example, when `spec.version: latest`, `latest` is just the name, whereas the actual version number is whatever is set in the `version` field in versions.yaml). But the operator currently doesn't even read the versions.yaml file. 

The operator currently doesn't really know anything about versions. It simply takes the `spec.version` string and uses it to construct the path for the chart it must render. 

We either need to remove the version names from versions.yaml and just use the version number as the version name (but this prevents us from using "latest" as the version name), or we make the operator read the versions.yaml file, allowing it to map the version name to the version number.

luksa:
Actually, we could get rid of the version names by simply defaulting an empty `spec.version` to whatever the latest version is. Although, without versions.yaml, the operator would have to read the directory names and find the highest semver.

dgn:
> Hmm, this approach won't work, because `spec.version` is currently not really an actual version number. It's just a string (for example, when `spec.version: latest`, `latest` is just the name, whereas the actual version number is whatever is set in the `version` field in versions.yaml). But the operator currently doesn't even read the versions.yaml file.
> 
> The operator currently doesn't really know anything about versions. It simply takes the `spec.version` string and uses it to construct the path for the chart it must render.
> 
> We either need to remove the version names from versions.yaml and just use the version number as the version name (but this prevents us from using "latest" as the version name), or we make the operator read the versions.yaml file, allowing it to map the version name to the version number.

I think this is good and we should do it as part of implementing version aliases. `latest` then just becomes an alias for the latest version in versions.yaml.

For this specific instance, we just need to be able have an upper limit on the version, so IMO it would be a bit overkill to implement this (and then revise it anyway when implementing version aliases).

controllers/istio/istio_controller.go
dgn: done!

dgn:
Closing this as per our discussion. We'll use a normal branching strategy and re-evaluate after a few minor releases

PR #534, opened 2024-12-17
updated multi primary multi network docs

jforce:
I was just trying out multi primary multi network by following the documentation here and it wasn't working until I added `spec.values.pilot.env.ROOT_CA_DIR: /etc/cacerts` in the `Istio` resource. 

Without this, I was getting the below error:
```
upstream connect error or disconnect/reset before headers. retried and the latest reset reason: remote connection failure, transport failure reason: TLS_error:|268435581:SSL routines:OPENSSL_internal:CERTIFICATE_VERIFY_FAILED:TLS_error_end
```
It seems like some other versions of the docs specified this too [github.com/openshift-service-mesh/sail-operator/pull/144](https://github.com/openshift-service-mesh/sail-operator/pull/144/files)
@

### DIFF
--- a/docs/multicluster/setup-multi-primary.sh
+++ b/docs/multicluster/setup-multi-primary.sh
@@ -76,7 +76,7 @@ kubectl get ns sail-operator --context "${CTX_CLUSTER1}" || make -C "${SCRIPT_DI
 kubectl config use-context "${CTX_CLUSTER2}"
 kubectl get ns sail-operator --context "${CTX_CLUSTER2}" || make -C "${SCRIPT_DIR}/../.." deploy
 
-# 3. Create istio-system namespace on each cluster and configure a common root CA. 
+# 3. Create istio-system namespace on each cluster and configure a common root CA.
 
 kubectl get ns istio-system --context "${CTX_CLUSTER1}" || kubectl create namespace istio-system --context "${CTX_CLUSTER1}"
 kubectl --context "${CTX_CLUSTER1}" label namespace istio-system topology.istio.io/network=network1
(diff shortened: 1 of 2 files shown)


docs/multicluster/setup-multi-primary.sh
```diff
@@ -110,6 +110,8 @@ spec:
         requests:
           cpu: 100m
           memory: 1024Mi
+      env:
+        ROOT_CA_DIR: /etc/cacerts
```
luksa, line 114: Hmm. The default path for `ROOT_CA_DIR` is `./etc/cacerts` (defined [here](https://github.com/istio/istio/blob/492485ffbbaf21aebb468ba668aab340c13366eb/pilot/pkg/bootstrap/istio_ca.go#L81)), but we mount the `cacerts` secret into `/etc/cacerts` ([here](https://github.com/istio/istio/blob/master/manifests/charts/istio-control/istio-discovery/templates/deployment.yaml#L232)). 

I wonder why the default doesn't match. Maybe because the working dir used to be `/`, which made `./etc/cacerts == /etc/cacerts`, but that's no longer the case?

luksa (approved):
LGTM.

We can revert this if/when we fix the default path.

luksa:
Hmm, interesting. The RedHat's build of the pilot image does set `WORKDIR /`. The upstream Istio image also uses `/` as the workdir:
```
$ docker run --entrypoint pwd docker.io/istio/pilot:1.24.1
/
```

So I'm not sure why it doesn't work in your case. What pilot image are you using, @jforce?

jforce:
I don't have the cluster I was using to hand now, but I was using OSSM 3.0 TP2 (istio 1.24.1) on OpenShift (so the Red Hat build)

dgn:
This was most likely caused by using the 1.23 images which were shipped with TP1 and had the bug. When using 1.24 this should not be required

PR #536, opened 2024-12-18
Fix usages of `istioctl create-remote-secret`

luksa:
By default, istioctl creates the ServiceAccount if it doesn't yet exist. If you run the `create-remote-secret` command before you create the Istio resource, the ServiceAccount won't yet exist and will therefore be created, along with a ClusterRole and ClusterRoleBinding. This will later cause the reconciliation of the Istio resource to fail because Helm will refuse to overwrite these resources.


luksa:
Looks like this is no longer necessary.

PR #537, opened 2024-12-18
Ensure profile files are included in the revisiontags chart

luksa:


dgn (commented):
Thanks, this is exactly the fix I had in mind. Only thing missing for me would be a test that verifies it (e.g. by altering an integration test to use a profile).

luksa:
I think an e2e would be better. It doesn't look like we have an e2e test for revision tags yet.

dgn (approved):

Issue #500, opened 2024-11-29
[EPIC] Ambient mode support in Sail Operator


sridhargaddam:
### Epic Description

Add necessary support in the Sail Operator to deploy Istio in Ambient mode.
- [x] [Implement API and CRD for ztunnel component](https://github.com/istio-ecosystem/sail-operator/pull/481)
- [x] [Update ztunnel versions in update_version-list.sh](https://github.com/istio-ecosystem/sail-operator/pull/509)
- [x] [Implement controller for managing lifecycle of ztunnel](https://github.com/istio-ecosystem/sail-operator/pull/512)
- [x] [Fix `make undeploy` in an Ambient deployment](https://github.com/istio-ecosystem/sail-operator/pull/533) 
- [x] [Allow overriding Ambient params in IstioCNI](https://github.com/istio-ecosystem/sail-operator/pull/563) 
- [x] [Implement an e2e test for Ambient config in IstioCNI](https://github.com/istio-ecosystem/sail-operator/pull/591) 
- [x] [Fix ztunnel controller to apply values.ztunnel config](https://github.com/istio-ecosystem/sail-operator/pull/568)
- [x] [Use logAsJson instead of logAsJSON in ZTunnel APIs](https://github.com/istio-ecosystem/sail-operator/pull/584) 
- [x] [Add necessary e2e tests for Sail Operator ztunnel controller.](https://github.com/istio-ecosystem/sail-operator/pull/541)
- [x] [Enhance make commands for ambient support and platform auto-discovery](https://github.com/istio-ecosystem/sail-operator/issues/545) 
- [x] [Create Ambient developer documentation.](https://github.com/istio-ecosystem/sail-operator/pull/579)
- [x] [Update Ambient docs with waypoint user-guide.](https://github.com/istio-ecosystem/sail-operator/pull/694)

PR #541, opened 2024-12-23
e2e tests for validating Ambient deployment

sridhargaddam:
This PR implements e2e tests for control plane and dataplane validation of Ambient profile in the Sail Operator.

Related to: https://github.com/istio-ecosystem/sail-operator/issues/500

### DIFF
--- a/tests/e2e/ambient/ambient_suite_test.go
+++ b/tests/e2e/ambient/ambient_suite_test.go
@@ -31,7 +31,6 @@ import (
 var (
 	cl                    client.Client
 	err                   error
-	ocp                   = env.GetBool("OCP", false)
 	operatorNamespace     = common.OperatorNamespace
 	deploymentName        = env.Get("DEPLOYMENT_NAME", "sail-operator")
 	controlPlaneNamespace = env.Get("CONTROL_PLANE_NS", "istio-system")
--- a/tests/e2e/ambient/ambient_suite_test.go
+++ b/tests/e2e/ambient/ambient_suite_test.go
@@ -1,0 +1,67 @@
+//go:build e2e
+
+// Copyright Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ambient
+
+import (
+	"testing"
+
+	k8sclient "github.com/istio-ecosystem/sail-operator/tests/e2e/util/client"
+	"github.com/istio-ecosystem/sail-operator/tests/e2e/util/common"
+	"github.com/istio-ecosystem/sail-operator/tests/e2e/util/env"
+	"github.com/istio-ecosystem/sail-operator/tests/e2e/util/kubectl"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	cl                    client.Client
+	err                   error
+	ocp                   = env.GetBool("OCP", false)
+	operatorNamespace     = common.OperatorNamespace
+	deploymentName        = env.Get("DEPLOYMENT_NAME", "sail-operator")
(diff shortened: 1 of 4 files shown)


sridhargaddam:
Test suite `e2e-kind-olm_sail-operator_main` will be fixed if the following PR (or any related fix) is merged - https://github.com/istio-ecosystem/sail-operator/pull/543
CC: @luksa

tests/e2e/ambient/ambient_suite_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/istio-ecosystem/sail-operator/tests/e2e/util/kubectl"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	cl                    client.Client
+	err                   error
+	ocp                   = env.GetBool("OCP", false)
```
yxun, line 34: Could we add a comment about why we need this `ocp` environment boolean ? Is there any specific configuration differences or test result differences when running ambient tests on OpenShift ?

tests/e2e/ambient/ambient_test.go
```diff
@@ -0,0 +1,365 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	sleepNamespace   = "sleep"
+	httpbinNamespace = "httpbin"
+)
+
+var _ = Describe("Ambient configuration ", Ordered, func() {
+	SetDefaultEventuallyTimeout(180 * time.Second)
```
yxun, line 44: Could we replace the hard coded `180` value with a const ?  We may update all those hard coded values together in the const block if needed.

tests/e2e/ambient/ambient_test.go
```diff
@@ -0,0 +1,365 @@
...
+	SetDefaultEventuallyPollingInterval(time.Second)
+
+	debugInfoLogged := false
```
yxun, line 47: Should we set the default value of debugInfoLogged to `true` ? When a test fails, it's useful to troubleshoot it with debug logs.

tests/e2e/ambient/ambient_test.go
```diff
@@ -0,0 +1,365 @@
...
+			Context(fmt.Sprintf("Istio version %s", version.Version), func() {
+				BeforeAll(func() {
+					Expect(k.CreateNamespace(controlPlaneNamespace)).To(Succeed(), "Istio namespace failed to be created")
+					Expect(k.CreateNamespace(istioCniNamespace)).To(Succeed(), "IstioCNI namespace failed to be created")
+					Expect(k.CreateNamespace(ztunnelNamespace)).To(Succeed(), "ZTunnel namespace failed to be created")
+				})
+
+				When("the IstioCNI CR is created with ambient profile", func() {
+					BeforeAll(func() {
+						cniYAML := `
```
yxun, line 85: I prefer putting all yaml const strings together in the const block. Just in case, when we need to update some yaml examples or manifests, we can find them at the beginning of the file.

tests/e2e/ambient/ambient_test.go
```diff
@@ -0,0 +1,365 @@
...
+
+						Expect(ztunnelObj).To(HaveContainersThat(ContainElement(WithTransform(getEnvVars,
+							ContainElement(corev1.EnvVar{Name: "ISTIO_META_ENABLE_HBONE", Value: "true"})))),
+							"Expected ISTIO_META_ENABLE_HBONE to be set to true, but not found")
+					})
+				})
+
+				// We spawn the following pods to verify the data-path connectivity.
+				// an httpbin service in httpbin namespace that listens of port 8000
+				// using a sleep pod from the sleep namespace, we try to connect to the httpbin service to verify that connectivity is successful.
```
yxun, line 222: What does this test validate ? A sleep service can still access a httpbin service even without ambient mode or label , right ? 
I think we should at least validate a ztunnel log about the sleep, httpbin connectivity log. 
For example, `kubectl -n ztunnel logs -l app=ztunnel | grep -E "inbound|outbound"`

sridhargaddam: Rather than looking at the logs, lets verify that the ztunnel has opened the listening sockets in the pod network namespace (more specifically there is a socket listening on HBONE port). Let me update the PR.

tests/e2e/ambient/ambient_suite_test.go
sridhargaddam: Updated the PR and removed the need for `ocp` boolean. 
Also pushed a separate PR for removing it from other tests - https://github.com/istio-ecosystem/sail-operator/pull/555

yxun (approved):

dgn (approved):

PR #543, opened 2024-12-24
Revert "Remove redundant variable copying in for loops"

sridhargaddam:
Reverts istio-ecosystem/sail-operator#473

### DIFF
--- a/common/config/.golangci.yml
+++ b/common/config/.golangci.yml
@@ -18,7 +18,7 @@ linters:
   disable-all: true
   enable:
     - errcheck
-    - copyloopvar
+    - exportloopref
     - depguard
     - gocritic
     - gofumpt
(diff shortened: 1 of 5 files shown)


sridhargaddam:
@luksa this PR is [fixing the issue](https://storage.googleapis.com/istio-prow/pr-logs/pull/istio-ecosystem_sail-operator/541/e2e-kind-olm_sail-operator_main/1871472343352610816/build-log.txt) with `e2e-kind-olm_sail-operator_main` job. Can you please take a look if we should go ahead with the revert (or bump the go version at some appropriate place)?

luksa:
@sridhargaddam I don't think we need to revert this, but we do need to run the tests with go 1.22+. Is that not the case?

sridhargaddam:
> @sridhargaddam I don't think we need to revert this, but we do need to run the tests with go 1.22+. Is that not the case?

I was noticing some failures in ambient e2e tests due to race conditions and suspected this PR. However, as you mentioned, this is not the reason for the failures and the go version we are using in our CI tests is 1.23.4. Thanks, I'll close this PR.

PR #553, opened 2025-01-09
Fix InUse detection in IstioRevisionTag controller

### LINKED ISSUE #551: [BUG] Istio revision tag status not working when the injection is enabled on a pod via sidecar.istio.io/inject: 'true'

### Bug Description

Having following Istio:
```
apiVersion: sailoperator.io/v1alpha1
kind: Istio
metadata:
  name: ossm-3
spec:
  namespace: istio-system
  updateStrategy:
    type: RevisionBased
  version: v1.24.1
```

and following IstioRevisionTag:
```
apiVersion: sailoperator.io/v1alpha1
kind: IstioRevisionTag
metadata:
  name: default
spec:
  targetRef:
    kind: IstioRevision
    name: ossm-3-v1-24-1
```
the `IstioRevisionTag` status shows `NotReferencedByAnything` even if there are pods with a `sidecar.istio.io/inject: 'true'` label and injected proxies.

```
oc get istiorevisiontags
NAME      STATUS                    IN USE   REVISION         AGE
default   NotReferencedByAnything   False    ossm-3-v1-24-1   24m
```

On the other hand, following is shown when you label an empty namespace (so there no proxies running) with `istio-injection=enabled`:
```
oc get istiorevisiontags
NAME      STATUS    IN USE   REVISION         AGE
default   Healthy   True     ossm-3-v1-24-1   24m
```

### Operator Version

0.3.0-nightly-2025-01-08

### Link to Gist with Logs

_No response_

### Additional Information

_No response_

dgn:
Fixes #551. I moved some shared code into the `revision` package and adapted the IstioRevision controller to use the new functions. Also adds unit tests for IstioRevisionTag controller.

### DIFF
--- a/controllers/istiorevision/istiorevision_controller.go
+++ b/controllers/istiorevision/istiorevision_controller.go
@@ -510,21 +510,34 @@ func (r *Reconciler) mapNamespaceToReconcileRequest(ctx context.Context, ns clie
 	return requests
 }
 
+// mapPodToReconcileRequest will collect all referenced revisions from a pod and its namespace and trigger reconciliation
 func (r *Reconciler) mapPodToReconcileRequest(ctx context.Context, pod client.Object) []reconcile.Request {
+	revisionNames := []string{}
 	revisionName := revision.GetInjectedRevisionFromPod(pod.GetAnnotations())
-	if revisionName == "" {
-		ns := corev1.Namespace{}
-		err := r.Client.Get(ctx, types.NamespacedName{Name: pod.GetNamespace()}, &ns)
-		if err != nil {
-			return nil
-		}
-		revisionName = revision.GetReferencedRevisionFromNamespace(ns.GetLabels())
-		if revisionName == "" {
-			revisionName = revision.GetReferencedRevisionFromPod(pod.GetLabels())
+	if revisionName != "" {
+		revisionNames = append(revisionNames, revisionName)
+	} else {
+		revisionName = revision.GetReferencedRevisionFromPod(pod.GetLabels())
+		if revisionName != "" {
+			revisionNames = append(revisionNames, revisionName)
 		}
 	}
+	ns := corev1.Namespace{}
+	err := r.Client.Get(ctx, types.NamespacedName{Name: pod.GetNamespace()}, &ns)
+	if err != nil {
+		return nil
+	}
+	revisionName = revision.GetReferencedRevisionFromNamespace(ns.GetLabels())
 	if revisionName != "" {
-		return []reconcile.Request{{NamespacedName: types.NamespacedName{Name: revisionName}}}
+		revisionNames = append(revisionNames, revisionName)
+	}
+
+	if len(revisionNames) > 0 {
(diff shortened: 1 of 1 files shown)


pkg/revision/references.go
```diff
@@ -0,0 +1,58 @@
...
+
+	return ""
+}
+
+func GetReferencedRevisionFromPod(podLabels map[string]string) string {
+	// we only look at pod labels to identify injection intent
+	if podLabels[constants.IstioSidecarInjectLabel] != "false" {
+		if revisionFromPod := podLabels[constants.IstioRevLabel]; revisionFromPod != "" {
+			return revisionFromPod
+		}
```
luksa, line 41: Nit. Since the scope of the variable is smaller, we can shorten its name. But yeah, it could have been shorter even in the previous code.
```suggestion
		if rev := podLabels[constants.IstioRevLabel]; rev != "" {
			return rev
		}
```

controllers/istiorevision/istiorevision_controller.go
```diff
@@ -532,24 +503,28 @@ func (r *Reconciler) mapNamespaceToReconcileRequest(ctx context.Context, ns clie
...
+	if revisionName == "" {
+		ns := corev1.Namespace{}
+		err := r.Client.Get(ctx, types.NamespacedName{Name: pod.GetNamespace()}, &ns)
+		if err != nil {
+			return nil
+		}
+		revisionName = revision.GetReferencedRevisionFromNamespace(ns.GetLabels())
+		if revisionName == "" {
+			revisionName = revision.GetReferencedRevisionFromPod(pod.GetLabels())
+		}
```
luksa, line 524: Does the namespace label take precedence over the pod label? 

Also, if they refer to different revisions, do we need to reconcile both revisions?

luksa: Apparently it does. At least according to the tests.

dgn: > Does the namespace label take precedence over the pod label?
Counter-intuitively, yes. Luckily, the existing test cases in the IstioRevision controller covered this already.

> Also, if they refer to different revisions, do we need to reconcile both revisions?
Hmm. I guess technically yes, to make sure that InUse can be flipped to `false` in the case that you previusly had a pod label and are now adding a namespace label. Will make the change.

dgn: Rewrote the func in the latest commit. PTAL

luksa: Yeah, but if you're adding a namespace label, then `mapNamespaceToReconcile` is called, so it's probably not necessary to handle this in `mapPodToReconcileRequest`. And I guess we don't really need to handle cases where the pod label or annotation changes, since that means that someone is manually messing with them.

dgn: oh, you're right. We probably don't need the change then. I can revert the commit

luksa (approved):
LGTM. Much cleaner now.

PR #557, opened 2025-01-15
Update build tools

dgn:


FilipB:
I think this should be done automatically via https://github.com/istio-ecosystem/sail-operator/pull/540

dgn:
> I think this should be done automatically via #540

You're right. Closing

PR #560, opened 2025-01-17
Switch to sail-dev quay org

### LINKED ISSUE #347: Move to sail-dev quay org
We're still using the maistra-dev quay org.

While at it, we should make sure that only our bot accounts have permissions to push there, so we don't accidentally push to the public repos during development.

dgn:
Moving off of the old `maistra-dev` org on quay.io. I also created a `sail-operator` org on quay which we might want to use for GA releases later on.

Fixes #347.


dgn:
I already updated the secrets for GitHub actions so that nightly and release images can be pushed to the new org.

sridhargaddam (approved):

PR #576, opened 2025-01-22
[release-1.0] Prepare 1.0 release

dgn:
Change build image and common-files to 1.24, update versions.yaml


bundle/manifests/extensions.istio.io_wasmplugins.yaml
```diff
@@ -133,14 +133,14 @@ spec:
                       type: string
                       x-kubernetes-validations:
                       - message: wildcard not allowed in label value match
-                        rule: '!self.contains("*")'
+                        rule: '!self.contains(''*'')'
```
FilipB, line 136: Just curious, why escaping single quote instead of using double quotes?

go.mod
```diff
@@ -25,8 +25,8 @@ require (
 	gomodules.xyz/jsonpatch/v2 v2.4.0
 	gopkg.in/yaml.v3 v3.0.1
 	helm.sh/helm/v3 v3.16.3
-	istio.io/client-go v1.24.0-alpha.0.0.20250103213757-fb95213c2bc2
-	istio.io/istio v0.0.0-20250119170654-0c5460eeba8a
+	istio.io/client-go v1.24.3-0.20250110022005-e8de2e86eed8
```
FilipB, line 28: The latest version in `versions.yaml` is `v1.24.2` should we use that one instead?

bundle/manifests/extensions.istio.io_wasmplugins.yaml
dgn: this is from the upstream CRD. no idea honestly, this is usually all generated

howardjohn: It goes through a tool so its all mechanical

FilipB: Thank you

go.mod
dgn: good point, will do

dgn: interestingly, that also removed some of the strange generated CRD changes. Good catch!

FilipB (approved):

jwendell (approved):

Issue #581, opened 2025-01-23
Disable imageDigest support using annotations in ztunnel controller


sridhargaddam:
### Is this the right place to submit this?

- [x] This is not a question about how to use the sail-operator

### Bug Description

Currently, we have some pending work with the ztunnel images that are used as imageDigests. Because of this, when deploying Ambient profile using the OSSM operator, the ztunnel pod fails to come up.

**Alternatives:**
1. Default Image: If no image details are provided in the Ztunnel CR, the operator will use the default image from the helm chart.
2. Explicitly specify image details: Following a recent [fix in the ztunnel controller](https://github.com/istio-ecosystem/sail-operator/pull/568), users can specify image details within the ztunnel CR itself in one of the following ways.
     2.1 `ztunnel.spec.values.ztunnel.image`: Define the image details directly for the ztunnel deployment.
     2.2 `ztunnel.spec.values.global.hub/tag`: Set the global hub and tag to be used for ztunnel deployments.

PR #568, opened 2025-01-21
Fix ztunnel controller to apply values.ztunnel config

sridhargaddam:
In the Sail Operator, any config specified under values.ztunnel in the ZTunnel resource was being ignored. This fix ensures that the ztunnel controller correctly applies these values.

Fixes: https://github.com/istio-ecosystem/sail-operator/issues/567

PR #582, opened 2025-01-23
Disable imageDigest support in ztunnel controller

sridhargaddam:
Currently, we have some pending work with the ztunnel images that are used as imageDigests. Because of this, when deploying Ambient profile using the OSSM operator, the ztunnel pod fails to come up. This PR temporarily disables this functionality until proper support is implemented.

Alternatives:
1. Default Image: If image details are not provided in the ZTunnel CR, the operator will use the default image from the helm chart.
2. Explicitly specify image details: Following a recent fix in the ztunnel controller ([PR#568](https://github.com/istio-ecosystem/sail-operator/pull/568)), users can specify image details within the ztunnel CR itself in one of the following ways. 
    2.1: ztunnel.spec.values.ztunnel.image
    2.2: ztunnel.spec.values.global.hub/tag

Note: This change impacts on the Ztunnel image when using the Ambient profile and will not have any affect on other images.

Fixes: https://github.com/istio-ecosystem/sail-operator/issues/581

### DIFF
--- a/controllers/ztunnel/ztunnel_controller_test.go
+++ b/controllers/ztunnel/ztunnel_controller_test.go
@@ -338,7 +338,9 @@ func TestDetermineReadyCondition(t *testing.T) {
 	}
 }
 
-func PTestApplyImageDigests(t *testing.T) {
+func TestApplyImageDigests(t *testing.T) {
+	t.Skip("https://github.com/istio-ecosystem/sail-operator/issues/581")
+
 	testCases := []struct {
 		name         string
 		config       config.OperatorConfig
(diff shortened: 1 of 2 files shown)


nyechiel:
It is important to note that this only impacts the Ambient profile, and not any usage of non-Ambient mode.

controllers/ztunnel/ztunnel_controller_test.go
```diff
@@ -337,7 +337,7 @@ func TestDetermineReadyCondition(t *testing.T) {
 	}
 }
 
-func TestApplyImageDigests(t *testing.T) {
+func PTestApplyImageDigests(t *testing.T) {
```
dgn, line 340: nit: can you use t.Skip() and point to https://github.com/istio-ecosystem/sail-operator/issues/581? I think that's clearer

sridhargaddam:
@dgn, I've updated the PR. PTAL, thanks.

yxun (approved):

PR #584, opened 2025-01-24
Use logAsJson instead of logAsJSON in ZTunnel APIs

sridhargaddam:
The Ztunnel charts (as well as other Istio resources) use logAsJson and not logAsJSON.

### DIFF
--- a/api/v1/values_types_extra.go
+++ b/api/v1/values_types_extra.go
@@ -93,7 +93,7 @@ type ZTunnelConfig struct {
 	// Same as `global.logging.level`, but will override it if set
 	Logging *GlobalLoggingConfig `json:"logging,omitempty"`
 	// Specifies whether istio components should output logs in json format by adding --log_as_json argument to each container.
-	LogAsJSON *bool `json:"logAsJSON,omitempty"`
+	LogAsJson *bool `json:"logAsJson,omitempty"` //nolint
 }
 
 // ZTunnelGlobalConfig is a subset of the Global Configuration used in the Istio ztunnel chart.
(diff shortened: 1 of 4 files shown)


api/v1alpha1/values_types_extra.go
```diff
@@ -93,7 +93,7 @@ type ZTunnelConfig struct {
 	// Same as `global.logging.level`, but will override it if set
 	Logging *GlobalLoggingConfig `json:"logging,omitempty"`
 	// Specifies whether istio components should output logs in json format by adding --log_as_json argument to each container.
-	LogAsJSON *bool `json:"logAsJSON,omitempty"`
+	LogAsJson *bool `json:"logAsJson,omitempty"` //nolint
```
sridhargaddam, line 96: @luksa @dgn If I do not add `//nolint` here, I'm getting linter errors as shown below. Is this fine or do you have any suggestions?

```
INFO [runner] linters took 6.119409605s with stages: goanalysis_metalinter: 6.108544436s 
api/v1alpha1/values_types_extra.go:96:2: var-naming: struct field LogAsJson should be LogAsJSON (revive)
	LogAsJson *bool `json:"logAsJson,omitempty"`
	^
api/v1alpha1/values_types_extra.go:121:2: var-naming: struct field LogAsJson should be LogAsJSON (revive)
	LogAsJson *bool `json:"logAsJson,omitempty"`
	^
```

luksa: How about the following:
```suggestion
	LogAsJSON *bool `json:"logAsJson,omitempty"`
```

This way, we stick to go standards and also ensure that the JSON field name matches the charts. But we should also then do the same in [value_types.go](https://github.com/istio-ecosystem/sail-operator/blob/6fb050a38ca14cf5f1be050fd06fc0f59496b6e2/api/v1alpha1/values_types.gen.go#L287) and anywhere else there's a similar issue (not just for JSON, but other acronyms as well).

We should also

sridhargaddam: I'm not seeing any linter errors for the `values_types.gen.go` file (which is auto-generated as you know). I was only seeing the linter error for this file. Let me update it.

luksa (approved):

sridhargaddam:
> @dgn should we cherrypick this to the release-0.1 branch?

Added cherry-pick label as I was seeing errors while backporting the API docs PR - https://github.com/istio-ecosystem/sail-operator/pull/593#issuecomment-2619372305

PR #616, opened 2025-02-05
converter for istio in-cluster operator config to sail operator config

ctartici:
#### What type of PR is this?

- [ ] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [x] Test
- [x] Documentation Update

[OSSM-8029](https://issues.redhat.com/browse/OSSM-8029)

Fixes #

Related Issue/PR #


fjglira (approved):
LGTM adding this we will  give to the users the possibility to easily convert their configuration to sail operator Istio resource. Also, this is needed to execute the upstream Istio integration test later against Sail Operator installed control plane increasing the testing coverage. Adding a hold in case someone else want to review it

fjglira (approved):

docs/README.md
```diff
@@ -303,6 +303,22 @@ Note the quotes around the value of `spec.values.pilot.env.PILOT_ENABLE_STATUS`.
...
 Sail Operator's Istio resource does not have a `spec.components` field. Instead, you can enable and disable components directly by setting `spec.values.<component>.enabled: true/false`. Other functionality exposed through `spec.components` like the k8s overlays is not currently available.
 
+### Converter Script
+This script is used to convert an Istio in-cluster operator configuration to a Sail Operator configuration. Upon execution, a new YAML file will be created in the same directory which ISTIO-OPERATOR-CONFIG-YAML is present. The new file will be named: sail-operator-config.yaml
+
+#### Usage
+It is necessary to set ISTIO_VERSION and ISTIO_NAMESPACE as environment variables to be able to run the script.
+
+```bash
+export ISTIO_VERSION=v1.23.0
```
sridhargaddam, line 313: nit: 

```suggestion
export ISTIO_VERSION=v1.24.0
```

tools/configuration-converter.sh
```diff
@@ -0,0 +1,82 @@
...
+
+WORKDIR=$(dirname "$1")
+ISTIO_CONFIG_FILE=$(basename "$1")
+SAIL_CONFIG_FILE="sail-operator-config.yaml"
+
+cd "$WORKDIR" || exit
+cp "$ISTIO_CONFIG_FILE" "$SAIL_CONFIG_FILE"
+
+function add_mandatory_fields(){
+    sail_api=".apiVersion=\"sailoperator.io/v1alpha1\""
```
sridhargaddam, line 34: Sail Operator APIs are now v1

```suggestion
    sail_api=".apiVersion=\"sailoperator.io/v1\""
```

tools/configuration-converter.sh
```diff
@@ -0,0 +1,82 @@
...
+
+# This script is used to convert istio configuration to sail operator configuration.
+# In the end of the execution new yaml file will be created with "sail-ISTIO_CONFIG_YAML" name.
+# Usage: ./configuration-converter.sh ISTIO_CONFIG_YAML_WITH_PATH, example:  ./configuration-converter.sh sample_config.yaml"
+
+if [[ -z $ISTIO_VERSION || -z $ISTIO_NAMESPACE ]]; then
+    echo "ISTIO_VERSION and ISTIO_NAMESPACE should be set as environment variable to execute converter."
+    exit
+fi
+
```
sridhargaddam, line 25: Lets add an error if the user runs the script without supplying any arguments.

if [ -z "$1" ]; then
    echo "Usage: $0 <ISTIO_OPERATOR_CONFIG_YAML_WITH_PATH>"
    exit 1
fi

tools/configuration-converter.sh
ctartici: Done

tests/e2e/controlplane/control_plane_test.go
```diff
@@ -168,17 +171,97 @@ spec:
 					})
 				})
 
+				When("Sail Operator configuration", func() {
```
luksa, line 174: Why is this part of the e2e control plane tests?

tests/e2e/controlplane/control_plane_test.go
```diff
@@ -168,17 +171,97 @@ spec:
...
   name: default
 spec:
+  values:
+    global:
+      istiod:
+        enableAnalysis: true
+    pilot:
+      env:
+        PILOT_ENABLE_STATUS: "true"
+      enabled: true
```
luksa, line 261: Why this change? The purpose of the test was to check if Istio is deployed properly with a minimal `Istio` resource.

tools/configuration-converter.sh
```diff
@@ -0,0 +1,91 @@
...
+    sail_kind=".kind=\"Istio\""
+    sail_metadata_name=".metadata.name=\"default\""
+    sail_istio_version=".spec.version=\"$ISTIO_VERSION\""
+    sail_istio_namespace=".spec.namespace=\"$ISTIO_NAMESPACE\""
+    sail_config_array=("$sail_api" "$sail_kind" "$sail_istio_version" "$sail_istio_namespace" "$sail_metadata_name")
+    for i in "${sail_config_array[@]}";
+    do
+        cat $SAIL_CONFIG_FILE | yq "$i" > tmp.yaml
+        mv tmp.yaml $SAIL_CONFIG_FILE
+    done
```
luksa, line 51: This can be done with a single `yq` command. See example here: https://github.com/istio-ecosystem/sail-operator/blob/586e39174179f91d66fac5ec595ad4ab62342c1d/hack/download-charts.sh#L117-L127

tests/e2e/controlplane/control_plane_test.go
ctartici: Since the converter is to convert CP configuration I thought it will be appropriate to put the tests under control plane suite. What would be your advise, creating separate suite?

luksa: E2e tests take a long time, so we don't run them as often as the other tests (e.g. unit tests).  I think this could be a normal unit test, as it's testing just the script and only takes milliseconds to run.

tools/configuration-converter.sh
ctartici: This is super useful. I didn't know about hub and tag also. ty!

ctartici: Done

tests/e2e/controlplane/control_plane_test.go
ctartici: I have removed the e2e test and created unit test for the converter

pkg/converter/converter_test.go
```diff
@@ -0,0 +1,196 @@
...
+	g := NewWithT(t)
+
+	istioYamlText := `apiVersion: install.istio.io/v1alpha1
+kind: IstioOperator
+metadata:
+  name: default
+spec:`
+
+	for _, version := range supportedversion.List {
+		g.Expect(setEnvVariable("ISTIO_VERSION", version.Name)).To(Succeed(), "error adding env variable: ISTIO_VERSION")
```
luksa, line 41: Is there any benefit to having `setEnvVariable` instead of simply calling `os.Setenv` directly?

Also, since setting the env var is not part of the test's assertions, why not simplify this to something like this:
```suggestion
		must(os.Setenv("ISTIO_VERSION", version.Name))
```

Just seems like way too much code that makes the test too verbose, taking way more time to see what the test is actually doing. It's currently hard to see what is the test setup and what are the assertions without slowly reading all the code.

pkg/converter/converter_test.go
```diff
@@ -0,0 +1,196 @@
...
+      externalIstiod: true
+      operatorManageWebhooks: true
+      configValidation: false
+    base:
+      enableCRDTemplates: true
+    pilot:
+      env:
+        PILOT_ENABLE_STATUS: true`
+
+	for _, version := range supportedversion.List {
```
luksa, line 89: Is there any point in running this for all supported versions? Since the converter doesn't care about the version and we pass the version string to the converter as an argument, doesn't a single test suffice to verify the behavior?

tools/configuration-converter.sh
```diff
@@ -0,0 +1,93 @@
...
+    | .metadata.name = "default"
+    | .spec.version = strenv(ISTIO_VERSION)
+    | .spec.namespace = strenv(ISTIO_NAMESPACE)
+    | del(.spec.values.istio_cni)
+    | del(.spec.meshConfig)
+    | del(.spec.hub) 
+    | del(.spec.tag) 
+    | del(.spec.values.gateways)
+    ' "$SAIL_CONFIG_FILE" > tmp.yaml
+    mv tmp.yaml $SAIL_CONFIG_FILE
```
luksa, line 54: I think you can use `yq eval -i` to update the original file directly; no need to write to a temporary file.

tools/configuration-converter.sh
```diff
@@ -0,0 +1,93 @@
...
+}
+
+#Convert boolean values to string if they are under *.env
+function boolean_2_string(){
+    SPEC_FIELDS="$(yq -r $SAIL_CONFIG_FILE | yq '(.spec.values.[].env)')"
+    if [ -n "$SPEC_FIELDS" ] ;then
+        cat $SAIL_CONFIG_FILE | yq '(.spec.values.[].env.[] | select(. == true)) |= "true"'  > tmp.yaml
+        mv tmp.yaml $SAIL_CONFIG_FILE
+        cat $SAIL_CONFIG_FILE | yq '(.spec.values.[].env.[] | select(. == false)) |= "false"' | sed '/env: \[\]/d'  > tmp.yaml
+        mv tmp.yaml $SAIL_CONFIG_FILE
```
luksa, line 64: This, too, can probably be done with a single `yq` command.

tools/configuration-converter.sh
```diff
@@ -0,0 +1,93 @@
...
+# Note that if there is an entry except spec.components.<component>.enabled: true/false converter will delete them and warn user 
+function validate_spec_components(){
+    COMPONENTS="$(yq -r $SAIL_CONFIG_FILE 2>/dev/null | yq '(.spec.components)' 2>/dev/null)" 
+    if [ "$COMPONENTS" == null ] ;then
+        return 0    
+    else
+        COMPONENTS_FIELDS="$(yq -r $SAIL_CONFIG_FILE 2>/dev/null | yq '(.spec.components.[] | keys[])' 2>/dev/null)"
+        if [[ -n "$COMPONENTS_FIELDS" && "$COMPONENTS_FIELDS" =~ "enabled" ]] ;then
+            cat $SAIL_CONFIG_FILE | yq 'del(.spec.components.[] | keys[] | select(. != "enabled"))' | yq '.spec.values *= .spec.components | del (.spec.components)' > tmp.yaml 
+            mv tmp.yaml $SAIL_CONFIG_FILE
```
luksa, line 77: Also could be a single `yq` command (+ no need for a temporary file).

tools/configuration-converter.sh
```diff
@@ -0,0 +1,93 @@
...
+            echo "Converter can only be used values with spec.components.<component>.enabled: true/false. Please see https://github.com/istio-ecosystem/sail-operator/tree/main/docs#components-field"
+        else
+            cat $SAIL_CONFIG_FILE | yq 'del (.spec.components)' > tmp.yaml  
+            mv tmp.yaml $SAIL_CONFIG_FILE
```
luksa, line 81:
```suggestion
            yq -i 'del (.spec.components)' "$SAIL_CONFIG_FILE"
```

docs/README.md
```diff
@@ -303,6 +303,22 @@ Note the quotes around the value of `spec.values.pilot.env.PILOT_ENABLE_STATUS`.
...
+### Converter Script
+This script is used to convert an Istio in-cluster operator configuration to a Sail Operator configuration. Upon execution, a new YAML file will be created in the same directory which ISTIO-OPERATOR-CONFIG-YAML is present. The new file will be named: sail-operator-config.yaml
+
+#### Usage
+It is necessary to set ISTIO_VERSION and ISTIO_NAMESPACE as environment variables to be able to run the script.
+
+```bash
+export ISTIO_VERSION=v1.24.0
+export ISTIO_NAMESPACE=istio-system
+./tools/configuration-converter.sh <ISTIO_OPERATOR_CONFIG_YAML_WITH_PATH>
```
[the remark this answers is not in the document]
ctartici: This script will be downloaded and executed from [integ-suite-ocp.sh](https://github.com/openshift-service-mesh/istio/blob/master/prow/integ-suite-ocp.sh) We are planning to run downstream integration tests with it and in there these env variables are already defined. Also since integ-suite-ocp.sh already gets many arguments(tests suite to run, tests to skip) It would have been so confusing.

pkg/converter/converter_test.go
ctartici: From [readme](https://github.com/istio-ecosystem/sail-operator/blob/main/docs/README.md#components-field) I get a notion that in future some more istio parameters might be supported in sail operator. That is why I have added in a loop to check in all versions. It might be removed for now.

docs/README.md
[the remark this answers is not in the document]
luksa: OK, but can't `integ-suite-ocp.sh` simply pass the environment variables as arguments to the converter? It's weird that a shell script takes one required input as an argument, and other required inputs as an env var. 

Also, won't users try to use this script as well? To convert examples from the upstream docs, for example.

tools/configuration-converter.sh
ctartici: yq -i is not working in some directories like /tmp if [installed with snap](https://github.com/mikefarah/yq?tab=readme-ov-file#snap-notes). /tmp is default directory for integration test execution. That is why I create a temporary file.

pkg/converter/converter_test.go
ctartici: Modified tests to use latest version

tools/configuration-converter.sh
ctartici: implemented a workaround since yq -i is not working on root directories

tools/configuration-converter.sh
ctartici: implemented a workaround since yq -i is not working on root directories

tools/configuration-converter.sh
ctartici: Done

tools/configuration-converter.sh
```diff
@@ -0,0 +1,77 @@
...
+WORKDIR=$(dirname "$1")
+ISTIO_CONFIG_FILE=$(basename "$1")
+ISTIO_VERSION="$2"
+ISTIO_NAMESPACE="$3"
+SAIL_CONFIG_FILE="sail-operator-config.yaml"
+
+cp "$WORKDIR"/"$ISTIO_CONFIG_FILE" "$SAIL_CONFIG_FILE" || exit 1
+
+function add_mandatory_fields(){
+    yq -i eval ".apiVersion = \"sailoperator.io/v1\"
```
MaxBab, line 37: I would add a check for the `yq` binary installed.
Otherwise, the command will fail without specifying a reason.

tools/configuration-converter.sh
```diff
@@ -0,0 +1,77 @@
...
+function boolean_2_string(){
+    yq -i -e '(.spec.values.[].env.[] | select(. == true)) |= "true" 
+    | (.spec.values.[].env.[] | select(. == false)) |= "false"' "$SAIL_CONFIG_FILE" 
+    yq -i -e 'del(.. | select(tag == "!!seq" and length == 0))' "$SAIL_CONFIG_FILE"
+}
+
+# Note that if there is an entry except spec.components.<component>.enabled: true/false converter will delete them and warn user 
+function validate_spec_components(){
+    yq -i 'del(.spec.components.[] | keys[] | select(. != "enabled")) | .spec.values *= .spec.components | del (.spec.components)' $SAIL_CONFIG_FILE
+    echo "Converter can only be used values with spec.components.<component>.enabled: true/false. Please see https://github.com/istio-ecosystem/sail-operator/tree/main/docs#components-field"
```
MaxBab, line 61: Converted can only use values with....

docs/README.md
```diff
@@ -303,6 +303,25 @@ Note the quotes around the value of `spec.values.pilot.env.PILOT_ENABLE_STATUS`.
...
 
+### Converter Script
+This script is used to convert an Istio in-cluster operator configuration to a Sail Operator configuration. Upon execution, a new YAML file will be created in the same directory which ISTIO-OPERATOR-CONFIG-YAML is present. The new file will be named: sail-operator-config.yaml
+
+#### Usage
+To run the configuration-converter.sh script, you need to provide three arguments:
+
+ISTIO_OPERATOR_CONFIG_YAML_WITH_PATH: Path to the Istio operator configuration YAML file.
+ISTIO_NAMESPACE: The Kubernetes namespace for the Istio deployment.
+ISTIO_VERSION: The version of Istio to be used.
```
luksa, line 314: Since the `ISTIO_OPERATOR_CONFIG_YAML_WITH_PATH`, `ISTIO_NAMESPACE`, and `ISTIO_VERSION` aren't shown in the example, you can probably just remove them here and just explain the arguments (in the correct order):
```suggestion
- The path to the Istio operator configuration YAML file.
- The Kubernetes namespace for the Istio deployment.
- The version of Istio to be used.
```

docs/README.md
```diff
@@ -303,6 +303,25 @@ Note the quotes around the value of `spec.values.pilot.env.PILOT_ENABLE_STATUS`.
 
 Sail Operator's Istio resource does not have a `spec.components` field. Instead, you can enable and disable components directly by setting `spec.values.<component>.enabled: true/false`. Other functionality exposed through `spec.components` like the k8s overlays is not currently available.
 
+### Converter Script
+This script is used to convert an Istio in-cluster operator configuration to a Sail Operator configuration. Upon execution, a new YAML file will be created in the same directory which ISTIO-OPERATOR-CONFIG-YAML is present. The new file will be named: sail-operator-config.yaml
```
luksa, line 307:
```suggestion
This script is used to convert an Istio in-cluster operator configuration to a Sail Operator configuration. Upon execution, a new YAML file will be created in the same directory as the source YAML file. The new file will be named: sail-operator-config.yaml
```

luksa: Also, wouldn't it be better to generate the new file name from the original file name, and just adding `sail.yaml` as the suffix? If someone has multiple files to convert, it would be a hassle to have to rename the output file every time. And if they don't rename it, the file gets overwritten; what if the person doesn't notice?

pkg/converter/converter_test.go
```diff
@@ -0,0 +1,176 @@
...
+	"github.com/istio-ecosystem/sail-operator/pkg/test/project"
+	"github.com/istio-ecosystem/sail-operator/pkg/test/util/supportedversion"
+	"github.com/istio-ecosystem/sail-operator/tests/e2e/util/shell"
+	. "github.com/onsi/gomega"
+)
+
+var (
+	controlPlaneNamespace = "istio-system"
+	// Test converter with latest istio version
+	istioVersion = supportedversion.List[len(supportedversion.List)-1].Name
```
luksa, line 32: I believe there's a `supportedversion.Default` or something like that? No need to get the version from the list.

pkg/converter/converter_test.go
```diff
@@ -0,0 +1,176 @@
...
+)
+
+func TestSimpleYamlConversion(t *testing.T) {
+	g := NewWithT(t)
+
+	istioYamlText := `apiVersion: install.istio.io/v1alpha1
```
luksa, line 38: Can rename this to `istioYaml`. Same for `sailYamlText`. Shorter names are usually better, especially if the variable's context is very small. It makes reading the code faster.

pkg/converter/converter_test.go
```diff
@@ -0,0 +1,176 @@
...
+	g.Expect(executeConfigConverter(istioYamlFileWithPath, controlPlaneNamespace, istioVersion)).To(Succeed(),
+		"error in execution of ./configuration-converter.sh")
+	isConversionValidated, err := validateYamlContent(sailYamlText)
+	g.Expect(err).To(Succeed(), "Can not open file to compare")
+	g.Expect(isConversionValidated).To(BeTrue(), "Converted content is not as expected")
+
+	err = os.Remove(filepath.Join(project.RootDir, "tools", "istioConfig.yaml"))
+	g.Expect(err).To(Succeed(), "Unable to delete istioConfig.yaml")
+	err = os.Remove(filepath.Join(project.RootDir, "tools", "sail-operator-config.yaml"))
+	g.Expect(err).To(Succeed(), "Unable to delete sail-operator-config.yaml")
```
luksa, line 65: Does this get executed if the assertions above fail?

pkg/converter/converter_test.go
```diff
@@ -0,0 +1,176 @@
...
+}
+
+func TestComplexYamlConversion(t *testing.T) {
```
luksa, line 68: Since this and the previous test both test the same function, it would be better if both tests were implemented in the same test function, with table-based sub-tests. Something like:
```
func TestConversion(t *testing.T) {
    var flagtests = []struct {
        name  string
        input string
        expectedOutput string
    }{
        {"simple", "simple input yaml", "simple output yaml"},
        {"complex", "complex input yaml", "complex output yaml"},
    }    

    for _, tt := range testcases {
        t.Run(tt.name, func(t *testing.T) {
            ...
        })
    }
}
```

This would reduce code duplication and allow the next person to easily add another sub-test.

pkg/converter/converter_test.go
```diff
@@ -0,0 +1,176 @@
...
+  version: %s
+  namespace: %s`
+	sailYamlText = fmt.Sprintf(sailYamlText, istioVersion, controlPlaneNamespace)
+
+	istioYamlFileWithPath, err := saveYamlToFile(istioYamlText)
+	g.Expect(err).To(Succeed(), "failed to write YAML file")
+
+	g.Expect(executeConfigConverter(istioYamlFileWithPath, controlPlaneNamespace, istioVersion)).To(Succeed(),
+		"error in execution of ./configuration-converter.sh")
+	isConversionValidated, err := validateYamlContent(sailYamlText)
```
luksa, line 58: One thing I don't like about the `validateYamlContent` function is that it isn't clear what it's comparing to the `sailYamlText`. It would be clearer if you passed in both YAMLs. Then, when someone sees `validateYamlContent(actual, expected)`, it would immediately clear what the function is doing. Right now, seeing this function causes confusion.

tools/configuration-converter.sh
ctartici: It will fail with yq command not found. You think I need to catch command not found and inform user with better error message?

tests/e2e/util/shell/shell.go
```diff
@@ -53,3 +54,34 @@ func ExecuteShell(command string, input string) (string, error) {
 
 	return stdout.String(), nil
 }
+
+// Execute bash script with optional arguments
+func ExecuteBashScript(scriptPath string, args ...string) (string, error) {
```
luksa, line 59: Why do we need this function? Doesn't `ExecuteShell` have everything we need already?

tools/configuration-converter.sh
```diff
@@ -0,0 +1,77 @@
...
+    exit 1
+fi
+
+WORKDIR=$(dirname "$1")
+ISTIO_CONFIG_FILE=$(basename "$1")
+ISTIO_VERSION="$2"
+ISTIO_NAMESPACE="$3"
+SAIL_CONFIG_FILE="sail-operator-config.yaml"
+
+cp "$WORKDIR"/"$ISTIO_CONFIG_FILE" "$SAIL_CONFIG_FILE" || exit 1
```
luksa, line 34: I'd move this after the functions. Currently, we copy the file here, but then modify it in a totally different place in the script. It would be best to keep all the steps together, possibly even in a function that is then called at the end of the script.

tools/configuration-converter.sh
MaxBab: You could add the following function.
It will check for the `yq` command available.
And if missing, will download it and use.

```
function verify_yq() {
    if ! command -v yq &> /dev/null; then
        if [[ "${OS}" == "darwin" ]]; then
            ERROR "Perform 'brew install yq' and try again."
        elif [[ "${OS}" == "linux" ]]; then
            WARNING "Missing yq command. Installing..."
            mkdir -p "$HOME"/.local/bin
            wget -qO- https://github.com/mikefarah/yq/releases/download/v4.16.2/yq_linux_amd64 \
                -O "$HOME"/.local/bin/yq && chmod +x "$HOME"/.local/bin/yq
            
            # Add local BIN dir to PATH
            [[ ":$PATH:" = *":$HOME/.local/bin:"* ]] || export PATH="$HOME/.local/bin:$PATH"
        fi
        INFO "The yq command installed"
    fi
    INFO "The yq command is found"
}
```

Or just check and explicitly tell the user what is missing.

tools/configuration-converter.sh
```diff
@@ -0,0 +1,77 @@
...
+
+}
+
+add_mandatory_fields
+boolean_2_string
+if [[ $(yq eval '.spec.components' "$SAIL_CONFIG_FILE") != "null" ]]; then
+    validate_spec_components
+fi 
+
+chmod +x $SAIL_CONFIG_FILE
```
luksa, line 71: Is this really necessary?

tools/configuration-converter.sh
```diff
@@ -0,0 +1,77 @@
...
+
+if ! mv "$SAIL_CONFIG_FILE" "$WORKDIR" 2>&1 | grep -q "are identical"; then
```
luksa, line 73: Is it a good idea to assume what the output text can be?

tools/configuration-converter.sh
ctartici: It is to be able to use created yaml in oc/kubectl command afterwards. oc apply -f sail-config.yaml

tools/configuration-converter.sh
ctartici: Done now using yq -i

tools/configuration-converter.sh
ctartici: Done now using yq -i

tools/configuration-converter.sh
ctartici: Done. added an error message for yq installation

tools/configuration-converter.sh
ctartici: Done. changed the sentence

docs/README.md
ctartici: Done modified readme

docs/README.md
ctartici: Done. added an optional argument for output file. If not present it will add -sail.yaml suffix to input file

pkg/converter/converter_test.go
ctartici: Done

pkg/converter/converter_test.go
ctartici: Done

pkg/converter/converter_test.go
ctartici: Added a cleanup function for removing created files.

pkg/converter/converter_test.go
ctartici: Done

tests/e2e/util/shell/shell.go
ctartici: Done. removed function

tools/configuration-converter.sh
ctartici: Done

tools/configuration-converter.sh
ctartici: Done. removed command since I couldn't reproduce the error.

tools/configuration-converter.sh
ctartici: Removed after adding output as argument

docs/README.md
```diff
@@ -304,18 +304,30 @@ Note the quotes around the value of `spec.values.pilot.env.PILOT_ENABLE_STATUS`.
...
 
-ISTIO_OPERATOR_CONFIG_YAML_WITH_PATH: Path to the Istio operator configuration YAML file.
-ISTIO_NAMESPACE: The Kubernetes namespace for the Istio deployment.
-ISTIO_VERSION: The version of Istio to be used.
+##### Sample command with custom output and namespace:
 
-##### Example
 ```bash
-./tools/configuration-converter.sh tools/istio_config.yaml istio-system v1.24.3
+./tools/configuration-converter.sh -i /home/user/input/istio_config.yaml -o /home/user/output/output.yaml -n custom-namespace -v v1.24.3
```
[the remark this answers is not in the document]
ctartici: Done. I have removed -i and -o arguments and updated readme

pkg/converter/converter_test.go
```diff
@@ -0,0 +1,191 @@
...
+	istioVersion          = supportedversion.Default
+	istioFile             = filepath.Join(project.RootDir, "tools", "istioConfig.yaml")
+	sailFile              = filepath.Join(project.RootDir, "tools", "istioConfig-sail.yaml")
+)
+
+func TestConversion(t *testing.T) {
+	t.Cleanup(func() {
+		os.Remove(istioFile)
+		os.Remove(sailFile)
+	})
```
luksa, line 41: Probably more appropriate to move this into the subtest.

pkg/converter/converter_test.go
```diff
@@ -0,0 +1,191 @@
...
+			g := NewWithT(t)
+			err := saveYamlToFile(tc.input, istioFile)
+			g.Expect(err).To(Succeed(), "failed to write YAML file")
+
+			if tc.name == "mandatory-arguments-only" {
+				g.Expect(executeConfigConverter(istioFile, "", "", istioVersion)).To(Succeed(), "error in execution of ./configuration-converter.sh")
+			} else {
+				g.Expect(executeConfigConverter(istioFile, sailFile, controlPlaneNamespace,
+					istioVersion)).To(Succeed(), "error in execution of ./configuration-converter.sh")
+			}
```
luksa, line 131: It would be cleaner if the file, namespace, and version were all taken from each test case. This way, you can see all the input arguments in the same place (for each test case). This makes it easier to review what each test is doing exactly.

Also, it's not the best idea to change the test behavior based on the subtest name. We usually add an additional field to the testcase struct, so that the different behavior is more explicit/visible. But in this case, if you simply do what I explained above, you won't need the additional field.
```suggestion
			g.Expect(executeConfigConverter(tc.input, tc.outputFile, tc.namespace,
				tc.version)).To(Succeed(), "error in execution of ./configuration-converter.sh")
```

luksa: Or, maybe instead of having the `namespace`, `version`, and such fields in the testcase struct, maybe a simple string that holds the (additional?) arguments is better, since the purpose of the test is to test a shell script. For example: `tc.args = "-n somenamespace -v version". But I don't know what we should do with the input and output files (should they be passed in the same `tc.args` or not. I'll let you figure this out yourelf.

pkg/converter/converter_test.go
```diff
@@ -0,0 +1,191 @@
...
+	output, err := shell.ExecuteCommand(cmd)
+	if err != nil {
+		return false, fmt.Errorf("error executing yq comparison: %w", err)
+	}
+	if output != "" {
+		return false, nil
+	}
+	// If no output from yq, the files are considered equal
+	return true, nil
+}
```
luksa, line 191: There's a better way to do this. You can use the `go-cmp` library to compare two objects (perhaps even strings?). We already do this in other tests; for example: https://github.com/istio-ecosystem/sail-operator/blob/8326acf08ecc509fc93eea29f581582903076e41/controllers/istio/istio_controller_test.go#L544

tests/e2e/util/shell/shell.go
```diff
@@ -34,7 +34,7 @@ func ExecuteCommand(command string) (string, error) {
 
 // ExecuteShell executes a command given the input string and returns the output and err if any
 func ExecuteShell(command string, input string) (string, error) {
-	cmd := exec.Command("sh", "-c", command)
+	cmd := exec.Command("bash", "-c", command)
```
luksa, line 37: Is this change needed? I'm not completely sure if this could break any of our test jobs. What if there's no bash on some of the systems we run these tests on? Do you really need to switch this to bash?

tools/configuration-converter.sh
```diff
@@ -0,0 +1,126 @@
...
+        *)
+            usage
+            ;;
+    esac
+done
+
+if [[ -z "$VERSION" ]]; then
+    echo "Error: Version (-v) is required."
+    usage
+fi
```
luksa, line 81: Version should not be required.

tools/configuration-converter.sh
```diff
@@ -0,0 +1,126 @@
...
+    shift 1
+fi
+
+# Ensure output is not a directory
+if [[ -d "$OUTPUT" ]]; then 
+    echo "Error: OUTPUT must be a file, not a directory."
+    exit 1
+fi
+
+# Parse optional namespace and version arguments
```
luksa, line 61: How hard is it to modify this script to allow specifying the arguments in any order (e.g. `converter.sh -n namespace input-file.yaml -v version`)?

tools/configuration-converter.sh
```diff
@@ -0,0 +1,126 @@
...
+    echo "Only values in the format spec.components.<component>.enabled: true/false are supported for conversion. For more details, refer to the documentation: https://github.com/istio-ecosystem/sail-operator/tree/main/docs#components-field"
+}
+
+# create output file
+cp "$INPUT" "$OUTPUT"
+
+# in-place edit created output file  
+add_mandatory_fields
+boolean_2_string
+if [[ $(yq eval '.spec.components' "$OUTPUT") != "null" ]]; then
```
luksa, line 122: I'd move this if statement into `validate_spec_components`, since the if condition contains business logic related to the one within that function. It's best to keep related code close together and not leak lower abstractions to a higher abstraction level, like we're doing here.

tools/configuration-converter.sh
```diff
@@ -0,0 +1,126 @@
...
+    | del(.spec.hub) 
+    | del(.spec.tag) 
+    | del(.spec.values.gateways)" "$OUTPUT"
+}
+
+#Convert boolean values to string if they are under *.env
+function boolean_2_string(){
+    yq -i -e '(.spec.values.[].env.[] | select(. == true)) |= "true" 
+    | (.spec.values.[].env.[] | select(. == false)) |= "false"' "$OUTPUT" 
+    yq -i -e 'del(.. | select(tag == "!!seq" and length == 0))' "$OUTPUT"
```
luksa, line 107: I like how this script is so much simpler now. Can we merge these two commands into one?

docs/README.md
```diff
@@ -303,6 +303,37 @@ Note the quotes around the value of `spec.values.pilot.env.PILOT_ENABLE_STATUS`.
...
+#### Usage
+To run the configuration-converter.sh script, you need to provide four arguments, only input file is required other arguments are optional:
+
+1. Input file path (<input>): The path to your Istio operator configuration YAML file (required).
+2. Output file path (<output>): The path where the converted Sail configuration will be saved. If not provided, the script will save the output with -sail.yaml appended to the input file name.
+3. Namespace (-n <namespace>): The Kubernetes namespace for the Istio deployment. Defaults to istio-system if not provided.
+4. Version (-v <version>): The version of Istio to be used. If not provided, the `spec.version` field will be omitted from the output file and the operator will deploy the latest version when the YAML manifest is applied.
+
+```bash
+./tools/configuration-converter.sh /path/to/input.yaml [optional-output.yaml] -n [optional-namespace] -v [optional-version]
```
luksa, line 318: Small correction needed:
```suggestion
./tools/configuration-converter.sh /path/to/input.yaml [optional-output.yaml] [-n optional-namespace] [-v optional-version]
```

pkg/converter/converter_test.go
```diff
@@ -0,0 +1,196 @@
...
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/istio-ecosystem/sail-operator/pkg/test/project"
+	"github.com/istio-ecosystem/sail-operator/pkg/test/util/supportedversion"
+	"github.com/istio-ecosystem/sail-operator/tests/e2e/util/shell"
+	. "github.com/onsi/gomega"
+	"gopkg.in/yaml.v2"
```
luksa, line 29: Can we use v3 instead of v2? We use v3 everywhere else.

```suggestion
	"gopkg.in/yaml.v3"
```

pkg/converter/converter_test.go
```diff
@@ -0,0 +1,196 @@
...
+	istioFile             = filepath.Join(project.RootDir, "tools", "istioConfig.yaml")
+	sailFile              = filepath.Join(project.RootDir, "tools", "istioConfig-sail.yaml")
+)
+
+func TestConversion(t *testing.T) {
+	testcases := []struct {
+		name           string
+		input          string
+		expectedOutput string
+		converterArgs  string
```
luksa, line 44: Maybe switch the order of these two, so that both input and converterArgs come first, then the expected output.

pkg/converter/converter_test.go
```diff
@@ -0,0 +1,196 @@
...
+  name: default
+spec:`,
+			expectedOutput: fmt.Sprintf(`apiVersion: sailoperator.io/v1
+kind: Istio
+metadata:
+  name: default
+spec:
+  version: %s
+  namespace: %s`, istioVersion, controlPlaneNamespace),
+			converterArgs: fmt.Sprintf("%s %s -n %s -v %s", istioFile, sailFile, controlPlaneNamespace, istioVersion),
```
luksa, line 60: Is there an actual benefit in using variables instead of using the simpler approach of just including the values in the YAML and args? The following takes much less effort to read:
```suggestion
			expectedOutput: `apiVersion: sailoperator.io/v1
kind: Istio
metadata:
  name: default
spec:
  version: v1.24.0
  namespace: my-namespace`,
			converterArgs: "/tmp/input.yaml /tmp/output.yaml -n my-namespace -v v1.24.0",
```

pkg/converter/converter_test.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+	for _, tc := range testcases {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Cleanup(func() {
+				os.Remove(istioFile)
+				os.Remove(sailFile)
+			})
+			g := NewWithT(t)
+			err := saveYamlToFile(tc.input, istioFile)
+			g.Expect(err).To(Succeed(), "failed to write YAML file")
```
luksa, line 129: Why not simply:
```suggestion
			g.Expect(os.WriteFile(istioFile, []byte(tc.input)).To(Succeed(), "failed to write YAML file")
```

pkg/converter/converter_test.go
```diff
@@ -0,0 +1,196 @@
...
+
+			g.Expect(executeConfigConverter(tc.converterArgs)).To(Succeed(), "error in execution of ./configuration-converter.sh")
+
+			isConversionValidated, err := validateYamlContent(sailFile, tc.expectedOutput)
+			g.Expect(err).To(Succeed(), fmt.Errorf("Error on validation: %w", err))
+			g.Expect(isConversionValidated).To(BeTrue(), "Converted content is not as expected")
```
luksa, line 135: I believe this can be simplified to the following (the error is handled automatically).
```suggestion
			g.Expect(validateYamlContent(sailFile, tc.expectedOutput)).To(BeTrue(), "Converted content is not as expected")
```

pkg/converter/converter_test.go
```diff
@@ -0,0 +1,196 @@
...
+
+func executeConfigConverter(scriptArgs string) error {
+	converter := filepath.Join(project.RootDir, "tools", "configuration-converter.sh")
+
+	cmdArgs := []string{converter, scriptArgs}
+
+	cmd := strings.Join(cmdArgs, " ")
+	output, err := shell.ExecuteCommand(cmd)
+	if err != nil {
+		fmt.Printf("Script output: %s", output)
```
luksa, line 156: I think you forgot to remove this.

pkg/converter/converter_test.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+// parseYaml takes a YAML string and unmarshals it into a map
+func parseYaml(yamlContent string) (map[string]interface{}, error) {
+	var config map[string]interface{}
+	err := yaml.Unmarshal([]byte(yamlContent), &config)
+	if err != nil {
+		return nil, err
+	}
+	return config, nil
```
luksa, line 195: Can be simplified to: 
```suggestion
	return config, err
```

pkg/converter/converter_test.go
```diff
@@ -0,0 +1,196 @@
...
+		return false, fmt.Errorf("failed to parse expected output: %w", err)
+	}
+
+	// Compare YAML content ignoring order
+	diff := cmp.Diff(expectedData, actualData)
+	if diff != "" {
+		fmt.Println("YAML files differ:", diff)
+		return false, nil
+	}
+	return true, nil
```
luksa, line 185: I think you just need to return diff here, without logging it. Then let `g.Expect(getDiff(...)).To(Equal(""))` print the diff if it isn't empty.

Note: `validateYamlContent` would be renamed to `getDiff` in this case.

pkg/converter/converter_test.go
```diff
@@ -0,0 +1,196 @@
...
+	output, err := shell.ExecuteCommand(cmd)
+	if err != nil {
+		fmt.Printf("Script output: %s", output)
+		return fmt.Errorf("error in execution of command %v: %w", cmdArgs, err)
+	}
+	return nil
+}
+
+// compareYamlContent checks if the provided YAML content matches the content of converted sail operator config
+func validateYamlContent(sailFile, expectedOutput string) (bool, error) {
```
luksa, line 163: I would actually get rid of this function and just inline the code in the test itself. If you replace all the ifs with g.Expect(), this would end up being just 3 lines in the test function.

pkg/converter/converter_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+func executeConfigConverter(scriptArgs string) error {
+	converter := filepath.Join(project.RootDir, "tools", "configuration-converter.sh")
+
+	cmdArgs := []string{converter, scriptArgs}
+
+	cmd := strings.Join(cmdArgs, " ")
+	_, err := shell.ExecuteCommand(cmd)
+	if err != nil {
+		return fmt.Errorf("error in execution of command %v: %w", cmdArgs, err)
```
luksa, line 161: Instead of printing the command here, I'd add the command to the error message in `ExecuteCommand`. Then this entire if statement could be replaced with a simple `return err`.

When wrapping errors, always check what information is provided by the function you're calling. Then just add any additional information that the called function doesn't have. In this case, there's really no such information. You could wrap the error and add a message saying "failed to execute config converter", but of `ExecuteCommand` already adds the full command to the error message, then it'll already contain the string `"configuration-conevrter.sh"`, so it's clear what the error is about.

pkg/converter/converter_test.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+}
+
+func executeConfigConverter(scriptArgs string) error {
+	converter := filepath.Join(project.RootDir, "tools", "configuration-converter.sh")
+
+	cmdArgs := []string{converter, scriptArgs}
+
+	cmd := strings.Join(cmdArgs, " ")
+	_, err := shell.ExecuteCommand(cmd)
```
luksa, line 159: Can be simplified to:
```suggestion
	_, err := shell.ExecuteCommand(converter + " " + scriptArgs)
```

tools/configuration-converter.sh
```diff
@@ -0,0 +1,140 @@
...
+    # Execute the yq command with the constructed string
+    yq -i eval "$yq_cmd" "$OUTPUT"    
+}
+
+
+#Convert boolean values to string if they are under *.env
+function boolean_2_string(){
+    yq -i -e '(.spec.values.[].env.[] | select(. == true)) |= "true" 
+    | (.spec.values.[].env.[] | select(. == false)) |= "false"' "$OUTPUT"
+    yq -i -e 'del(.. | select(tag == "!!seq" and length == 0))' "$OUTPUT"
```
luksa, line 120: Still two separate commands? :)

Maybe I missed your comment explaining why it can't be a single command.

tools/configuration-converter.sh
```diff
@@ -0,0 +1,140 @@
...
+            | del(.spec.tag) 
+            | del(.spec.values.gateways)"
+    
+    # If VERSION is not empty, add .spec.version
+    if [[ -n "$VERSION" ]]; then
+        yq_cmd="$yq_cmd | .spec.version = \"$VERSION\""
+    fi
+
+    # Execute the yq command with the constructed string
+    yq -i eval "$yq_cmd" "$OUTPUT"
```
luksa, line 112: I think running an additional yq command just for the version would make the code simpler here. Since there's no need to optimize the number of yq executions, code simplicity is the main factor to consider.

tools/configuration-converter.sh
```diff
@@ -83,41 +83,37 @@ while [[ $# -gt 0 ]]; do
...
-            | del(.spec.meshConfig)
-            | del(.spec.hub) 
-            | del(.spec.tag) 
-            | del(.spec.values.gateways)"
-    
     # If VERSION is not empty, add .spec.version
     if [[ -n "$VERSION" ]]; then
-        yq_cmd="$yq_cmd | .spec.version = \"$VERSION\""
+        yq -i ".spec.version = \"$VERSION\"" "$OUTPUT"  
     fi
```
luksa, line 95: nit: I'd prefer to see this *after* the main yq command, as this here is just an additional step; the main part of this function is the large yq command, so it should come first.

pkg/converter/converter_test.go
```diff
@@ -0,0 +1,160 @@
...
+	istioFile = filepath.Join(project.RootDir, "tools", "istioConfig.yaml")
+	sailFile  = filepath.Join(project.RootDir, "tools", "istioConfig-sail.yaml")
+)
+
+func TestConversion(t *testing.T) {
+	testcases := []struct {
+		name           string
+		input          string
+		expectedOutput string
+		converterArgs  string
```
luksa, line 40:
```suggestion
		converterArgs  string
		expectedOutput string
```

pkg/converter/converter_test.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+}
+
+func executeConfigConverter(scriptArgs string) error {
+	converter := filepath.Join(project.RootDir, "tools", "configuration-converter.sh")
+
+	if _, err := shell.ExecuteCommand(converter + " " + scriptArgs); err != nil {
+		return fmt.Errorf("error in execution of command %w", err)
+	}
+	return nil
```
luksa, line 152: The wrapped error doesn't add any new information, so there's no need to wrap the original error. This can be simplified to:
```suggestion
	_, err := shell.ExecuteCommand(converter + " " + scriptArgs)
	return err
```

tests/e2e/util/shell/shell.go
```diff
@@ -48,7 +48,7 @@ func ExecuteShell(command string, input string) (string, error) {
 	err := cmd.Run()
 	if err != nil {
 		// Return both stdout and stderr to help with debugging
-		return stdout.String() + stderr.String(), fmt.Errorf("error executing command: %s", stderr.String())
+		return stdout.String() + stderr.String(), fmt.Errorf("error executing command: %s", cmd)
```
luksa, line 51: Both the command and its error output should be included in the error:
```suggestion
		return stdout.String() + stderr.String(), fmt.Errorf("error executing command: %s: %s", cmd, stderr.String())
```

tools/configuration-converter.sh
```diff
@@ -0,0 +1,136 @@
...
+if ! command -v yq &>/dev/null; then
+    echo "Error: 'yq' is not installed. Please install it before running the script."
+    exit 1
+fi
+
+function add_mandatory_fields(){
+    # If VERSION is not empty, add .spec.version
+    if [[ -n "$VERSION" ]]; then
+        yq -i ".spec.version = \"$VERSION\"" "$OUTPUT"  
+    fi
```
luksa, line 95: I'd move this to the bottom of this function. The most important part is that large yq command, so it makes sense for that to be the first thing we do. Adding the version is just an additional step. We should always try to keep the most important part of the function at the top, because it makes searching through the code faster (you immediately see the main point of a function without having to read the entire thing).

tools/configuration-converter.sh
```diff
@@ -0,0 +1,136 @@
...
+    | del(.spec.values.gateways)" "$OUTPUT" 
+
+}
+
+#Convert boolean values to string if they are under *.env
+function boolean_2_string(){
+    yq -i -e '
+    (.spec.values.[].env.[] | select(. == true)) |= "true" |
+    (.spec.values.[].env.[] | select(. == false)) |= "false" |
+    del(.. | select(length == 0))' "$OUTPUT"
```
luksa, line 116: What does the last line do? Can you add a comment?

tools/configuration-converter.sh
```diff
@@ -0,0 +1,136 @@
...
+INPUT=""
+OUTPUT=""
+NAMESPACE="istio-system"
+VERSION=""
+
+# Identify the first and second file paths (treat the first file as INPUT)
+for arg in "$@"; do
+    if [[ -f "$arg" && -z "$INPUT" ]]; then
+        INPUT="$arg"  # First file is the INPUT
+    elif [[ -f "$arg" && -z "$OUTPUT" ]]; then
```
luksa, line 47: Will this work when the output file doesn't exist yet? Is there a test for this?

tests/e2e/util/shell/shell.go
```diff
@@ -48,7 +48,7 @@ func ExecuteShell(command string, input string) (string, error) {
 	err := cmd.Run()
 	if err != nil {
 		// Return both stdout and stderr to help with debugging
-		return stdout.String() + stderr.String(), fmt.Errorf("error executing command: %s", stderr.String())
+		return stdout.String() + stderr.String(), fmt.Errorf("error executing command: %s: %s", cmd, stderr.String())
```
sridhargaddam, line 51:
```suggestion
		return stdout.String() + stderr.String(), fmt.Errorf("error executing command: %s: %s", cmd.String(), stderr.String())
```

tools/configuration-converter.sh
```diff
@@ -0,0 +1,133 @@
...
+# This script is used to convert istio configuration to sail operator configuration.
+# In the end of the execution new yaml file will be created with "sail-ISTIO_CONFIG_YAML" name.
+# Usage: ./configuration-converter.sh ISTIO_CONFIG_YAML_WITH_PATH, example:  ./configuration-converter.sh sample_config.yaml"
+set -e
+
+# Function to show usage
+usage() {
+    echo "Usage: $0 <input> [output] [-n <namespace>] [-v <version>]"
+    echo "  <input>         : Input file (required)"
+    echo "  [output]        : Output file (optional, defaults to input's base directory with '-sail.yaml' suffix)"
```
sridhargaddam, line 26:
```suggestion
    echo "  [output]        : Output file (optional, defaults to input's file name with '-sail.yaml' suffix)"
```

tools/configuration-converter.sh
```diff
@@ -0,0 +1,133 @@
...
+elif [[ -d "$OUTPUT" ]]; then
+    echo "Error: OUTPUT must be a file, not a directory."
+    exit 1
+fi
+
+# Parse optional namespace (-n) and version (-v) arguments
+while [[ $# -gt 0 ]]; do
+    case "$1" in
+        -n)
+            NAMESPACE="$2"
```
sridhargaddam, line 70: Nit:  
If the user accidentally supplies "-n" without the value (similarly for version), it will not work as we are not checking the value of $2.

tools/configuration-converter.sh
luksa: This if statement doesn't work properly. Here's what happens:
```
[luksa@p1 ctartici/main:sail-operator]$ ls test2.yaml
ls: cannot access 'test2.yaml': No such file or directory

[luksa@p1 ctartici/main:sail-operator]$ ./tools/configuration-converter.sh test1.yaml test2.yaml
Only values in the format spec.components.<component>.enabled: true/false are supported for conversion. For more details, refer to the documentation: https://github.com/istio-ecosystem/sail-operator/tree/main/docs#components-field
Sail configuration file created with name: /home/luksa/projects/jboss/maistra/sail-operator/test1-sail.yaml

[luksa@p1 ctartici/main:sail-operator]$ ls test2.yaml
ls: cannot access 'test2.yaml': No such file or directory

[luksa@p1 ctartici/main:sail-operator]$ ls test1-sail.yaml
test1-sail.yaml
```

tools/configuration-converter.sh
```diff
@@ -0,0 +1,140 @@
...
+                exit 1
+            fi
+            VERSION="$2"
+            shift 2
+            ;;
+        *)
+            shift
+            ;;
+    esac
+done
```
luksa, line 88: This can be improved to support any order of positional arguments and flags. Like this:
```suggestion
# Parse positional arguments and flags 
while [[ $# -gt 0 ]]; do
    case "$1" in
        -n)
            if [[ -z "$2" || "$2" == -* ]]; then
                echo "Error: -n requires a non-empty argument."
                exit 1
            fi
            NAMESPACE="$2"
            shift 2
            ;;
        -v)
            if [[ -z "$2" || "$2" == -* ]]; then
                echo "Error: -v requires a non-empty argument."
                exit 1
            fi
            VERSION="$2"
            shift 2
            ;;
        *)
            if [[ -z "$INPUT" ]]; then
                INPUT="$1"  # First positional argument is the INPUT
            elif [[ -z "$OUTPUT" ]]; then
                OUTPUT="$1"  # Second positional argument is the OUTPUT
            fi
            shift
            ;;
    esac
done

# Ensure the input file is provided and is valid
if [[ -z "$INPUT" || ! -f "$INPUT" ]]; then
    echo "Error: Input file is missing or invalid."
    usage
fi

# If OUTPUT is not specified, generate a default output file name
if [[ -z "$OUTPUT" ]]; then
    OUTPUT="$(dirname "$INPUT")/$(basename "$INPUT" .yaml)-sail.yaml"
elif [[ -d "$OUTPUT" ]]; then
    echo "Error: OUTPUT must be a file, not a directory."
    exit 1
fi
```

ctartici: Done

luksa: Did you also add the tests?

ctartici: The complex one already had different order of the simple one. I now moved order like:
-v v1.24.3 istioFile -n istio-system sailFile

tools/configuration-converter.sh
```diff
@@ -0,0 +1,136 @@
...
+    echo "  -n <namespace>  : Namespace (optional, defaults to 'istio-system')"
+    echo "  -v <version>    : Istio Version (optional)"
+    exit 1
+}
+
+# Ensure at least one file argument is provided
+if [[ $# -lt 1 ]]; then
+    usage
+fi
+
```
luksa, line 36: This is not needed. The check on line 72 already takes care of this.

luksa (approved):
Looks great! There's just one more thing that could be removed, so if you can remove it before merging, that would be great.

dgn (approved):

PR #618, opened 2025-02-05
Capture ztunnel details on e2e failures

sridhargaddam:
Currently, when any ambient e2e test fails, we are not capturing details of the ztunnel and the associated test pods. This PR updates the code to include this info.

### DIFF
--- a/tests/e2e/util/common/e2e_utils.go
+++ b/tests/e2e/util/common/e2e_utils.go
@@ -43,6 +43,22 @@ import (
 	"istio.io/istio/pkg/ptr"
 )
 
+type testSuite string
+
+const (
+	Ambient           testSuite = "ambient"
+	ControlPlane      testSuite = "control-plane"
+	DualStack         testSuite = "dual-stack"
+	MultiCluster      testSuite = "multi-cluster"
+	Operator          testSuite = "operator"
+	MultiControlPlane testSuite = "multi-control-plane"
+)
+
+const (
+	SleepNamespace   = "sleep"
+	HttpbinNamespace = "httpbin"
+)
+
 var (
 	OperatorImage     = env.Get("IMAGE", "quay.io/sail-dev/sail-operator:latest")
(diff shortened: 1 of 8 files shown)


tests/e2e/util/common/e2e_utils.go
```diff
@@ -43,6 +43,21 @@ import (
...
 
+type testSuite string
+
+const (
+	Ambient      testSuite = "ambient"
+	ControlPlane testSuite = "control-plane"
+	DualStack    testSuite = "dual-stack"
+	MultiCluster testSuite = "multi-cluster"
+	Operator     testSuite = "operator"
+)
```
fjglira, line 54: I like the change in general, but I don't see a real value in this, only when it is set to Ambient is being used

fjglira: I mean, we can leave it like this, but I think using `common.ControlPlane` does not say too much to the user reading the test and trying to understand what is going on. Mostly because `common.SleepNamespace` is a namespace and `common.Ambient` is a suite name, but later, we only validate that the test suite is equal only to Ambient

fjglira (commented):
lgtm in general but I left some minor comments

tests/e2e/util/common/e2e_utils.go
sridhargaddam: The reason I added these constants which are passed as arguments to the API (LogDebugInfo - Line 150) is to allow the API to capture any testSuite specific logging (one such example is Ambient where we are capturing ztunnel). If there are any similar use-cases for other testSuites, we can easily update the LogDebugInfo API with additional logging. 

common.SleepNamespace/HttpbinNamespace are just constants used between LogDebugInfo() API and ambientSuite.

Let me know what you think and if you have any suggestions.

fjglira: Works for me

fjglira (approved):

PR #619, opened 2025-02-06
Set Istiod env COMPLIANCE_POLICY when running on a FIPS enabled OpenShift cluster

yxun:
- [X] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [ ] Test
- [ ] Documentation Update

This PR detects FIPS enabled mode on an OpenShift cluster and then set the Istiod env `COMPLIANCE_POLICY=fips-140-2`.
This configuration is documented in upstream release Note:
https://github.com/istio/istio/blob/master/releasenotes/notes/fips.yaml

Istiod needs to have set 'COMPLIANCE_POLICY=fips-140-2' env to run on a FIPS cluster. Otherwise, the sidecar proxy containers are not able to start with cipher suites initialization errors.


yxun:
I will add more unit tests for the missing codecov parts tomorrow.

pkg/istiovalues/profiles.go
```diff
@@ -52,6 +52,21 @@ func ApplyUserValues(mergedValues, userValues helm.Values,
 	return values, nil
 }
 
+// ApplyFipsValues: Detect FIPS enabled mode on OpenShift and add pilot.env.COMPLIANCE_POLICY
+func ApplyFipsValues(mergedHelmValues helm.Values, checkFilePath string) (helm.Values, error) {
```
luksa, line 56: This should probably go into `fips.go` instead of `profiles.go`, since it has nothing to do with profiles.

Also, why does the filepath need to be passed in? It's a well-known path, so it could be defined in a constant in `fips.go`.

yxun: @luksa I updated my PR about the two suggestions above. Thanks, Marko.

pkg/istiovalues/fips.go
```diff
@@ -0,0 +1,46 @@
...
+	"fmt"
+	"os"
+
+	"github.com/istio-ecosystem/sail-operator/pkg/helm"
+)
+
+const (
+	FipsEnableFilePath = "/proc/sys/crypto/fips_enabled"
+	TestModeFilePath   = "/tmp/fips_enabled"
+)
```
luksa, line 27: By changing this into a `var`, tests can override it.
```suggestion
var FipsEnableFilePath = "/proc/sys/crypto/fips_enabled"
```

pkg/istiovalues/fips.go
```diff
@@ -0,0 +1,46 @@
...
+// To avoid changing the host machine FIPS_ENABLED_FILE_PATH configuration,
+// set testMode to true when running unit tests.
+func ApplyFipsValues(mergedHelmValues helm.Values, testMode bool) (helm.Values, error) {
+	fipsEnabledFile := FipsEnableFilePath
+	if testMode {
+		fipsEnabledFile = TestModeFilePath
+	}
+
+	contents, err := os.ReadFile(fipsEnabledFile)
+	if err == nil && string(contents) == "1\n" {
```
luksa, line 39: Perhaps it's safer to trim the contents and compare them to `"1"` instead of relying on the `\n` always being there.

pkg/istiovalues/fips.go
```diff
@@ -0,0 +1,46 @@
...
+		fmt.Println("Running on FIPS enabled system.")
+		if err = mergedHelmValues.SetIfAbsent("pilot.env.COMPLIANCE_POLICY", "fips-140-2"); err != nil {
```
luksa, line 41: Does the `"fips-140-2"` need to be configurable (e.g. via an environment variable for the operator)?

yxun: I don't think that need to be exposed at the operator layer. According to the upstream note[1]. It's an environment variable for the istiod container. So our operator pod doesn't need that env.

I haven't heard a use case to set this value or other policy value using sail operator. If we need to add a field from our CRD, we can add it as a feature in later release. What do you think ?

[1] https://github.com/openshift-service-mesh/istio/blob/master/releasenotes/notes/fips.yaml

pkg/istiovalues/filps_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/istio-ecosystem/sail-operator/pkg/helm"
+)
+
+func TestApplyFipsValues(t *testing.T) {
+	tests := []struct {
+		name         string
+		expectValues helm.Values
+		expectErr    bool
```
dgn, line 30: I'd prefer a `fipsEnabled bool` in the testCase struct that controls the writing of the tmp file instead of just enabling fips after the first test because I think we shouldn't be making assumptions about the order of test executions

pkg/istiovalues/filps_test.go
dgn, line 1: typo: `fips_test.go`

yxun: Thanks for catching that. I have updated it.

pkg/revision/values.go
```diff
@@ -41,6 +41,14 @@ func ComputeValues(
 		return nil, fmt.Errorf("failed to apply profile: %w", err)
 	}
 
+	// apply FipsValues on top of mergedHelmValues from profile
+	if platform == config.PlatformOpenShift {
+		mergedHelmValues, err = istiovalues.ApplyFipsValues(mergedHelmValues)
+		if err != nil {
+			return nil, fmt.Errorf("failed to apply FIPS values: %w", err)
+		}
+	}
```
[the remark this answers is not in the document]
yxun: No, the line 36 userValues is only for setting "apply image digests".
And this ApplyFipsValues is calling "values.SetIfAbsent" method. If users set this env variable explicitly, it will not get overwritten.

pkg/istiovalues/fips.go
```diff
@@ -0,0 +1,40 @@
...
+var FipsEnableFilePath = "/proc/sys/crypto/fips_enabled"
+
+// ApplyFipsValues: Detect FIPS enabled mode on OpenShift and set value pilot.env.COMPLIANCE_POLICY
+// To avoid changing the host machine FIPS_ENABLED_FILE_PATH configuration,
+// set testMode to true when running unit tests.
+func ApplyFipsValues(values helm.Values) (helm.Values, error) {
+	readline, err := os.ReadFile(FipsEnableFilePath)
+	contents := strings.TrimSuffix(string(readline), "\n")
+	if err == nil && contents == "1" {
+		fmt.Println("Running on FIPS enabled system.")
```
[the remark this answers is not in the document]
yxun: yes, I have updated that and added a logger there. I have tested this on my cluster and the logging info is in the Sail operator pod log.

pkg/istiovalues/fips.go
```diff
@@ -0,0 +1,46 @@
...
+	readline, err := os.ReadFile(FipsEnableFilePath)
+	contents := strings.TrimSuffix(string(readline), "\n")
+	if err == nil && contents == "1" {
+		log.Info("Running on FIPS enabled system. Set COMPLIANCE_POLICY: fips-140-2.")
+		if err = values.SetIfAbsent("pilot.env.COMPLIANCE_POLICY", "fips-140-2"); err != nil {
+			log.Error(err, "Failed to set pilot.env.COMPLIANCE_POLICY")
+			return nil, fmt.Errorf("failed to set pilot.env.COMPLIANCE_POLICY: %w", err)
+		}
+	} else {
+		log.Info("Not running on FIPS enabled system.")
```
luksa, line 43: Do we need to log this, especially at the Info level?

pkg/istiovalues/fips.go
```diff
@@ -0,0 +1,46 @@
...
+
+// ApplyFipsValues: Detect FIPS enabled mode on OpenShift and set value pilot.env.COMPLIANCE_POLICY
+// To avoid changing the host machine FIPS_ENABLED_FILE_PATH configuration,
+// set testMode to true when running unit tests.
+func ApplyFipsValues(ctx context.Context, values helm.Values) (helm.Values, error) {
+	log := logf.FromContext(ctx)
+	readline, err := os.ReadFile(FipsEnableFilePath)
+	contents := strings.TrimSuffix(string(readline), "\n")
+	if err == nil && contents == "1" {
+		log.Info("Running on FIPS enabled system. Set COMPLIANCE_POLICY: fips-140-2.")
```
luksa, line 37: This too.

pkg/istiovalues/fips.go
yxun: hm, I was using that "Not running" in my cluster testing. I can remove this line logging.

pkg/istiovalues/fips.go
yxun: Well, that's what @nrfox was suggesting to use in the comment above.
I do think we need a way for showing users the detection is finding the FIPS enabled content.

yxun: I see. I can remove that logging.

yxun: I have implemented the optimization @nrfox mentioned above. And I have tested this latest change in a FIPS cluster. It works fine.

cmd/main.go
```diff
@@ -163,6 +168,16 @@ func main() {
...
+	// Detect FIPS enabled mode on OpenShift
+	if reconcilerCfg.Platform == config.PlatformOpenShift {
+		contents, err := os.ReadFile(FipsEnableFilePath)
+		fipsEnabled := strings.TrimSuffix(string(contents), "\n")
+		if err == nil && fipsEnabled == "1" {
+			setupLog.Info("FIPS mode is enabled in the system.")
+			istiovalues.FipsEnabled = true
+		}
+	}
+
```
luksa, line 180: It would be great if you moved this to a separate package (e.g. `fips`), since this is too low-level code for the main function.

yxun: I have updated this and refactored the code. The step above is in a function of istiovalues/fips.go file.

cmd/main.go
```diff
@@ -163,6 +164,14 @@ func main() {
 		reconcilerCfg.DefaultProfile = "default"
 	}
 
+	// Detect FIPS mode
+	if reconcilerCfg.Platform == config.PlatformOpenShift {
+		fipsEnabled, err := istiovalues.DetectFipsMode()
+		if err == nil && fipsEnabled {
+			setupLog.Info("FIPS mode is enabled in the system.")
+		}
```
luksa, line 172: The DetectFipsMode function could log this by itself.

luksa: Actually, I would refactor all this and have `fips.go` initialize the `FipsEnabled` automatically when it's first accessed (or, if required, explicitly call an init function here, but with no additional code). This would then allow you to have the file path as an parameter in the DetectFipsMode function (the init function would call this with the correct path). This way, you could easily test both the DetectFipsMode and the ApplyFipsValues functions and, at the same time, properly encapsulate the functionality within fips.go.

yxun: I have updated this change and put the call of detectFipsMode function into the fips.go init().
There were two other trivial changes for fixing lint error and logging. 

I removed the ReadFile error checking to avoid lint error when calling but not catching that error in init() call.
I removed the "FIPS mode is enabled" log line. It's a trivial log. And it needs passing a context for the logger there Seems like we don't want to change the ComputeValues method signature for this trivial logger.

dgn (approved):
LGTM. out of curiosity: could we also add this detection into istiod later? or is the filename RHEL-specific?

yxun:
> LGTM. out of curiosity: ... or is the filename RHEL-specific?

This file path was only tested from RHEL/RHCOS. More checking commands are listed in the JIRA OSSM-8693 comment.
Marko helped me pick the easiest one for checking here.

> could we also add this detection into istiod later?

Well, no. Initially my PR was putting this detection into the istiod / Istio CR creation later. However, that would cause each reconciliation reads the same file. That is not optimal from performance side. See Nick's comment above. 
So the current PR is detecting this file content once when this package is loaded.

PR #625, opened 2025-02-07
Add documentation for version alias

yannuil:
#### What type of PR is this?

- [ ] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [ ] Test
- [x] Documentation Update

Document how the version alias works. 

Related Issue/PR #566

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -89,7 +89,7 @@ spec:
 
 Istio uses a ConfigMap for its global configuration, called the MeshConfig. All of its settings are available through `spec.meshConfig`.
 
-To support canary updates of the control plane, Sail Operator includes support for multiple Istio versions. You can select a version by setting the `version` field in the `spec` to the version you would like to install, prefixed with a `v`. You can then update to a new version just by changing this field.
+To support canary updates of the control plane, Sail Operator includes support for multiple Istio versions. You can select a version by setting the `version` field in the `spec` to the version you would like to install, prefixed with a `v`. You can then update to a new version just by changing this field. An alias can be used for the latest z/patch versions of each supported y/minor versions. As per the example above, `v1.23-latest` can be specified in the `version` field. By doing so, the operator will keep the istio version with the latest z version of the same y version. 
 
 Sail Operator supports two different update strategies for your control planes: `InPlace` and `RevisionBased`. When using `InPlace`, the operator will immediately replace your existing control plane resources with the ones for the new version, whereas `RevisionBased` uses Istio's canary update mechanism by creating a second control plane to which you can migrate your workloads to complete the update.
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -89,7 +89,7 @@ spec:
 
 Istio uses a ConfigMap for its global configuration, called the MeshConfig. All of its settings are available through `spec.meshConfig`.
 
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -89,7 +89,7 @@ spec:
 
 Istio uses a ConfigMap for its global configuration, called the MeshConfig. All of its settings are available through `spec.meshConfig`.
 
-To support canary updates of the control plane, Sail Operator includes support for multiple Istio versions. You can select a version by setting the `version` field in the `spec` to the version you would like to install, prefixed with a `v`. You can then update to a new version just by changing this field.
+To support canary updates of the control plane, Sail Operator includes support for multiple Istio versions. You can select a version by setting the `version` field in the `spec` to the version you would like to install, prefixed with a `v`. You can then update to a new version just by changing this field. An alias can be used for the latest z/patch versions of each supported y/minor versions. As per the example above, `v1.23-latest` can be specified in the `version` field. By doing so, the operator will keep the istio version with the latest z version of the same y version.
```
fjglira, line 92: Just a minor comment; I think it will be better to change this:
`the operator will keep the istio version with the latest z version of the same y version`
To this
the operator will keep the istio version with the latest `z` version of the same `y` version
This is just to highlight that `z` and `y` represent specific values.

yannuil: Done

docs/README.md
```diff
@@ -89,7 +89,7 @@ spec:
 
 Istio uses a ConfigMap for its global configuration, called the MeshConfig. All of its settings are available through `spec.meshConfig`.
 
-To support canary updates of the control plane, Sail Operator includes support for multiple Istio versions. You can select a version by setting the `version` field in the `spec` to the version you would like to install, prefixed with a `v`. You can then update to a new version just by changing this field.
+To support canary updates of the control plane, Sail Operator includes support for multiple Istio versions. You can select a version by setting the `version` field in the `spec` to the version you would like to install, prefixed with a `v`. You can then update to a new version just by changing this field. An `vX.Y-latest` alias can be used for the latest z/patch versions of each supported y/minor versions. As per the example above, `v1.23-latest` can be specified in the `version` field. By doing so, the operator will keep the istio version with the latest `z` version of the same `y` version.
```
FilipB, line 92: Is this possible only for minor versions? So there is no `latest` alias? How will this affect istio revision name?

fjglira (approved):
lgtm

docs/README.md
yannuil: Do you mean `v1.23.0`? The revision name is not affected. The `version` value in the revision should get processed the same way.

sridhargaddam:
Please note that while the documentation has been merged, the actual implementation PR is still under review. Ideally, we should add a 'Depends on' and merge the documentation only after the implementation is merged to avoid any confusion for users reading the docs.

docs/README.md
sridhargaddam: > Is this possible only for minor versions? So there is no latest alias?

If i understand correctly, we will no longer support `latest` but will be supporting the following.
`master`
`v1.24-latest`
`v1.23-latest` 
...


And IstioRevision would point to the actual version.

FilipB: This should be also documented here. Users will have the same questions.

longmuir: Given that this is pretty important, I actually wonder if there should be a "Version" section in the document (perhaps between Gateways and Update strategy) that describes how this field works in a bit more detail, with more concrete examples spelled out. For example, if I set the version field to "v1.23-latest", what exactly happens if I am on Istio 1.23.2 and 1.23.3 becomes available? I think

PR #636, opened 2025-02-12
Allow setting vendor-specific defaults

dgn:
This adds a `vendor_defaults.yaml` which can be overwritten in a fork in order to define vendor-specific default values. I also added a section to README.md to at least mention these vendor-specific modifications.

### DIFF
--- a/pkg/revision/values.go
+++ b/pkg/revision/values.go
@@ -25,6 +25,7 @@ import (
 
 // ComputeValues computes the Istio Helm values for an IstioRevision as follows:
 // - applies image digests from the operator configuration
+// - applies vendor-specific default values
 // - applies the user-provided values on top of the default values from the default and user-selected profiles
 // - applies overrides that are not configurable by the user
 func ComputeValues(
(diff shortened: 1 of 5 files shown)


pkg/revision/values.go
```diff
@@ -35,6 +36,12 @@ func ComputeValues(
 	// apply image digests from configuration, if not already set by user
 	userValues = istiovalues.ApplyDigests(version, userValues, config.Config)
 
+	// apply vendor-specific default values
+	userValues, err := istiovalues.ApplyVendorDefaults(version, userValues)
```
sridhargaddam, line 40: If required, in a future PR, we can support this feature for IstioCNI/Ztunnel.

sridhargaddam (approved):
That was quick, thanks @dgn. 

I've added hold, just in case @luksa want to take a look. Feel free to remove the hold if needed.

pkg/revision/values.go
dgn: I was wondering about that. The `ComputeValues` func should probably be used across all resources, right? Currently, the IstioCNI and ZTunnel controllers do this separately

Issue #457, opened 2024-10-30
[FEATURE] Disable pruning of stale revisions for istios that manage external clusters


nrfox:
### Is this the right place to submit this?

- [X] This is not a question about how to use the sail-operator

### Feature Description

If a control plane manages an external cluster but it isn't in use on the control plane cluster, the Sail Operator will consider the control plane revision stale and remove it. This is a problem if there are workloads on the remote cluster still connected to the control plane.

The Sail Operator should not automatically prune revisions that manage external clusters.

nrfox:
@luksa we previously spoke offline about the operator checking the status of the remote revision before pruning and doing this by using the remote secret that istiod itself uses. An additional idea I had on that same concept is: rather than having the operator in the primary cluster manually reconcile all the resources in the remote cluster to determine if its in use, what if the operator in the remote cluster writes the `Istio` status somewhere the primary operator can read and the primary operator reads that to determine the status of the remote istio.

e.g.

Remote operator --> reconciles status and writes status to a `sail.operator.io` annotation on the mutating webhook.
Primary operator --> detects remote secret --> uses remote secret to read the status of remote istiorevision from the mutating webhook. If there's a remote revision that is in use then the primary revision counts as in use as well.

If there's a remote secret but there's no webhook (where the remote status gets written to) then the primary operator can assume there's no remote revisions. 

Trying to avoid having a single operator need to reconcile/watch all workload resources in remote clusters. I don't think that will scale very well.

PR #654, opened 2025-02-14
Disables pruning for istios that potentially manage external revs

nrfox:
#### What type of PR is this?

- [x] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [ ] Test
- [ ] Documentation Update

Disables pruning of istios that manage external clusters because the operator has no way of knowing if it is in use or not on the remote cluster(s).

Fixes #

Related Issue/PR #
Relates to https://github.com/istio-ecosystem/sail-operator/issues/457

### DIFF
--- a/controllers/istio/istio_controller_test.go
+++ b/controllers/istio/istio_controller_test.go
@@ -872,6 +872,47 @@ func TestGetPruningGracePeriod(t *testing.T) {
 	}
 }
 
+func TestManagesExternalRevision(t *testing.T) {
+	tests := []struct {
+		name     string
+		spec     v1.IstioSpec
+		expected bool
+	}{
+		{
+			name:     "Empty spec.values does not manage",
+			spec:     v1.IstioSpec{},
+			expected: false,
+		},
+		{
+			name:     "Empty spec.values.pilot does not manage",
+			spec:     v1.IstioSpec{Values: &v1.Values{}},
+			expected: false,
+		},
+		{
+			name:     "Empty spec.values.pilot.env does not manage",
+			spec:     v1.IstioSpec{Values: &v1.Values{Pilot: &v1.PilotConfig{}}},
+			expected: false,
+		},
+		{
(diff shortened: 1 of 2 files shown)


controllers/istio/istio_controller_test.go
```diff
@@ -872,6 +872,47 @@ func TestGetPruningGracePeriod(t *testing.T) {
...
 }
 
+func TestManagesExternalRevision(t *testing.T) {
+	tests := []struct {
+		name     string
+		spec     v1.IstioSpec
+		expected bool
+	}{
+		{
+			name:     "Empty spec.values does not manage",
```
luksa, line 882: What if `spec.values` is empty, but `spec.profile` points to a profile that sets the `EXTERNAL_ISTIOD` value. I don't think such a profile exist right now, but should we handle this to protect from such profiles in the future? But I don't even know how we should do this exactly, since the profiles are applied within the helm rendering process.

nrfox: > What if spec.values is empty, but spec.profile points to a profile that sets the EXTERNAL_ISTIOD value.

If the profile sets the `EXTERNAL_ISTIOD` value, isn't it already applied to the values at this point by this: https://github.com/istio-ecosystem/sail-operator/blob/84c35a10c7bcd11d9ac95179cb4258385b9e0017/pkg/revision/values.go#L45-L49

luksa: No. This was how we initially applied the profiles. Now, the profiles are implemented within the helm charts, so the controller doesn't even know what the final values are. But we kept this ApplyProfilesAndPlatform function, because a few fields are still applied the old way.

nrfox: Well then if there was a profile that set `EXTERNAL_ISTIOD: true` explicitly then this patch wouldn't help. Long term the sail operator should be able to determine if the external cluster is in use.

luksa (approved):

PR #656, opened 2025-02-14
e2e revisiontag

### LINKED ISSUE #549: e2e test for IstioRevisionTag
We need an e2e test that covers the functionality of IstioRevisionTag.

fjglira:
#### What type of PR is this?

- [ ] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [X] Test
- [ ] Documentation Update

Add an e2e test related to control plane updates and the use of IstioRevisionTag resource. Additional changes:

- Move Operator installation to BeforeSuite
- Move Operator deletion and cleanUp to AfterSuite

Fixes #549

Related Issue/PR #


tests/e2e/controlplane/control_plane_test.go
```diff
@@ -128,8 +113,8 @@ metadata:
 spec:
   version: %s
   namespace: %s`
-						yaml = fmt.Sprintf(yaml, name, istioCniNamespace)
-						Log("IstioCNI YAML:", indent(2, yaml))
+						yaml = fmt.Sprintf(yaml, version.Name, istioCniNamespace)
```
FilipB, line 116: Is this expected change?

tests/e2e/controlplane/control_plane_update_test.go
```diff
@@ -0,0 +1,329 @@
...
+	SetDefaultEventuallyTimeout(180 * time.Second)
+	SetDefaultEventuallyPollingInterval(time.Second)
+	debugInfoLogged := false
+
+	Describe("using IstioRevisionTag", func() {
+		if len(istioversion.List) < 2 {
+			Skip("Skipping update tests because there are not enough versions in versions.yaml")
+		}
+
+		baseVersion := istioversion.Old
```
FilipB, line 48: How does this work when there is more versions? e.g. 5, which will be base and which new?

tests/e2e/controlplane/control_plane_update_test.go
```diff
@@ -0,0 +1,329 @@
...
+					istioYAML = fmt.Sprintf(istioYAML, baseVersion, controlPlaneNamespace)
+					Log("Istio YAML:", indent(istioYAML))
+					Expect(k.CreateFromString(istioYAML)).
+						To(Succeed(), "Istio CR failed to be created")
+					Success("Istio CR created")
+				})
+
+				It("deploys istiod and pod is Ready", func(ctx SpecContext) {
+					Eventually(common.GetObject).WithArguments(ctx, cl, kube.Key("default"), &v1.Istio{}).
+						Should(HaveCondition(v1.IstioCNIConditionReady, metav1.ConditionTrue), "Istiod is not Available; unexpected Condition")
```
FilipB, line 95: Why using `IstioCNIConditionReady` for istiod?

tests/e2e/controlplane/control_plane_update_test.go
```diff
@@ -0,0 +1,329 @@
...
+						Should(HaveField("Status.IstioRevision", ContainSubstring(revisionName)),
+							"IstioRevisionTag version does not match the IstioRevision name of the base version")
+					Success("IstioRevisionTag version matches the Istio version")
+				})
+			})
+
+			When("sample pod is deployed", func() {
+				BeforeAll(func(ctx SpecContext) {
+					Expect(k.CreateNamespace(sampleNamespace)).To(Succeed(), "Sample namespace failed to be created")
+					Expect(k.Patch("namespace", sampleNamespace, "merge", `{"metadata":{"labels":{"istio-injection":"enabled"}}}`)).
```
FilipB, line 135: Why not using `oc label` instead?

tests/e2e/controlplane/control_plane_update_test.go
```diff
@@ -0,0 +1,329 @@
...
+							if pod.Status.Phase != corev1.PodRunning {
+								return false
+							}
+						}
+						return true
+					}).Should(BeTrue(), "At least one of the istiod pods is not running")
+					Success("Istiod pods are Running")
+				})
+
+				It("there are one IstionRevision for each version", func(ctx SpecContext) {
```
FilipB, line 199:
```suggestion
				It("there is one IstionRevision for each version", func(ctx SpecContext) {
```

tests/e2e/controlplane/control_plane_update_test.go
```diff
@@ -0,0 +1,329 @@
...
+					Success("Two IstionRevision found")
+				})
+
+				It("both IstionRevision are in use", func(ctx SpecContext) {
+					istioRevisions := &v1.IstioRevisionList{}
+					Expect(cl.List(ctx, istioRevisions)).To(Succeed())
+					for _, revision := range istioRevisions.Items {
+						Expect(revision).To(HaveCondition(v1.IstioRevisionTagConditionInUse, metav1.ConditionTrue), "IstioRevisionTag is not in use")
+					}
+					Success("Both IstionRevision are in use")
```
FilipB, line 218: We are just checking `IstioRevisionTagConditionInUse`  why do we say `Both IstionRevision are in use`

tests/e2e/controlplane/control_plane_update_test.go
```diff
@@ -0,0 +1,329 @@
...
+							if err != nil || !sidecarVersion.Equal(istioversion.Map[newVersion].Version) {
+								return false
+							}
+						}
+						return true
+					}).Should(BeTrue(), "Sidecar Istio version does not match the expected version")
+					Success("Istio sidecar version matches the expected new Istio version")
+				})
+
+				It("IstionRevision resource and istiod pod are deleted", func(ctx SpecContext) {
```
FilipB, line 276: Could you please add comment why it's expected it's deleted? It's because the old revision is not referenced by anything?

FilipB (changes requested):
Added a few comments

tests/e2e/controlplane/control_plane_test.go
fjglira: Yes, the lint complains because the argument was always 2. Until we do not need another value, we can skip adding the argument

tests/e2e/controlplane/control_plane_update_test.go
fjglira: We need at least two versions in `istioversion.List` to not skip the test. Checking the code:
```
	if len(list) > 0 {
		newVersion = list[0].Name
		defaultVersion = newVersion
		if len(list) > 1 {
			oldVersion = list[1].Name
		}
	}
```
New will always be the latest version, and Old will always be the latest minus one. The `versions.yaml` file is sorted from the newest version to the oldest version

tests/e2e/controlplane/control_plane_update_test.go
fjglira: My mistake, fixed

tests/e2e/controlplane/control_plane_update_test.go
```diff
@@ -0,0 +1,329 @@
...
+  targetRef:
+    kind: Istio
+    name: default`
+					Log("IstioRevisionTag YAML:", indent(IstioRevisionTagYAML))
+					Expect(k.CreateFromString(IstioRevisionTagYAML)).
+						To(Succeed(), "IstioRevisionTag CR failed to be created")
+					Success("IstioRevisionTag CR created")
+				})
+
+				It("creates the resource with condition InUse false", func(ctx SpecContext) {
```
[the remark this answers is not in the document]
fjglira: Added

tests/e2e/controlplane/control_plane_update_test.go
fjglira: CHanged and opened an issue to standardize the way namespaces and other objects are labeled across all the e2e test

tests/e2e/controlplane/control_plane_update_test.go
fjglira: Changed

tests/e2e/controlplane/control_plane_update_test.go
fjglira: Because at that point of the test, we are going to have 2 IstioRevisions, and both are going to be with the condition InUse to true because one is being used by the proxies and the new one is being referenced by the tag. I add a comment to explain Why

tests/e2e/controlplane/control_plane_update_test.go
fjglira: Added

tests/e2e/controlplane/control_plane_update_test.go
FilipB: It would be nice to add a comment about old == latest patch version -1, so when we have e.g. 1.24.2, 1.24.1 and 1.23.4 it will never pick 1.23.4 right for old?

FilipB (changes requested):
Just a minor comment

tests/e2e/controlplane/control_plane_update_test.go
fjglira: Currently, old and new will always be first in the version list and second. So, it can happen that the new is a different minor version from the old. To avoid this, I opened a new [issue](https://github.com/istio-ecosystem/sail-operator/issues/681) to add the possibility to select a patch for the same version (This is going to be needed anyway for update testing)

fjglira: I added a comment

FilipB (approved):

PR #667, opened 2025-02-19
Ensure e2e tests use the correct versions yaml file

luksa:
We use `-ldflags` to set the `istioversion.versionsFilename` variable, but this only works when building the binary. It doesn't work when running ginkgo-based e2e tests. When the `versionsFilename` is not set via ldflags, it will now use the environment variable `VERSIONS_YAML_FILE`. If this env var isn't set, we default to `versions.yaml`.

### DIFF
--- a/pkg/istioversion/version.go
+++ b/pkg/istioversion/version.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 
 	"github.com/Masterminds/semver/v3"
+	"github.com/istio-ecosystem/sail-operator/pkg/env"
 	"gopkg.in/yaml.v3"
 
 	"istio.io/istio/pkg/log"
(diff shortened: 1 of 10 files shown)


pkg/istioversion/version.go
```diff
@@ -137,3 +139,11 @@ func mustParseVersionsYaml(yamlBytes []byte) (
 
 	return list, defaultVersion, oldVersion, newVersion, versionMap, aliasList
 }
+
+func getEnvVar(key string, defaultValue string) string {
```
fjglira, line 143: There is already an `env.Get` defined in the e2e util package: https://github.com/istio-ecosystem/sail-operator/blob/72c12640e313737464b342e965bcfcfcda3aef8d/tests/e2e/util/env/env.go#L26-L27

Maybe worth moving it to a common place and used it both in the `version.go` and in the e2e test. wdyt?

fjglira (commented):
Just a minor comment

pkg/istioversion/version.go
luksa: Yes, you're absolutely right. 

I've moved it to `pkg/env`, but I had to remove the ginkgo logging. I hated that thing anyway, because it got printed even when the tests weren't running in verbose mode. If we need those lines, we should move print them somewhere else.

fjglira (approved):

sridhargaddam (approved):

PR #686, opened 2025-02-28
Improve istioversion and e2e test

### LINKED ISSUE #681: Improve istioversion package by adding a way to get latest patch version for each minor version
Currently, the `istioversion` package parses the versions.yaml file and get information about the versions from there. To be able to properly filter the latest patch version for each minor version we need to improve the func and give the user the possibility to select for example latest patch for each minor version

fjglira:
#### What type of PR is this?

- [ ] Bug Fix
- [ ] Refactor
- [X] Optimization
- [X] Test
- [ ] Documentation Update

This PR adds a new function to get all the patch versions from each Major.Minor version in the versions.yaml to be used later by the e2e test to iterate over only the latest patch version on each Minor. 

- GetLatestPatchVersions returns the last patch version for each Minor
- GetPatchVersionsByMajorMinor returns all the patch versions for every Minor version. This is needed to ensure that the update tests are over the same Major.Minor version

This will decrease the execution time of the e2e framework because before we where executing a lot of test for all the versions inside the versions.yaml file, now, we are only going to run the scenario for the latest patch for each Minor version 

Fixes #681

Related Issue/PR #

### DIFF
--- a/pkg/istioversion/version.go
+++ b/pkg/istioversion/version.go
@@ -139,3 +139,27 @@ func mustParseVersionsYaml(yamlBytes []byte) (
 
 	return list, defaultVersion, oldVersion, newVersion, versionMap, aliasList
 }
+
+// GetLatestPatchVersions returns the latest patch versions for all the Major.Minor versions
+func GetLatestPatchVersions() map[string]VersionInfo {
+	latestPatchVersions := make(map[string]VersionInfo)
+	for _, version := range List {
+		majorMinorVersion := fmt.Sprintf("%d.%d", version.Version.Major(), version.Version.Minor())
+		latestPatchVersion, ok := latestPatchVersions[majorMinorVersion]
+		if !ok || version.Version.GreaterThan(latestPatchVersion.Version) {
+			latestPatchVersions[majorMinorVersion] = version
+		}
+	}
+	return latestPatchVersions
+}
+
+// GetPatchVersionsByMajorMinor returns all the patch versions for each Major.Minor version in the list
+// The resulting map will have the Major.Minor version as the key and a list of patch versions as the value
+func GetPatchVersionsByMajorMinor() map[string][]VersionInfo {
+	allPatchVersions := make(map[string][]VersionInfo)
+	for _, version := range List {
+		majorMinorVersion := fmt.Sprintf("%d.%d", version.Version.Major(), version.Version.Minor())
+		allPatchVersions[majorMinorVersion] = append(allPatchVersions[majorMinorVersion], version)
+	}
+	return allPatchVersions
+}
--- a/tests/e2e/ambient/ambient_test.go
+++ b/tests/e2e/ambient/ambient_test.go
@@ -61,7 +61,7 @@ var _ = Describe("Ambient configuration ", Ordered, func() {
 	})
 
 	Describe("for supported versions", func() {
-		for _, version := range istioversion.List {
+		for _, version := range istioversion.GetLatestPatchVersions() {
 			// The minimum supported version is 1.24 (and above)
 			if version.Version.LessThan(semver.MustParse("1.24.0")) {
 				continue
--- a/tests/e2e/controlplane/control_plane_test.go
+++ b/tests/e2e/controlplane/control_plane_test.go
@@ -96,7 +96,7 @@ metadata:
 	})
 
 	Describe("given Istio version", func() {
-		for name, version := range istioversion.Map {
+		for name, version := range istioversion.GetLatestPatchVersions() {
 			Context(name, func() {
 				BeforeAll(func() {
 					Expect(k.CreateNamespace(controlPlaneNamespace)).To(Succeed(), "Istio namespace failed to be created")
--- a/tests/e2e/controlplane/control_plane_update_test.go
+++ b/tests/e2e/controlplane/control_plane_update_test.go
@@ -45,11 +45,11 @@ var _ = Describe("Control Plane updates", Label("update"), Ordered, func() {
 			Skip("Skipping update tests because there are not enough versions in versions.yaml")
 		}
 
-		// istioversion.Old is the version second version in versions.yaml file and istioversion.New is the first version in the List
-		// istioversion.Old is going to be the base version from where we are going to update to istioversion.New
-		// TODO: improve this: https://github.com/istio-ecosystem/sail-operator/issues/681
-		baseVersion := istioversion.Old
-		newVersion := istioversion.New
+		baseVersion, newVersion := getBaseAndNewVersion()
+		if baseVersion == "" || newVersion == "" {
+			Skip("Skipping update tests because there are not enough versions in versions.yaml")
+		}
+
 		Context(baseVersion, func() {
 			BeforeAll(func(ctx SpecContext) {
 				Expect(k.CreateNamespace(controlPlaneNamespace)).To(Succeed(), "Istio namespace failed to be created")
@@ -334,3 +334,41 @@ spec:
 		})
 	})
 })
+
+// getBaseAndNewVersion returns the base and new Istio versions to be used in the tests.
+// It will take the latest Major.Minor version and get the two newest patch of that Major.Minor version.
+func getBaseAndNewVersion() (string, string) {
+	type versionPair struct {
+		base, newVersion string
+		newSemver        *semver.Version
+	}
+	var latestPair versionPair
+
+	patchVersions := istioversion.GetPatchVersionsByMajorMinor()
+	for _, group := range patchVersions {
+		// Only consider groups with at least two patch versions.
+		if len(group) < 2 {
+			continue
+		}
+
+		newVer, err1 := semver.NewVersion(group[0].Name)
+		if err1 != nil {
+			continue
+		}
+		if _, err2 := semver.NewVersion(group[1].Name); err2 != nil {
+			continue
+		}
+
+		// Update latestPair if it's the first valid one or if this group is newer.
+		if latestPair.newSemver == nil || newVer.GreaterThan(latestPair.newSemver) {
+			latestPair.newSemver = newVer
+			latestPair.newVersion = group[0].Name
+			latestPair.base = group[1].Name
+		}
+	}
+
+	if latestPair.newSemver == nil {
+		return "", ""
+	}
+	return latestPair.base, latestPair.newVersion
+}
--- a/tests/e2e/dualstack/dualstack_test.go
+++ b/tests/e2e/dualstack/dualstack_test.go
@@ -64,7 +64,7 @@ var _ = Describe("DualStack configuration ", Ordered, func() {
 	})
 
 	Describe("for supported versions", func() {
-		for _, version := range istioversion.List {
+		for _, version := range istioversion.GetLatestPatchVersions() {
 			// The minimum supported version is 1.23 (and above)
 			if version.Version.LessThan(semver.MustParse("1.23.0")) {
 				continue
--- a/tests/e2e/multicluster/multicluster_multiprimary_test.go
+++ b/tests/e2e/multicluster/multicluster_multiprimary_test.go
@@ -68,7 +68,7 @@ var _ = Describe("Multicluster deployment models", Ordered, func() {
 
 	Describe("Multi-Primary Multi-Network configuration", func() {
 		// Test the Multi-Primary Multi-Network configuration for each supported Istio version
-		for _, version := range istioversion.List {
+		for _, version := range istioversion.GetLatestPatchVersions() {
 			Context(fmt.Sprintf("Istio version %s", version.Version), func() {
 				When("Istio resources are created in both clusters", func() {
 					BeforeAll(func(ctx SpecContext) {
--- a/tests/e2e/multicluster/multicluster_primaryremote_test.go
+++ b/tests/e2e/multicluster/multicluster_primaryremote_test.go
@@ -69,7 +69,7 @@ var _ = Describe("Multicluster deployment models", Ordered, func() {
 
 	Describe("Primary-Remote - Multi-Network configuration", func() {
 		// Test the Primary-Remote - Multi-Network configuration for each supported Istio version
-		for _, v := range istioversion.List {
+		for _, v := range istioversion.GetLatestPatchVersions() {
 			// The Primary-Remote - Multi-Network configuration is only supported in Istio 1.24+.
 			if version.Constraint("<1.24").Check(v.Version) {
 				Log(fmt.Sprintf("Skipping test, because Istio version %s does not support Primary-Remote Multi-Network configuration", v.Version))


tests/e2e/controlplane/control_plane_update_test.go
```diff
@@ -334,3 +334,41 @@ spec:
 		})
 	})
 })
+
+// getBaseAndNewVersion returns the base and new Istio versions to be used in the tests.
+// It will take the latest Major.Minor version and get the two newest patch of that Major.Minor version.
+func getBaseAndNewVersion() (string, string) {
```
luksa, line 340: Shouldn't this be in version.go?

tests/e2e/controlplane/control_plane_update_test.go
```diff
@@ -334,3 +334,41 @@ spec:
...
+	var latestPair versionPair
+
+	patchVersions := istioversion.GetPatchVersionsByMajorMinor()
+	for _, group := range patchVersions {
+		// Only consider groups with at least two patch versions.
+		if len(group) < 2 {
+			continue
+		}
+
+		newVer, err1 := semver.NewVersion(group[0].Name)
```
luksa, line 354: `.Name` is the version name, not the semantic version number. You should probably use `.Version` here.

tests/e2e/controlplane/control_plane_update_test.go
dgn, line 340: Why this new logic? Can't we just use .Old and .New like before?

dgn: and if we replace .Old and .New, we should remove them and put this func (or a similar one) in the istioversion package :slightly_smiling_face:

fjglira: Because .Old and.New can be different minor versions. Do we want to test also upgrades between versions? I mean, things can be different between minor versions

pkg/istioversion/version.go
```diff
@@ -139,3 +139,27 @@ func mustParseVersionsYaml(yamlBytes []byte) (
 
 	return list, defaultVersion, oldVersion, newVersion, versionMap, aliasList
 }
+
+// GetLatestPatchVersions returns the latest patch versions for all the Major.Minor versions
+func GetLatestPatchVersions() map[string]VersionInfo {
+	latestPatchVersions := make(map[string]VersionInfo)
+	for _, version := range List {
+		majorMinorVersion := fmt.Sprintf("%d.%d", version.Version.Major(), version.Version.Minor())
```
luksa, line 147: Why use `Sprintf`? Isn't the following simpler (and much easier to read)?
```suggestion
		majorMinorVersion := version.Version.Major() + "." + version.Version.Minor()
```

Also, introducing `ver := version.Version` would make the function less verbose.

fjglira:
#fixes 645. This issue was related to not running against all the versions in the versions.yaml file

pkg/istioversion/version_test.go
```diff
@@ -143,3 +138,97 @@ func TestParseVersionsYaml_InvalidYaml(t *testing.T) {
...
+		{Name: "1.23.2", Version: semver.MustParse("1.23.2")},
+		{Name: "1.22.0", Version: semver.MustParse("1.22.0")},
+	}
+
+	base, newVersion := GetBaseAndNewVersion()
+
+	// base and newVersion should be empty since there are no consecutive patch versions in the List for same Minor version
+	assert.Equal(t, "", base)
+	assert.Equal(t, "", newVersion)
+}
```
luksa, line 234: I think a single table-based test is much better. Less code and the entire test for the function is in a single test function. Functions like `TestGetBaseAndNewVersion` are perfect for table-based tests that list the input and the expected output. Such tests make it very easy and fast to see what the function is (supposed to be) doing.

pkg/istioversion/version.go
```diff
@@ -130,12 +124,71 @@ func mustParseVersionsYaml(yamlBytes []byte) (
...
+		if !ok || version.Version.GreaterThan(latestPatchVersion.Version) {
+			latestPatchVersions[majorMinorVersion] = version
 		}
 	}
+	return latestPatchVersions
+}
+
+// GetPatchVersionsByMajorMinor returns all the patch versions for each Major.Minor version in the list
+// The resulting map will have the Major.Minor version as the key and a list of patch versions as the value
+func getPatchVersionsByMajorMinor() map[string][]VersionInfo {
```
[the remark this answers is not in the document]
fjglira: I think not because the idea of this function is to group all the `VersionInfo` by major.minor release. With this, we will be able to use it later to run tests only over some specific major.minor version (For now, it is only being used to Get the base and new version). If we are going to return on a list of `VersionInfo` we already have a List for this, the other alternative is to return a `map[string][]Versions.` wdyt?

tests/e2e/ambient/ambient_test.go
```diff
@@ -61,7 +61,7 @@ var _ = Describe("Ambient configuration ", Ordered, func() {
 	})
 
 	Describe("for supported versions", func() {
-		for _, version := range istioversion.List {
+		for _, version := range istioversion.GetLatestPatchVersions() {
```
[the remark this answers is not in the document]
fjglira: I can arrange the list to be in order, but returning `[]VersionInfo` is the same as using `List`. The idea is to run the test only against the latest version for each major.minor version

pkg/istioversion/version_test.go
fjglira: Done

tests/e2e/ambient/ambient_test.go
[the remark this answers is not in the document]
fjglira: In fact, after checking `GetLatestPatchVersions()` return the List in the same order as they are in the versions.yaml file, so, it will execute the test in the same order as they are now: from newest version to oldest, and in last place the master branch

pkg/istioversion/version_test.go
```diff
@@ -143,3 +139,107 @@ func TestParseVersionsYaml_InvalidYaml(t *testing.T) {
...
+
+	assert.Equal(t, "1.24.1", base)
+	assert.Equal(t, "1.24.2", newVersion)
+
+	List = []VersionInfo{}
+
+	base, newVersion = GetBaseAndNewVersion()
+
+	assert.Equal(t, "", base)
+	assert.Equal(t, "", newVersion)
```
luksa, line 232: Since these are three different subtests, you can use `t.Run()` to make them proper subtests. This allows you to explain what each subtest does.

```suggestion
	t.Run("valid versions", func(t *testing.T) {
		List = []VersionInfo{
			{Name: "1.24.2", Version: semver.MustParse("1.24.2")},
			{Name: "1.24.1", Version: semver.MustParse("1.24.1")},
			{Name: "1.23", Version: semver.MustParse("1.23")},
			{Name: "1.23.2", Version: semver.MustParse("1.23.2")},
			{Name: "1.23.1", Version: semver.MustParse("1.23.1")},
			{Name: "1.22.0", Version: semver.MustParse("1.22.0")},
		}

		base, newVersion := GetBaseAndNewVersion()

		assert.Equal(t, "1.24.1", base)
		assert.Equal(t, "1.24.2", newVersion)
	})
	
	t.Run("empty list", func(t *testing.T) {
		List = []VersionInfo{}

		base, newVersion := GetBaseAndNewVersion()

		assert.Equal(t, "", base)
		assert.Equal(t, "", newVersion)
	})
```

Same for the last sub-test.

But, instead of having three completely independent subtests, you could also convert them into a table-based test. Not sure which of the two is better, since the test code is very simple. I'm fine with either approach, but I think most people would implement this as a table-based test.

pkg/istioversion/version_test.go
```diff
@@ -143,3 +139,107 @@ func TestParseVersionsYaml_InvalidYaml(t *testing.T) {
 		mustParseVersionsYaml(yamlBytes)
 	})
 }
+
+func TestGetLatestPatchVersions_Valid(t *testing.T) {
```
luksa, line 143: These `TestGetLatestPatchVersions_*` tests could also be a single table-based test.

fjglira: Done

pkg/istioversion/version_test.go
```diff
@@ -143,3 +139,132 @@ func TestParseVersionsYaml_InvalidYaml(t *testing.T) {
...
+
+func TestGetBaseAndNewVersion(t *testing.T) {
+	t.Run("valid versions", func(t *testing.T) {
+		List = []VersionInfo{
+			{Name: "1.24.2", Version: semver.MustParse("1.24.2")},
+			{Name: "1.24.1", Version: semver.MustParse("1.24.1")},
+			{Name: "1.23", Version: semver.MustParse("1.23")},
+			{Name: "1.23.2", Version: semver.MustParse("1.23.2")},
+			{Name: "1.23.1", Version: semver.MustParse("1.23.1")},
+			{Name: "1.22.0", Version: semver.MustParse("1.22.0")},
```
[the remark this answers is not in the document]
luksa: And the version list should also contain at least one example, where the name is not a version (e.g. `master` and `1.24-latest`).

pkg/istioversion/version.go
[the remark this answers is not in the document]
luksa: Are you sure this function is actually going to be needed elsewhere? Currently the two functions (`GetBaseAndNewVersion`  and this one) seems overly complex for what they actually need to do.

pkg/istioversion/version_test.go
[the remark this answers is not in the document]
fjglira: Done

pkg/istioversion/version_test.go
```diff
@@ -143,3 +139,132 @@ func TestParseVersionsYaml_InvalidYaml(t *testing.T) {
...
+		expected := map[string][]VersionInfo{
+			"1.24": {
+				{Name: "1.24.2", Version: semver.MustParse("1.24.2")},
+			},
+		}
+
+		if diff := cmp.Diff(expected, groups); diff != "" {
+			t.Errorf("unexpected result; diff (-expected, +actual):\n%v", diff)
+		}
+	})
```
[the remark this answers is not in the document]
fjglira: I'm simplifying the func, so, this test is not needed. I'll add a new one that covers everything

pkg/istioversion/version.go
```diff
@@ -130,12 +125,83 @@ func mustParseVersionsYaml(yamlBytes []byte) (
...
+	var latestPair versionPair
+
+	patchVersions := getPatchVersionsByMajorMinor()
+	for _, group := range patchVersions {
+		// Only consider groups with at least two patch versions.
+		if len(group) < 2 {
+			continue
+		}
+
+		newVer, err1 := semver.NewVersion(group[0].Name)
```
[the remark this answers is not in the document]
fjglira: Fixed, I changed the entire `GetBaseAndNewVersion` func to first sort the entire `List` by `Version` and, from there, validate the two newer versions. Adding also to the test validation where the name can be any string to validate that it is working properly

pkg/istioversion/version.go
```diff
@@ -130,12 +125,83 @@ func mustParseVersionsYaml(yamlBytes []byte) (
...
+	for _, version := range List {
+		majorMinorVersion := fmt.Sprintf("%d.%d", version.Version.Major(), version.Version.Minor())
+		allPatchVersions[majorMinorVersion] = append(allPatchVersions[majorMinorVersion], version)
+	}
+	return allPatchVersions
+}
+
+// GetBaseAndNewVersion returns old and new Istio patch versions.
+// It will take the latest Major.Minor version and get the two newest patch of that Major.Minor version.
+func GetBaseAndNewVersion() (string, string) {
```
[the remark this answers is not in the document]
fjglira: Fixed, now is working as expected. Is returning the latest two patch versions as base and new

tests/integration/api/istio_test.go
```diff
@@ -215,9 +215,10 @@ var _ = Describe("Istio resource", Ordered, func() {
 	Describe("update", func() {
 		var workloadNs *corev1.Namespace
 		rev := &v1.IstioRevision{}
+		baseVersion, newVersion := istioversion.GetBaseAndNewVersion()
 
 		BeforeAll(func() {
-			if istioversion.Old == "" {
+			if len(istioversion.List) < 2 {
```
luksa, line 221: In theory this can be false, while baseVersion and newVersion are not set.
```suggestion
			if baseVersion == "" || newVersion == "" {
```

pkg/istioversion/version_test.go
```diff
@@ -143,3 +139,98 @@ func TestParseVersionsYaml_InvalidYaml(t *testing.T) {
...
+	t.Run("empty list", func(t *testing.T) {
+		List = []VersionInfo{}
+
+		versions := GetLatestPatchVersions()
+
+		assert.Len(t, versions, 0)
+	})
+}
+
+func TestGetBaseAndNewVersion(t *testing.T) {
```
luksa, line 178: This would be much simpler if it were table-based.

pkg/istioversion/version.go
```diff
@@ -130,12 +125,70 @@ func mustParseVersionsYaml(yamlBytes []byte) (
...
+	for _, v := range List {
+		key := v.Version.String()
+		if seen[key] {
+			continue
 		}
+		if newVersion == "" {
+			newVersion = v.Name
+		} else if base == "" {
+			base = v.Name
+			break
```
luksa, line 185: Instead of this break, you could simply have a `return base, newVersion` here and you wouldn't need the `if base == ""` statement at the end of the function.

pkg/istioversion/version.go
```diff
@@ -62,10 +63,6 @@ var (
 	Map map[string]VersionInfo
 	// Default is the default version
 	Default string
-	// Old is the previous supported version
-	Old string
-	// New is the latest supported version
-	New string
```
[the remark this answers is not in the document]
fjglira: Ok, that makes sense. I think I deleted it at the beginning of the refactor when I had another approach before. I'll add these vars again and renamed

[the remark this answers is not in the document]
fjglira: Ok, I reverted some changes. After checking all your comments `GetBaseAndNewVersion()` does not make sense to have it, because is adding complexity. The previous way to set old and new was really simple and the main idea of creating `GetBaseAndNewVersion` was to get base and new from the latest minor version returning only two versions from the same minor version. But, after seeing all the complexity that I'm adding makes more sense to keep the old way to set old and new because doing the upgrade test works fine. If you have some time tomorrow we can check this together

pkg/istioversion/version.go
fjglira: Not needed anymore

pkg/istioversion/version.go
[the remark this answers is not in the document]
fjglira: `GetLatestPatchVersions` still make sense to have it because it will simplify iteration in test by running the test only over the latest patch of each minor version

tests/e2e/controlplane/control_plane_test.go
```diff
@@ -203,7 +203,7 @@ spec:
 
 				When("sample pod is deployed", func() {
 					BeforeAll(func() {
-						if isAlias(name, version) {
+						if isAlias(version.Name, version) {
```
luksa, line 206: As far as I can tell, `version` can never be an alias, because it comes from `istioversion.GetLastestPatchVersions()`, which gets the versions from `istioversions.List`, which only contains actual versions (no aliases).

Also, this function call seems very wrong. It passes in `version.name` and `version`, and the `isAlias` function then returns `name != version.Name`, so effectively it's returning `version.Name != version.Name`, which is always false. 

I suggest we remove `isAlias`.

tests/e2e/multicontrolplane/multi_control_plane_suite_test.go
```diff
@@ -30,9 +30,10 @@ import (
 )
 
 var (
-	cl                     client.Client
-	err                    error
-	version                = istioversion.New
+	cl  client.Client
+	err error
+	// version will be the first version in the list, this version is the newest Istio version in the versions.yaml file
+	version                = istioversion.List[0].Name
```
luksa, line 36: Why this change? Isn't this the same as `istioversion.New` that was here before?

luksa (approved):

PR #689, opened 2025-03-03
Cleanup created clusterrolebinding after tests

asmigala:
#### What type of PR is this?

- [x] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [x] Test
- [ ] Documentation Update

Re-running the e2e tests with `SKIP_DEPLOY` on the same cluster would fail because a clusterrolebinding was not cleaned up


dgn (approved):
Thanks! Good catch

PR #697, opened 2025-03-06
Add a note that version is optional in Istio CRs

sridhargaddam:
This PR removes the version field from some examples and adds a note to clarify that it is optional. It does not modify YAMLs used in upgrade examples or other cases where version details are included in sample output.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -77,7 +77,6 @@ kind: Istio
 metadata:
   name: default
 spec:
-  version: v1.23.2
   namespace: istio-system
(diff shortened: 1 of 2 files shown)


fjglira (approved):
We should check and do this to all the examples in the docs

sridhargaddam:
> We should check and do this to all the examples in the docs

My intention was to add a note in the main examples and I avoided upgrade examples. 
If you think we can apply this change in other places too, please feel free to push a PR.

Issue #698, opened 2025-03-07
[BUG] changing namespace of IstioCNI resource fails


dgn:
### Is this the right place to submit this?

- [x] This is not a question about how to use the sail-operator

### Bug Description

Changing the `spec.namespace` of an existing IstioCNI resource fails with the following error (from operator logs):
```
2025-03-07T08:21:52Z    ERROR   ctrlr.istiocni  Reconciler error        {"IstioCNI": "default", "reconcileID": "4202192d-7070-42ce-8afd-ba99e5ae5500", "error": "failed to install/update Helm chart \"cni\": failed to install helm chart cni: Unable to continue with install: ClusterRole \"istio-cni\" in namespace \"\" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key \"meta.helm.sh/release-namespace\" must equal \"cni\": current value is \"istio-cni\""}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler
        sigs.k8s.io/controller-runtime@v0.19.3/pkg/internal/controller/controller.go:316
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem
        sigs.k8s.io/controller-runtime@v0.19.3/pkg/internal/controller/controller.go:263
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func2.2
        sigs.k8s.io/controller-runtime@v0.19.3/pkg/internal/controller/controller.go:224
```

The `Istio` resource's `namespace` field is immutable. Probably this one should be immutable, too - at least until we support changing the field.

### Operator Version

1.0.0-rc

PR #699, opened 2025-03-07
make spec.namespace field of `IstioCNI` immutable

dgn:
We don't support migrating `IstioCNI` to another namespace, so we should make the field immutable for now. We already did it for the spec.namespace field of the `Istio` resource. Once we implement support for namespace migrations of these resources, we can revert this.

Fixes https://github.com/istio-ecosystem/sail-operator/issues/698


luksa (approved):

FilipB:
@dgn cherry pick to release-1.0?

PR #705, opened 2025-03-10
Add `make cluster` command

### LINKED ISSUE #554: [FEATURE] Add `make clusters` to allow creating KinD clusters for local development

### Feature Description

There's already support for KinD clusters in e2e tests, but these clusters get destroyed once the tests are done.
It would be good to have an ability to stand up one or more KinD clusters to use for local development & testing.
This could also be useful for PoCs and demos.

### Additional Information

_No response_

mkolesnik:
#### What type of PR is this?

- [ ] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [ ] Test
- [ ] Documentation Update

Add the ability to create a cluster for testing by using `make cluster`. Multiple clusters can be created by utilizing the `MULTICLUSTER=true` option.
This allows developers to easily create a cluster to run manual testing on.

Most of the code has been carved out of `tests/e2e/integ-suite-kind.sh` which now relies on the cluster existing (the Makefile has been updated to create it when running the tests).

Fixes #554 

Related Issue/PR #

### DIFF
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -334,6 +334,13 @@ create-gh-release: helm-package ## Create a GitHub release and upload the helm c
 		--generate-notes \
 		$(GH_RELEASE_ADDITIONAL_FLAGS)
 
+.PHONY: cluster
+cluster: SKIP_CLEANUP=true
+cluster: ## Creates a KinD cluster(s) to use in local deployments.
+	source ${SOURCE_DIR}/tests/e2e/setup/setup-kind.sh; \
+	export HUB="$${KIND_REGISTRY}"; \
+	OCP=false ${SOURCE_DIR}/tests/e2e/setup/build-and-push-operator.sh
+
 .PHONY: deploy
 deploy: verify-kubeconfig helm ## Deploy controller to an existing cluster.
 	$(info NAMESPACE: $(NAMESPACE))
--- a/README.md
+++ b/README.md
@@ -85,6 +85,23 @@ spec:
 You’ll need a Kubernetes cluster to run against. You can use [KIND](https://sigs.k8s.io/kind) to get a local cluster for testing, or run against a remote cluster.
 **Note:** Your controller will automatically use the current context in your kubeconfig file (i.e. whatever cluster `kubectl cluster-info` shows).
 
+### Quick start using a local KIND cluster
+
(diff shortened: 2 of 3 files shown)


scripts/cluster.sh
dgn, line 1: nit: let's not create yet another `hack`/`tools` directory :)

mkolesnik: Thanks, where do you propose we put this?

I looked at `common/scripts` but it seems it's all auto updated from istion/common-files.
Perhaps in `tools` then?

fjglira: I think you can place it inside the `test/e2e` folder, at the end is going to be used by the `integ-suite-kind.sh` script mainly by the e2e, and now will be used also by the make target

fjglira:
Hi @mkolesnik, I added in another PR some changes to reuse the setup kind in another test step. I think you can rebase the PR and set the new make target to use https://github.com/istio-ecosystem/sail-operator/blob/d2db329e9688b99681919056632fcfc0fee17eec/tests/documentation_tests/scripts/run-docs-examples.sh#L81

Makefile.core.mk
```diff
@@ -334,6 +334,11 @@ create-gh-release: helm-package ## Create a GitHub release and upload the helm c
 		--generate-notes \
 		$(GH_RELEASE_ADDITIONAL_FLAGS)
 
+.PHONY: cluster
+cluster: SKIP_CLEANUP=true
+cluster: ## Creates a KinD cluster(s) to use in local deployments.
+	${SOURCE_DIR}/tests/e2e/setup/setup-kind.sh
+
```
fjglira, line 341: This sets up the kind cluster and the local registry, which is great. However, I think it's missing a convenient step to build and push the image to the local registry, similar to what we already do in the e2e tests.

We already have make targets for building and pushing images, but having this as part of the same target would make testing changes easier and more streamlined.

To do this, you'd just need to set the following env vars:
```
export IMAGE_BASE="${IMAGE_BASE:-sail-operator}"
export TAG="${TAG:-latest}"
export HUB="${HUB:-localhost:5000}"
```
Then call the push target.

Without that, the added value over just running kind create cluster manually is limited, IMO.
What do you think, @dgn?

fjglira (approved):
lgtm

PR #711, opened 2025-03-13
Check IstioCNI status for IstioRevisions that depend on IstioCNI

luksa:
Previously, when an IstioRevision depended on IstioCNI, either because the user enabled the use of IstioCNI through `values.pilot.cni.enabled` or because the platform profile enables it, the IstioRevision resource would show as healthy even though workloads would fail because IstioCNI hasn't been deployed. This commit adds a new `DependenciesHealthy` condition to IstioRevision and makes `IstioRevision.status.state` and the `STATUS` column report the fact that IstioCNI is missing or unhealthy. This makes it easier for users to see that their setup is misconfigured.

### DIFF
--- a/api/v1/istio_types.go
+++ b/api/v1/istio_types.go
@@ -235,6 +235,23 @@ const (
 )
 
 const (
+	// IstioConditionDependenciesHealthy signifies whether the dependencies required by this Istio are healthy.
+	// For example, an Istio with spec.values.pilot.cni.enabled=true requires the IstioCNI resource to be deployed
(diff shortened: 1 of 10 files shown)


dgn:
should we cover this in the IstioCNI integration test? Just a quick check of the happy/sad paths maybe - we could easily simulate an unhealthy IstioCNI there by manipulating status

luksa:
@dgn integration test added.

dgn (approved):
Thanks Marko!

PR #727, opened 2025-03-21
Add ConfigMap with final Helm values to charts that don't have one

luksa:
Ever since the profiles were reimplemented within the Helm charts, there has been no way to see the actual values that were used to render the charts. This includes the values from the user-selected profile, the platform profile, and the compatibility version profile.

Previously, when the values from the profiles were applied before Helm was invoked, you could use `helm get values` to see the final set of values. Now, this is no longer possible, because the profile values are applied within the rendering process by the zzz_profile.yaml template and aren't stored anywhere.

This commit introduces a new ConfigMap with the name `values` for the default revision, and `values-<revision>` for all others. This ConfigMap will allow the operator (and users) to look up the actual values used to render the charts.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -69,6 +69,10 @@
 # User Documentation
 Sail Operator manages the lifecycle of your Istio control planes. Instead of creating a new configuration schema, Sail Operator APIs are built around Istio's helm chart APIs. All installation and configuration options that are exposed by Istio's helm charts are available through the Sail Operator CRDs' `values` fields.
 
+Similar to using Istio's Helm charts, the final set of values used to render the charts is determined by a combination of user-provided values, default chart values, and values from selected profiles. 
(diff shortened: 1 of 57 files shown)


luksa:
I'll submit a PR upstream to make this ConfigMap part of the upstream charts, but since it most likely won't be cherry-picked to all the old charts, we (also) need to add the ConfigMap by patching the (old) charts.

dgn:
I think generating a ConfigMap is really the only way to capture this information. I'm just not sure whether we should apply it to the cluster. I'm still leaning towards including it in the `status` - but maybe not by default, but only when the user activates it. Maybe a setting like `spec.debug.showAppliedValues`?

dgn (commented):
LGTM, can you add a paragraph to the docs about this? It's going to be helpful when debugging things

PR #738, opened 2025-03-27
Add unit tests for env.go

luksa:


pkg/env/env_test.go
```diff
@@ -0,0 +1,115 @@
...
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGet(t *testing.T) {
+	key := "ENV_TEST_TEST_GET"
+	value := "test value"
+	assert.NoError(t, os.Setenv(key, value))
```
nrfox, line 27: Very nit picky and the risk is pretty low but using `t.Setenv(...)` will automatically unset the env var after the test completes so that the env var doesn't affect other tests.

pkg/env/env_test.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+func TestGetBool(t *testing.T) {
+	boolKey := "ENV_TEST_TEST_GET_BOOL_VALID"
+	boolValue := "true"
+	assert.NoError(t, os.Setenv(boolKey, boolValue))
+
+	notBoolKey := "ENV_TEST_TEST_GET_BOOL_INVALID"
+	notBoolValue := "not a bool"
+	assert.NoError(t, os.Setenv(notBoolKey, notBoolValue))
```
nrfox, line 70: Same here.

luksa: Fixed.

nrfox (approved):

PR #744, opened 2025-04-01
add Istio 1.25.1

dgn:


api/v1/istio_types.go
```diff
@@ -37,10 +37,10 @@ const (
...
-	// +kubebuilder:validation:Enum=v1.24-latest;v1.24.4;v1.24.3;v1.24.2;v1.24.1;v1.24.0;v1.23-latest;v1.23.5;v1.23.4;v1.23.3;v1.23.2;v1.22-latest;v1.22.8;v1.22.7;v1.22.6;v1.22.5;v1.21.6;master;v1.25-alpha.c2ac935c
-	// +kubebuilder:default=v1.24.4
+	// Must be one of: v1.25-latest, v1.25.1, v1.24-latest, v1.24.4, v1.24.3, v1.24.2, v1.24.1, v1.24.0, v1.23-latest, v1.23.5, v1.23.4, v1.23.3, v1.23.2, v1.22-latest, v1.22.8, v1.22.7, v1.22.6, v1.22.5, v1.21.6, master, v1.25-alpha.c2ac935c.
```
nrfox, line 40: Should we also remove `v1.25-alpha.c2ac935c`?

FilipB:
I though new istio version should be added by automator? Or is it adding only patch versions and minor versions must be added manually?

dgn:
> I thought new istio version in main should be added by automator? Or is it adding only patch versions and minor versions must be added manually?

We only add new patch versions automatically, at least currently.

api/v1/istio_types.go
dgn: No, that's the latest commit on `master` that we support. It's not present in the release builds anyway

FilipB (approved):

PR #746, opened 2025-04-02
codecov: ignore dirs not holding code

dgn:


luksa (approved):

dgn:
closing this, the problematic behavior I observed in another PR was not caused by it (it was only taking the integration tests into account instead of unit+integration, causing it to undercount)

PR #750, opened 2025-04-03
POC: running docs examples using runme binary

### LINKED ISSUE #454: Automatic testing of documentation
We should make sure to have automated tests for our documentation, so that

* we can be sure the steps in our documentation are working as expected and
* we don't make changes to the code that accidentally cause a change in behavior

Istio is automatically testing the instructions in their documentation, maybe we can draw some inspiration from their solution.

fjglira:
#### What type of PR is this?

- [ ] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [ ] Test
- [ ] Documentation Update

POC to show in the contributor's meeting How we can run the docs example by using the `runme` binary. I added an example md file with sail operator installation and a control plane installation. If we decide this solution we will need to figured out some things like:

- Setup for the operator
- Setup and deletion of the kind cluster
- The guidelines can be similar to what I already defined in the other PR, we will just need to add the information specifically to `runme`

Fixes #454 

Related Issue/PR #


fjglira:
Hey @mkolesnik  from your [PR](https://github.com/istio-ecosystem/sail-operator/pull/705) I added some of the changes that you need here because they are needed it also to reuse the setup for the docs automation test. Take a look and we can check if work for both goals

tests/documentation_tests/scripts/update-docs-examples.sh
```diff
@@ -0,0 +1,103 @@
...
+
+# Create the output directory if it does not exist
+mkdir -p "$OUTPUT_DIR"
+
+# Create first a list with all the files in the docs/ directory and subdirectories to be copied
+# Exclude the files in the EXCLUDE_FILES array and include only the files that contains at least one pattern in their code blocks like this:
+# bash { name=
+FILES_TO_COPY=()
+
+for file in "$DOCS_DIR"/*; do
```
MaxBab, line 48: Just a suggestion.
Maybe we could avoid multiple `for` loops, by using the following approach:

```
FILES_TO_COPY=()

# Convert EXCLUDE_FILES to a regex pattern
EXCLUDE_PATTERN="^($(IFS='|'; echo "${EXCLUDE_FILES[*]}"))$"

# Find all .md files (top-level and in subdirs of $DOCS_DIR)
find "$DOCS_DIR" -type f -name "*.md" | while read -r file; do
  basename=$(basename "$file")

  # Skip excluded files
  if [[ "$basename" =~ $EXCLUDE_PATTERN ]]; then
    continue
  fi

  # Check for pattern match in file
  if grep -q "bash { name=" "$file"; then
    FILES_TO_COPY+=("$file")
  fi
done
```

tests/e2e/common-operator-integ-suite.sh
```diff
@@ -32,42 +35,19 @@ parse_flags() {
...
-        shift
-        # no point building if we don't deploy
-        SKIP_BUILD=true
-        SKIP_DEPLOY=true
-        ;;
-      --olm)
-        shift
-        OLM=true
-        ;;
+      --ocp) shift; OCP=true ;;
```
MaxBab, line 38: I'm just wondering.

Does this change and all changes below are required for this PR?
I'm asking it only because the PR is already big and this additional part overloads it more.

MaxBab (commented):
@fjglira 

I left comments for this PR.
But overall, it looks really great improvement for all the docs aspect of the repo.

docs/guidelines/guidelines.md
```diff
@@ -0,0 +1,159 @@
...
+  name: default
+spec:
+  targetRef:
+    kind: Istio
+    name: default
+EOF
+```
+````
+
+- Validation steps are always needed, to add a validation step you can comment directly the entire code block with the bash command that you want to run. This validation steps need to use the same tag as the example where they are going to run and should be prefixed with `validation-validation-description`. For example:
```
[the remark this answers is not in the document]
fjglira: Done

docs/guidelines/example-documentation-following-guidelines.md
```diff
@@ -0,0 +1,112 @@
...
+kubectl create namespace sail-operator
+helm install sail-operator sail-operator/sail-operator --version 1.0.0 -n sail-operator
+```
+
+- Wait for the operator to be ready:
+```bash { name=wait-operator tag=example}
+kubectl wait --for=condition=available --timeout=600s deployment/sail-operator -n sail-operator
+```
+
+<!-- ```bash { name=wait-operator tag=example}
```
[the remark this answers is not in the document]
fjglira: No, is an example and I propabily forgot to delete the duplicated validation

tests/documentation_tests/example-documentation-following-guidelines-runme.md
```diff
@@ -0,0 +1,112 @@
...
+```bash { name=check-istio tag=example}
+kubectl get pods -n istio-system
+kubectl get istio
+```
+
+- Deploy sample application:
+```bash { name=deploy-sample-app tag=example}
+kubectl create namespace sample
+kubectl label namespace sample istio-injection=enabled
+kubectl apply -n sample -f https://raw.githubusercontent.com/istio/istio/release-1.23/samples/bookinfo/platform/kube/bookinfo.yaml
```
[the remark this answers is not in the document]
fjglira: Changed

tests/e2e/common-operator-integ-suite.sh
fjglira: This specific change is not required for the PR but almost all the changes in this files are required because I''m moving out functions to simplify and set independend responsabilities for each file. I can revert the changes that are not needed for this PR but mostly the non needed are simplifications to be more simple in the script

CONTRIBUTING.md
```diff
@@ -37,3 +38,6 @@ The Sail Operator project has a weekly contributor call every Thursday at 4PM CE
 ## Security Issues
 
 If you find a security issue in the Sail Operator project, please refer to the [Security Policy](https://github.com/istio-ecosystem/sail-operator/security/policy) for more information on how to report security issues. Please do not report security issues in the public GitHub repository.
+
+## Writing Documentation
+If you want to add new documentation or examples to our existing documentation please take a look at the [documentation guidelines](https://github.com/istio-ecosystem/sail-operator/docs/guidelines/guidelines.md) to help you get started. Take into account that we run automation test over the documentation examples, so please ensure that the examples are correct and follows the guidelines to ensure it will be properly tested.
```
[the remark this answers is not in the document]
fjglira: Done

fjglira (commented):
Hi, I did the changes from the review and added also some comments

docs/guidelines/guidelines.md
```diff
@@ -0,0 +1,159 @@
...
+- Check all the md files in the `docs` folder and exclude from the list the files described inside the `EXCLUDE_FILES` variable. The resulting list will be all the .md files that contains at least one time this pattern `bash { name=`. This pattern is used by the `readme` tool to generate the resulting jupiter notebook by looking into all the code blocks that has certain fields decorator. The fields decorator and how to use them are explained in the next section.
+- Once the files are copied into the destination path we check those files to uncomment the bash commented steps. This bash commented code block are going to be hiden in the original md files but we will uncomment this code block to be able to run validation steps that wait for conditions an avoid flakiness of the test. THis means that all the validation steps need to be commented code blocks
+
+### Adding a topic example to the automation workflow
+To add a new topic to the automation workflow you need to:
+1. Add into the example or the new example in the docs the next patterns in exact order into each code block `bash` that will be running the example:
```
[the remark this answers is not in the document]
fjglira: Improved

tests/e2e/common-operator-integ-suite.sh
fjglira: I'll rollback some of those changes to try to keep the PR more easy to review

fjglira: Done

tests/e2e/common-operator-integ-suite.sh
```diff
@@ -243,35 +164,20 @@ if [ "${SKIP_BUILD}" == "false" ]; then
     if [ "${OLM_VERSION}" != "" ]; then
       OLM_INSTALL_ARGS+="--version ${OLM_VERSION}"
     fi
+
+    # Ensure kubeconfig is set to the kind cluster
+    kind export kubeconfig --name="${KIND_CLUSTER_NAME}"
```
MaxBab, line 169: If this flow is being used by the multicluster, could you apply similar approach as defined here?
https://github.com/istio-ecosystem/sail-operator/blob/main/tests/e2e/integ-suite-kind.sh#L75-L85

Otherwise, it creates an issue during MULTICLUSTER execution scenario, because when setting a kubeconfig var that way, it changes the ip address of one of the multicluster kind clusters to something random, like 127.0.0.1:47564, instead of expected docker ip range.

fjglira: Initially, this flow is only for when OLM is set to true, and we are not using this for multicluster tests. I'll take a look, but yesterday, I was having some issues with this approach

Makefile.core.mk
```diff
@@ -188,6 +188,42 @@ test.e2e.describe: ## Runs ginkgo outline -format indent over the e2e test to sh
...
+		curl -fsL $$URL -o /tmp/runme.tar.gz || { \
+			echo "Download failed! Please check the URL and RUNME_VERSION."; \
+			exit 1; \
+		}; \
+		tar -xzf /tmp/runme.tar.gz -C /tmp || { \
+			echo "Extraction failed!"; \
+			exit 1; \
+		}; \
+		mv /tmp/runme $(LOCALBIN)/runme; \
+		rm -f /tmp/runme.tar.gz; \
```
dgn, line 207:
```suggestion
		GOBIN=$(LOCALBIN) GO111MODULE=on go install github.com/runmedev/runme/v3@v$(RUNME_VERSION) > /dev/stderr; \
```

this way it's simpler and you build it from source instead of pulling artifacts from the company behind it

docs/guidelines/guidelines.md
```diff
@@ -0,0 +1,158 @@
...
+````
+
+- Validation steps are always needed, to add a validation step you can comment directly the entire code block with the bash command that you want to run. This validation steps need to use the same tag as the example where they are going to run and should be prefixed with `validation-description`. For example:
+```md
+<!-- ```bash { name=validation-wait-operator tag=example}
+kubectl wait --for=condition=available --timeout=600s deployment/sail-operator -n sail-operator
+``` -->
+```
+
+*Important:* please add always validation steps to avoid flakyness during the execution, this steps will ensure that resource conditions are met and the test will not fail because of a timeout or any other issue. This validation needs to add failures to the test if the condition is not met.
```
dgn, line 152:
```suggestion
*Important:* please always add validation steps to avoid flakyness during the execution, these steps will ensure that resource conditions are met and the test will not fail because of a timeout or any other issue. This validation needs to fail the test if the condition is not met, by returning a non-zero return code. Use `kubectl wait` over `sleep` wherever possible.
```

tests/documentation_tests/example-documentation-following-guidelines-runme.md
```diff
@@ -0,0 +1,104 @@
...
+if ! kubectl get pods -n sample -l app=productpage -o jsonpath='{range .items[*]}{@.metadata.name}{" "}{range .spec.containers[*]}{@.name}{" "}{end}{"\n"}{end}' | grep -q istio-proxy; then
+  echo "No Istio sidecar (istio-proxy) injected in productpage pod!"
+  exit 1
+fi
+```
+
+- Check the proxy version of the sample application:
+```bash { name=check-proxy-version tag=example}
+istioctl proxy-status 
+```
```
dgn, line 104: should we add an example for hidden code blocks with verification steps?

dgn (commented):
first pass review. This looks amazing @fjglira, thank you very much!

docs/guidelines/guidelines.md
```diff
@@ -0,0 +1,158 @@
...
+- Once the files are copied into the destination path we check those files to uncomment the bash commented steps. This bash commented code block are going to be hiden in the original md files but we will uncomment this code block to be able to run validation steps that wait for conditions an avoid flakiness of the test. THis means that all the validation steps need to be commented code blocks
+
+### Adding a topic example to the automation workflow
+To add a new topic to the automation workflow you need to:
+1. In your documentation topic, each bash code block that you want to execute as part of the automation must include the following pattern:
+- `bash { name=example-name tag=example-tag }`: the fields used here are:
+  - `name`: the name of the example step, this is usefull to identify the step in the output of the test. The name should be short and descriptive. For example: `deploy-operator`, `wait-operator`, etc.
+  - `tag`: the tag of the example that is going to be used to run the test, is important to be a unique name. This tag should be unique and should not be used in other examples. The tag can be used to run only a specific example inside a file.
```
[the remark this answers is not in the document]
fjglira: Good point, if someone redeclares the tag in the same file, it will be an issue, if those tags are in different files, there will not be issues because we execute later by setting file and tag. But, I'll be adding a validation steps to avoid duplicity of the tags

tests/documentation_tests/example-documentation-following-guidelines-runme.md
fjglira: This is the resulting file, that is why the validation step here is uncommented. In the doc inside /docs/guidelines there are some examples, check `name=check-sidecar-exist`

dgn: oops, sorry- ignore :)

docs/guidelines/guidelines.md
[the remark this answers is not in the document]
fjglira: I added a validation step before start copying the files where we check if the same tag is being used in another file. If is being used it will fail. I think adding a validation for inside the same file is too dificult because we will have the tag multiple times already

tests/e2e/common-operator-integ-suite.sh
fjglira: I added a validation step to avoid running this flow with MULTICLUSTER. In fact, this flow will fail anyway because it needs to install OLM in both clusters and currently the code does not handle multiple clusters

dgn (approved):

bartoszmajsak (commented):
A little bit late to the party, but PTAL @fjglira :)

docs/guidelines/guidelines.md
```diff
@@ -0,0 +1,158 @@
...
+
+### What update-docs-examples.sh do?
+The script `update-docs-examples.sh` is going to run the following steps:
+- Check all the md files in the `docs` folder and exclude from the list the files described inside the `EXCLUDE_FILES` variable. The resulting list will be all the .md files that contains at least one time this pattern `bash { name=`. This pattern is used by the `readme` tool to generate the resulting jupiter notebook by looking into all the code blocks that has certain fields decorator. The fields decorator and how to use them are explained in the next section.
+- Once the files are copied into the destination path we check those files to uncomment the bash commented steps. This bash commented code block are going to be hiden in the original md files but we will uncomment this code block to be able to run validation steps that wait for conditions an avoid flakiness of the test. THis means that all the validation steps need to be commented code blocks
```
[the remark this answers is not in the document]
fjglira: Already discussed on Slack

tests/e2e/common-operator-integ-suite.sh
```diff
@@ -118,120 +118,41 @@ initialize_variables() {
...
 
   echo "Setting Istio manifest file: ${ISTIO_MANIFEST}"
   ISTIO_NAME=$(yq eval '.metadata.name' "${WD}/../../$ISTIO_MANIFEST")
 
-  TIMEOUT="3m"
-
-  VERBOSE=${VERBOSE:-"false"}
-}
-
-get_internal_registry() {
```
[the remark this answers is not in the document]
fjglira: Well, they are and they aren't. I need a way to set up the environment to be able to run the test, so I added some extra changes to the PR to reuse the setup that we already used in the e2e framework. It was not clean, but I did it to avoid non-working changes until another change was merge

docs/guidelines/guidelines.md
```diff
@@ -0,0 +1,158 @@
...
+spec:
+  namespace: istio-system
+  updateStrategy:
+    type: RevisionBased
+    inactiveRevisionDeletionGracePeriodSeconds: 30
+  version: v1.24.2
+```
+````
+
+*Note:* take into account that all the code blocks that are going to be run by the `runme` tool should be tagged with the language `bash` at the beginning of the block and they need to contain only commands that are going to be run in the terminal and all the steps that are not needed to be executed because they are not commands or they are not needed for the goal of the example should be ignored with the `ignore=true` tag. For example
```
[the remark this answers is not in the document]
fjglira: I simplified the text in the new PR

PR #755, opened 2025-04-04
Add PROFILE column

luksa:
The purpose of this column is mostly to make it immediately clear when an Istio deployment is actually a remote or ambient installation. We previously had a TYPE column on IstioRevision, which indicated whether the revision was remote or not. The PROFILE column improves this, since it's also used for other current and future profiles.


nrfox (commented):
+1 for adding this. @dgn any opposition?

sridhargaddam:
So on OCP, will it print `openshift` profile?

luksa:
> So on OCP, will it print `openshift` profile?

No. This only prints the user's selected profile, not the platform or the compatibility version profile.

dgn (approved):

sridhargaddam (approved):

PR #766, opened 2025-04-10
Use VERSION in release TAG

nrfox:
#### What type of PR is this?

- [ ] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [ ] Test
- [ ] Documentation Update

Use `VERSION` in release branch for `TAG`. Removes `-latest` from tag on 1.25 release branch.

Fixes #

Related Issue/PR #

### DIFF
--- a/bundle/manifests/sailoperator.clusterserviceversion.yaml
+++ b/bundle/manifests/sailoperator.clusterserviceversion.yaml
@@ -33,8 +33,8 @@ metadata:
       ]
     capabilities: Seamless Upgrades
     categories: OpenShift Optional, Integration & Delivery, Networking, Security
-    containerImage: quay.io/sail-dev/sail-operator:1.25-latest
-    createdAt: "2025-04-09T13:53:38Z"
+    containerImage: quay.io/sail-dev/sail-operator:1.25.0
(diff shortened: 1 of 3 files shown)


bundle/manifests/sailoperator.clusterserviceversion.yaml
```diff
@@ -776,7 +776,7 @@ spec:
                 - --zap-log-level=info
                 command:
                 - /sail-operator
-                image: quay.io/sail-dev/sail-operator:1.25-latest
+                image: quay.io/sail-dev/sail-operator:1.25
```
FilipB, line 779: Don't we need 1.25.0 tag?

nrfox: Yes... Do we have minor version tags as well? i.e. both `1.25.0` and a floating `1.25` tag?

dgn (approved):

bundle/manifests/sailoperator.clusterserviceversion.yaml
nrfox: I updated to just be `VERSION` for now. If we want floating tags later for MINOR_VERSION that can be changed in `main`.

PR #768, opened 2025-04-10
Add olm replaces and skipRange annotations to csv

nrfox:
#### What type of PR is this?

- [ ] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [ ] Test
- [ ] Documentation Update

Bumps version to 1.25.1 and adds OLM `skipRange` and `replaces` annotations to the CSV.

Fixes #

Related Issue/PR #


bundle/manifests/sailoperator.clusterserviceversion.yaml
```diff
@@ -46,6 +46,8 @@ metadata:
     features.operators.openshift.io/token-auth-aws: "false"
     features.operators.openshift.io/token-auth-azure: "false"
     features.operators.openshift.io/token-auth-gcp: "false"
+    olm.replaces: sailoperator.v1.0.0
```
FilipB, line 49: Checking CSV api [doc](https://docs.redhat.com/en/documentation/openshift_container_platform/4.18/html/operatorhub_apis/clusterserviceversion-operators-coreos-com-v1alpha1#spec-2), `replaces` belongs under `spec`.

bundle/manifests/sailoperator.clusterserviceversion.yaml
```diff
@@ -46,6 +46,8 @@ metadata:
...
+    olm.skipRange: '>=1.0.0 <1.25.1'
```
FilipB, line 50: We should discuss if we want to use `skip` to only ignore one bad release (1.0.1) or `skipRange` which would be used in the future to skip all previous patch releases as described [here](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/how-to-update-operators.md#z-stream-support). If using `skipRange` I'm not sure if `replaces` should point to `1.0.0` or `1.0.1`. Reading [this](https://olm.operatorframework.io/docs/concepts/olm-architecture/operator-catalog/creating-an-update-graph/#skiprange), I would say we should use `1.0.1`.

FilipB (changes requested):

bundle/manifests/sailoperator.clusterserviceversion.yaml
dgn: but there is no `1.0.1` release? I think it should be `spec.replaces: sailoperator.v1.0.0`

nrfox: Discussed this offline and setting this to `1.25.0` for the `1.25.1` release.

FilipB: I was thinking `1.25.0` and writing `1.0.1` :)

FilipB (approved):

PR #772, opened 2025-04-14
Remove omitempty from nested struct fields or replace with omitzero

luksa:
This change removes the `omitempty` tag from nested struct fields, since it has no effect. 

Additionally, instead of removing `omitempty` tag from fields with the type `metav1.Time` (e.g. the field `lastTransitionTime`), we instead now use `omitzero`, which ensures that the field is omitted when not set.

### DIFF
--- a/bundle/manifests/sailoperator.io_istiocnis.yaml
+++ b/bundle/manifests/sailoperator.io_istiocnis.yaml
@@ -1519,6 +1519,9 @@ spec:
                 description: Reports the current state of the object.
                 type: string
             type: object
+        required:
+        - metadata
+        - spec
         type: object
         x-kubernetes-validations:
         - message: metadata.name must be 'default'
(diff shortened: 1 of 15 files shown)


bundle/manifests/sailoperator.io_istiocnis.yaml
```diff
@@ -1512,6 +1512,9 @@ spec:
                 description: Reports the current state of the object.
                 type: string
             type: object
+        required:
+        - metadata
+        - spec
```
dgn, line 1517: aren't these changes that would require a version bump?

luksa: I initially thought that both of these can actually be required and since all existing objects stored in the API should have these fields, we can make them required. However, I've now checked the core k8s kinds and they all mark these two as optional, so I've now done the same.

dgn (approved):

PR #775, opened 2025-04-15
Fix updating pre-release istio versions

FilipB:
This has been broken since introducing the version aliases.
Note that `tools/update_deps.sh` is still failing on `make gen` with:
```
/work/bin/controller-gen crd:allowDangerousTypes=true webhook paths="./..." output:crd:artifacts:config=chart/crds
github.com/istio-ecosystem/sail-operator/api/v1:-: use of unimported package "structpb"
github.com/istio-ecosystem/sail-operator/api/v1:-: use of unimported package "structpb"
/work/api/v1/values_types.gen.go:618:16: unknown type structpb.Struct
/work/api/v1/values_types.gen.go:956:18: unknown type structpb.Value
Error: not all generators ran successfully

```


dgn:
@FilipB I fixed the remaining errors in the update-deps script. They were caused by two new fields with undefined types

FilipB:
> @FilipB I fixed the remaining errors in the update-deps script. They were caused by two new fields with undefined types
Great! Thank you.

dgn (approved):

PR #776, opened 2025-04-15
Fix appVersion inconsistency in revisiontags chart

luksa:
The revisiontags chart shows the appVersion as v1.25.1, whereas all other charts show it as 1.25.1.
```
$ helm ls
NAME                       	NAMESPACE   	... APP VERSION
default-revisiontags       	istio-system	... v1.25.1
myistio-v1-25-latest-istiod	istio-system	... 1.25.1
```

This change removes the v prefix.


luksa:
Closing this, as this requires a different solution. The `v` prefix is not the only problem, actually. Unlike the other charts, the appVersion contains only the partial instead of the full commit hash.

PR #782, opened 2025-04-16
Remove the `install.operator.istio.io/owning-resource` label from all resources

luksa:
When charts are deployed via the Sail operator, this label is always set to `unknown`, so there's no point in including it in the resource labels. It's also specific to the upstream operator and is planned for removal (https://github.com/istio/istio/issues/52670).


sridhargaddam (approved):
> It's also specific to the upstream operator and is planned for removal (https://github.com/istio/istio/issues/52670).

Looks like the upstream issue was closed as it was stale.

PR #783, opened 2025-04-16
[wip] Add label app.kubernetes.io/managed-by: sail-operator

luksa:
There was previously no mention of the Sail operator in the resources it deploys. The `app.kubernetes.io/managed-by` label indicated that the resources were managed by Helm, but this is somewhat misleading. Even though resources are indeed rendered by Helm, this is only an implementation detail. The resources are in fact managed by the Sail operator, so the label should reflect this.

This label will also allow the operator to watch for changes only in resources that are managed by it and ignore all others.


dgn:
Hmm, it feels wrong to me to be doing this by manipulating the charts. I'd prefer to do it as part of the rendering. Maybe a postRender step would be better? That way, we don't have to match on any specifics in the charts (ie the if-block) but can just look for the label, and if it's present, replace "Helm" with "sail-operator" (or "Sail Operator"?)

luksa:
Yeah, that was actually how I first started implementing this. Then I realized it's much simpler to just modify the chart. The benefit is that the chart is then the source of truth. If we override the label in the post-renderer, people may find the label in the chart and miss the fact that it's being overridden.

Of course, setting the label in the renderer ensures that absolutely all resources have this label, even if a chart doesn't set it. I guess this is more important than not having a misleading label value in the charts. I'll rewrite the PR.

FilipB:
Is this good label to be used in the must gather [here](https://github.com/openshift-service-mesh/istio-must-gather/blob/main/gather_istio.sh#L173)?

luksa:
> Is this good label to be used in the must gather [here](https://github.com/openshift-service-mesh/istio-must-gather/blob/main/gather_istio.sh#L173)?

Yes. If we get the change into Helm, all the resources created by the Sail operator will have this label.

luksa:
Closing, as Helm maintainers have rejected the `app.kubernetes.io/managed-by` label value change. We need to use a different label at least until the operator starts using Helm to only render the charts, but applies the resources by itself (and can thus override the label value).

Replaced by https://github.com/istio-ecosystem/sail-operator/pull/801

PR #786, opened 2025-04-17
Check if IstioRevisionTag references a remote IstioRevision

luksa:
The istiorevisiontag controller doesn't render anything when the reference IstioRevision has the `istiodRemote.enabled` value set to `true`, leaving users confused why the MutatingWebhookConfiguration isn't there. Instead of quietly rendering nothing, the controller will now return a validation error, clearly signalling to the user what's wrong.


nrfox (commented):
LGTM but is there really no way to use stable rev tags in a multi-cluster environment?

nrfox (approved):

PR #798, opened 2025-04-23
Adding examples docs automation and creating prebuilt functions

### LINKED ISSUE #781: Tag examples over the docs to add the topics to the documentation testing workflow
Once the job is already set in the prow, we need to start adding the tag to all the documentation steps to increase the coverage over the docs. We can raise the topic in the community call to set the scope of the testing over all the docs

fjglira:
#### What type of PR is this?

- [ ] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [X] Test
- [X] Documentation Update

This PR is addressing:
* Adding to docs examples tags on two of the examples (more test are going to be included in a consecutive PR)
* Add a prebuilt function script to gather the functions that are going to be reused across all the docs

*Note*: This first PR adding a test to the automation aims to open a discussion about the automation in case something needs to be changed or improved over this PR. In a consecutive PR, more tests are going to be included

Related Issue/PR #781
More PR are going to be needed to close the issue because there are more test to automate

#### Additional information:
TODO: 
* Update guidelines with specific information regarding prebuilt functions and how to use them. I'll do it after this [PR](https://github.com/istio-ecosystem/sail-operator/pull/791) is merged
* Automate more tests, more information about which test I think should be automated in the [issue](https://github.com/istio-ecosystem/sail-operator/issues/781)
* I had to change the prebuilt function to use sh because I had some weird issues sourcing the files and running the script only from make target. need to check in-depth what is going on there

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -149,6 +149,7 @@ To update the CNI plugin, just change the `version` field to the version you wan
 
 1. Create the `IstioCNI` resource.
     ```bash { name=install-cni tag=cni-update}
+    kubectl create ns istio-cni
     cat <<EOF | kubectl apply -f-
     apiVersion: sailoperator.io/v1
     kind: IstioCNI
@@ -165,9 +166,8 @@ To update the CNI plugin, just change the `version` field to the version you wan
     EOF
     ```
 <!-- ```bash { name=validation-wait-cni tag=cni-update}
-kubectl create ns istio-cni
 . scripts/prebuilt-func.sh
-with_retries wait_cni_ready "istio-cni"
+wait_cni_ready "istio-cni"
 with_retries resource_version_equal "istiocni" "default" "v1.24.2"
 ``` -->
 2. Confirm the installation and version of the CNI plugin.
@@ -194,7 +194,7 @@ print_cni_info
 echo "Sleeping for 5 seconds to wait for the patch to be applied"
 sleep 5s
 . scripts/prebuilt-func.sh
-with_retries wait_cni_ready "istio-cni"
+wait_cni_ready "istio-cni"
 with_retries resource_version_equal "istiocni" "default" "v1.24.3"
 ``` -->
 4. Confirm the CNI plugin version was updated.
@@ -457,9 +457,9 @@ Steps:
       version: v1.22.5
     EOF
     ```
-<!-- ```bash { name=lidation-wait-istio-pods tag=inplace-update}
+<!-- ```bash { name=validation-wait-istio-pods tag=inplace-update}
     . scripts/prebuilt-func.sh
-    with_retries wait_istio_ready "istio-system"
+    wait_istio_ready "istio-system"
     print_istio_info
 ``` -->
 3. Confirm the installation and version of the control plane.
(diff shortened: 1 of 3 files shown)


docs/README.md
```diff
@@ -165,8 +164,13 @@ To update the CNI plugin, just change the `version` field to the version you wan
           - kube-system
     EOF
     ```
+<!-- ```bash { name=validation-wait-cni tag=cni-update}
+kubectl create ns istio-cni
```
FilipB, line 168: Shouldn't this be part of the actual doc step instead of validation step?

docs/README.md
```diff
@@ -165,8 +164,13 @@ To update the CNI plugin, just change the `version` field to the version you wan
...
+. scripts/prebuilt-func.sh
+with_retries wait_cni_ready "istio-cni"
```
FilipB, line 170: `wait_cni_ready` already contains retries do we need `with_retries` here?

docs/README.md
```diff
@@ -175,11 +179,24 @@ To update the CNI plugin, just change the `version` field to the version you wan
     NAME                   READY   STATUS    RESTARTS   AGE
     istio-cni-node-hd9zf   1/1     Running   0          90m
     ```
+
+<!-- ```bash { name=print-cni tag=cni-update}
+. scripts/prebuilt-func.sh
+print_cni_info
```
FilipB, line 185: This is not doing any validation so it's just like a logging function?

tests/documentation_tests/scripts/prebuilt-func.sh
```diff
@@ -0,0 +1,196 @@
+#!/bin/sh
+
```
FilipB, line 2: Maybe adding `set -euo pipefail`?

tests/documentation_tests/scripts/prebuilt-func.sh
FilipB, line 1: Maybe we should add a make target which is running shellcheck for this file?

docs/README.md
fjglira: Yes, it is only to print the information that is being skipped in the console block. This will help to debug things if something fails during the execution of the test. I can add the commands to an actual example and put the expected output as a console block, also, it will be clearer. wdyt?

docs/README.md
fjglira: Moved into the actual step

tests/documentation_tests/scripts/prebuilt-func.sh
fjglira: We already run make lint that includes:
```
lint-scripts:
	@${FINDFILES} -name '*.sh' -print0 | ${XARGS} -n 256 shellcheck
```

tests/documentation_tests/scripts/prebuilt-func.sh
fjglira: We are using `sh` I had a TODO to be able to use `bash` (More info in the description of the PR). So I added `set -euo` only because pipefail is a bash feature

docs/README.md
FilipB: I think it's fine as it is.

FilipB (approved):

dgn (approved):
LGTM. I don't really like adding `. scripts/prebuilt-func.sh` everywhere, but we can fix it in a follow-up - maybe by adding it into every `bash` code block when we do the copying.

PR #802, opened 2025-04-28
Implement istiobase controller

luksa:
The istiobase controller manages the application of the base chart. As only one instance of this chart can be installed at once, this has to be done by a new controller. If we had added the rendering of this chart to the istiorevision controller, we could have two instances of that controller both trying to update the chart/release.


luksa:
Needs #801

controllers/istiobase/istiobase_controller.go
```diff
@@ -0,0 +1,244 @@
...
+		return fmt.Errorf("failed to reconcile: %w", err)
+	}
+	log.Info("Installing Helm chart")
+	return r.installHelmChart(ctx, rev)
+}
+
+func (r *Reconciler) getDefaultRevision(ctx context.Context) (*v1.IstioRevision, error) {
+	// 1. get the IstioRevision referenced in the default IstioRevisionTag
+	tag := v1.IstioRevisionTag{}
+	if err := r.Client.Get(ctx, types.NamespacedName{Name: "default"}, &tag); err != nil {
```
nrfox, line 108: This can only ever be `default`? In the helm installation of istio, it seems you can set `defaultRevision` to be another revision name. From: https://istio.io/latest/docs/setup/install/helm/#installation-steps

> When performing a revisioned installation, the base chart requires the --set defaultRevision=<revision> value to be set for resource validation to function. Below we install the default revision, so --set defaultRevision=default is configured.

dgn: Yeah, technically you can, at least upstream - we need to deprecate that field in our API. If we're going to support it, individual `Istio` resources are probably not the place to configure it

luksa: Already deprecated: https://github.com/istio-ecosystem/sail-operator/pull/760

controllers/istiobase/istiobase_controller.go
```diff
@@ -0,0 +1,244 @@
...
+			}
+			return true
+		},
+	}
+}
+
+func clearIgnoredFields(obj client.Object) {
+	obj.SetResourceVersion("")
+	obj.SetGeneration(0)
+	obj.SetManagedFields(nil)
```
nrfox, line 238: Do any of the controllers use SSA? If not, we can probably set [this](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.20.4/pkg/cache#TransformStripManagedFields) on the manager/cache.

nrfox (commented):
LGTM, do we need any tests for this?

controllers/istiobase/istiobase_controller.go
luksa: This has to be done globally, when instantiating the manager, right? So, best to do it in a separate PR.

dgn:
> LGTM, do we need any tests for this?

I was about to write the same comment, some tests would be good

luksa:
Yeah, I'm just about done with the tests.

luksa:
Not sure what's up with the integration test coverage job, because the integration tests pass fine when you run them locally (and in CI). There is a strange error message related to the revisiontag controller:
```
 2025-04-30T14:15:40Z	ERROR	ctrlr.revtag	Reconciler error	{"IstioRevisionTag": "default", "reconcileID": "5ca78ebf-7352-4088-9039-a474e9bed946", "error": "failed to uninstall Helm chart \"revisiontags\": uninstallation completed with 1 error(s): uninstall: Failed to purge the release: release: not found"}
```

controllers/istiorevision/istiorevision_controller.go
```diff
@@ -165,26 +165,39 @@ func (r *Reconciler) installHelmCharts(ctx context.Context, rev *v1.IstioRevisio
...
 	if err != nil {
 		return fmt.Errorf("failed to install/update Helm chart %q: %w", constants.IstiodChartName, err)
 	}
+	if rev.Name == v1.DefaultRevision {
+		_, err := r.ChartManager.UpgradeOrInstallChart(ctx, r.getChartDir(rev, constants.BaseChartName),
+			values, r.Config.OperatorNamespace, getReleaseName(rev, constants.BaseChartName), &ownerReference)
+		if err != nil {
+			return fmt.Errorf("failed to install/update Helm chart %q: %w", constants.BaseChartName, err)
+		}
+	}
```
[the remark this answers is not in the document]
nrfox: Basically are we worried about the description text of this PR
> As only one instance of this chart can be installed at once, this has to be done by a new controller. If we had added the rendering of this chart to the istiorevision controller, we could have two instances of that controller both trying to update the chart/release.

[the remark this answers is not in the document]
nrfox: > a) we only ever run one instance of each controller anyway

I think this is configurable though with `MaxConcurrentReconciles` and might be an option we want to set in the future to handle more scale.

> b) we'd have the same problem with every other chart/release.

The difference, I think, is that normally objects go into a workqueue key'd off their namespace/name and the objects are then popped off and earmarked so you can only have one instance of a controller working on that CR at a time. If each instance of a controller does not use the workqueue to coordinate then you could have multiple controllers each trying to install the base chart in their reconcile loop. We can test this by increasing `MaxConcurrentReconciles`.

> it shouldn't really be a problem if multiple controllers touch the chart, in the end the only thing that counts is that there's always just 1 instance installed, and that's ensured by using the same release name.

There would probably be some error logs around failing to install the release but they would probably settle out. At any rate I don't think it'd be an issue until we set `MaxConcurrentReconciles` above 1. There may even be other issues with setting `MaxConcurrentReconciles > 1` that we don't know about yet so we can create a separate issue for that and, if need be, refactor the chart installation at that point?

PR #804, opened 2025-05-06
Adding new examples to docs automation

### LINKED ISSUE #781: Tag examples over the docs to add the topics to the documentation testing workflow

fjglira:
#### What type of PR is this?

- [ ] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [x] Test
- [x] Documentation Update

Adding new examples to the automation docs tests workflow:
* CNI update test
* Update strategy
* Multiple meshes
* Dual-stack

Fixes #781

Related Issue/PR #

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -668,7 +668,7 @@ Steps:
     kubectl get istio
     kubectl get istiorevision -n istio-system
     with_retries istio_active_revision_match "default-v1-23-2"
-    istio_revisions_ready_count "2"
+    with_retries istio_revisions_ready_count "2"
 ``` -->
 11. Confirm there are two control plane pods running, one for each revision.
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -652,9 +652,7 @@ Steps:
     ```
 <!-- ```bash { name=validation-wait-istio-updated tag=revision-based-update}
     . scripts/prebuilt-func.sh
-    old_pod=$(kubectl get pods -n istio-system -l app=istiod -o name)
-    kubectl patch istio default -n istio-system --type='merge' -p '{"spec":{"version":"v1.23.2"}}'
-    kubectl wait --for=delete $old_pod -n istio-system --timeout=60s
+    with_retries istiod_pods_count "2"
     wait_istio_ready "istio-system"
     print_istio_info
 ``` -->
--- a/docs/README.md
+++ b/docs/README.md
@@ -893,7 +893,11 @@ Steps:
     istiod-default-v1-23-3-c98fd9675-r7bfw    1/1     Running   0          10m
     istiod-default-v1-24-1-7495cdc7bf-v8t4g   1/1     Running   0          113s
     ```
-
+<!-- ```bash { name=validation-istiod-running tag=istiorevisiontag}
+    . scripts/prebuilt-func.sh
+    with_retries istiod_pods_count "2"
+    wait_istio_ready "istio-system"
+``` -->
 12. Confirm the proxy sidecar version remains the same:
 
     ```bash { name=validation-istio-proxy-version tag=istiorevisiontag}
--- a/docs/README.md
+++ b/docs/README.md
@@ -195,11 +195,9 @@ print_cni_info
     kubectl patch istiocni default -n istio-cni --type='merge' -p '{"spec":{"version":"v1.24.3"}}'
     ```
 <!-- ```bash { name=validation-wait-cni tag=cni-update}
-echo "Sleeping for 5 seconds to wait for the patch to be applied"
-sleep 5s
 . scripts/prebuilt-func.sh
-wait_cni_ready "istio-cni"
 with_retries resource_version_equal "istiocni" "default" "v1.24.3"
+wait_cni_ready "istio-cni"
 ``` -->
 4. Confirm the CNI plugin version was updated.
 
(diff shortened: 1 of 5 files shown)


docs/README.md
```diff
@@ -267,7 +267,7 @@ is installed. `Succeeded` should appear in the **Status** column.
 1. Create the `openshift-operators` namespace (if it does not already exist).
 
     ```bash
-    $ kubectl create namespace openshift-operators
+    kubectl create namespace openshift-operators
```
FilipB, line 270: We should probably stay consistent and use $ everywhere or remove it

docs/README.md
```diff
@@ -290,7 +290,7 @@ is installed. `Succeeded` should appear in the **Status** column.
 
 1. Verify that the installation succeeded by inspecting the CSV status.
 
-    ```bash
+    ```console
```
FilipB, line 293: Why using `console` here and `bash` in other examples?

docs/README.md
```diff
@@ -650,27 +678,41 @@ Steps:
...
 14. Restart all Deplyments in the `bookinfo` namespace.
 
-    ```bash
+    ```bash { name=restart-bookinfo tag=revision-based-update}
     kubectl rollout restart deployment -n bookinfo
     ```
-
+<!-- ```bash { name=validation-wait-bookinfo tag=revision-based-update}
+    . scripts/prebuilt-func.sh
+    sleep 20s
```
FilipB, line 710: We should probably avoid hardcoded sleeps? Let's use `oc wait`?

tests/documentation_tests/scripts/prebuilt-func.sh
```diff
@@ -194,3 +198,72 @@ pods_istio_version_match() {
...
+    fi
+
+    echo "Found $revisions istiorevisions"
+}
+
+# Check the number of istiod pods running in the istio-system namespace
+istiod_pods_count() {
+    expected_count="$1"
+
+    pods=$(kubectl get pods -n istio-system -l app=istiod -o jsonpath='{.items[*].status.phase}')
```
FilipB, line 234: Following might be used too: `kubectl get pods -n istio-system -l app=istiod --output json | jq -j '.items | length'`

tests/documentation_tests/scripts/prebuilt-func.sh
```diff
@@ -194,3 +198,72 @@ pods_istio_version_match() {
...
+        echo "Expected istiorevision tag default to be $expected_status, got $status"
+        return 1
+    fi
+}
+
+
+istio_revision_tag_inuse() {
+    inuse="$1"
+
+    conditions=$(kubectl get istiorevisiontag default -o jsonpath='{.status.conditions}' 2>/dev/null)
```
FilipB, line 262: Maybe the tag name should be parametrized?

tests/documentation_tests/scripts/prebuilt-func.sh
```diff
@@ -194,3 +198,72 @@ pods_istio_version_match() {
...
+    istiod_revision_tag=$(echo "$conditions" | jq -r '.[] | select(.type == "InUse") | .status' | tr '[:upper:]' '[:lower:]' | xargs)
```
FilipB, line 263: Could you please add a comment what is this doing?

docs/README.md
fjglira: The console should be used only to show the command and the expected output, and bash for commands that you can copy/paste to use them directly

tests/documentation_tests/scripts/prebuilt-func.sh
fjglira: For examples is using the default name, but yes it can be parametrized and pass "default as value"

docs/README.md
fjglira: This is already being done on this PR:
* For all the consoles that are examples, commands and their output have the $ sign, because this is not intended to be copied and pasted by the user following the example
* For all the bash code block I deleted the $ sign because this code blocks are intended to be copied by the user to execute the command, if we leave there the $ sign it will result in this: `$ kubectl create namespace openshift-operators`

docs/README.md
fjglira: Ok, I changed the logic to not use hardcoded sleeps and use more checks on the rollout status, wait, etc

tests/documentation_tests/scripts/prebuilt-func.sh
fjglira: I added a note:
`# istiod_revision_tag will be the result of getting InUse status condition in lower case and without any spaces`
I got several issues checking the inUse status, so, to simplify the result and avoid failures, I'm converting the status to lowercase and removing surrounding whitespace

tests/documentation_tests/scripts/prebuilt-func.sh
```diff
@@ -231,7 +231,7 @@ istio_revisions_ready_count() {
 istiod_pods_count() {
     expected_count="$1"
 
-    pods=$(kubectl get pods -n istio-system -l app=istiod -o jsonpath='{.items[*].status.phase}')
+    pods=$(kubectl get pods -n istio-system -l app=istiod --output json | jq -j '.items | length')
 
     count=$(echo "$pods" | wc -w) # Correctly count the number of pods
```
FilipB, line 236: This is not needed, the command above is already returning count of pods I think

fjglira: Right, came from the previous change. Changing

FilipB (approved):

PR #821, opened 2025-05-15
Remove EOL'd istio versions

dgn:
This adds an `eol: true` setting to the versions defined in `versions.yaml`. Setting this will cause the version's charts to be removed from the repo, rendering it uninstallable, but it will keep the version as a valid enum value to avoid breaking API guarantees and having to bump CRD versions.

I also added an integration test that verifies that using a version marked as EOL results in a ReconcileError.

In a separate commit I renamed the `HaveCondition` function to `HaveConditionStatus`. As I added a `HaveConditionMessage` matcher that works on condition message, this makes more sense now.


dgn:
I did a small refactoring in a separate commit so I changed the merge method to rebase

dgn:
looks like the docs test failure was legit in this case. It was trying to install 1.22.5 and failing as I'm removing it

hack/remove-old-versions.sh
```diff
@@ -20,7 +20,7 @@ VERSIONS_YAML_FILE=${VERSIONS_YAML_FILE:-"versions.yaml"}
 VERSIONS_YAML_DIR=${VERSIONS_YAML_DIR:-"pkg/istioversion"}
 
 function removeOldVersions() {
-    versions=$(yq eval '.versions[].name' "${VERSIONS_YAML_DIR}/${VERSIONS_YAML_FILE}" | tr $'\n' ' ')
+    versions=$(yq eval '.versions[] | select(.ref == null) | select(.eol != true) | .name' "${VERSIONS_YAML_DIR}/${VERSIONS_YAML_FILE}" | tr $'\n' ' ')
```
FilipB, line 23: Should we remove the whole file. It's not going to be used anymore.

hack/update-version-list.sh
```diff
@@ -35,9 +35,9 @@ function updateVersionsInIstioTypeComment() {
       -e "s/(\+kubebuilder:default=.*version: \")[^\"]*\"/\1$defaultVersion\"/g" \
       api/v1/istio_types.go api/v1/istiocni_types.go
     
-    cniSelectValues=$(yq '.versions[]| select(. | has("ref") | not) | .name | ", \"urn:alm:descriptor:com.tectonic.ui:select:" + . + "\""' "${VERSIONS_YAML_PATH}" | tr -d '\n')
-    cniVersionsEnum=$(yq '.versions[]| select(. | has("ref") | not) | .name' "${VERSIONS_YAML_PATH}" | tr '\n' ';' | sed 's/;$//g')
-    cniVersions=$(yq '.versions[]| select(. | has("ref") | not) | .name' "${VERSIONS_YAML_PATH}" | tr '\n' ',' | sed -e 's/,/, /g' -e 's/, $//g')
+    cniSelectValues=$(yq '.versions[] | select (.eol != true) | select(.ref == null) | .name | ", \"urn:alm:descriptor:com.tectonic.ui:select:" + . + "\""' "${VERSIONS_YAML_PATH}" | tr -d '\n')
```
FilipB, line 38: CNI can't use aliases?

fjglira:
Hey @dgn, seems to be that the docs issue still is legit:
```
NAME              NAMESPACE      PROFILE   READY   STATUS           IN USE   VERSION   AGE
default-v1-25-3   istio-system             True    Healthy          True     v1.25.3   46s
default-v1-26-0   istio-system             False   IstiodNotReady   True     v1.26.0   8s
```

The failures:
```Pod details-v1-84dbc5d78d-gpc8x.bookinfo: got version 1.26.0, expected 1.24.1```

hack/remove-old-versions.sh
FilipB: I mean `hack/remove-old-versions.sh` is not needed anymore as we will only use `eol` flag now.

dgn:
> Hey @dgn, seems to be that the docs issue still is legit:
> 
> ```
> NAME              NAMESPACE      PROFILE   READY   STATUS           IN USE   VERSION   AGE
> default-v1-25-3   istio-system             True    Healthy          True     v1.25.3   46s
> default-v1-26-0   istio-system             False   IstiodNotReady   True     v1.26.0   8s
> ```
> 
> The failures: `Pod details-v1-84dbc5d78d-gpc8x.bookinfo: got version 1.26.0, expected 1.24.1`

thanks, I missed that one! fixed now

hack/update-version-list.sh
dgn: hmm. yeah, strange, I didn't introduce this, but it seems we're not adding the aliases to the dropdown, but it is possible

hack/remove-old-versions.sh
dgn: it's still needed, e.g. when branching off of `main`, and to delete `eol: true` charts

FilipB: ok, thanks. I will try it with 1.26 now.

FilipB: Sorry, I had bad assumption. I checked how is this script used and now I understand.

hack/update-version-list.sh
FilipB: Sorry, I'm blind. These values are for `istiorevision_types.go`. Values for `istiocni_types.go` correctly contain the aliases.

FilipB (approved):

PR #826, opened 2025-05-19
Adding CNI optional for smoke test cases in the e2e test framework

fjglira:
#### What type of PR is this?

- [ ] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [x] Test
- [ ] Documentation Update

This PR adds the possibility to skip the test steps that install/validate CNI. This will allow the user to validate a custom third-party CNI installation. The PR adds a wrapper over When and DescribeTable context from Ginkgo to enable the skipping of these nodes based on the use of `SKIP_CNI` env var. 

Besides this, also this PR move CNI steps that where outside the logic of the test for CNI

Fixes #

Related Issue/PR #

#### Additional information:
Pending:
* Check if we can add the use of the wrapper also for the non-smoke test group: ambient and multicluster. As they are now, this test will need more changes to the structure to enable the use of the wrapper


sridhargaddam:
> This will allow the user to validate a custom third-party CNI installation.

PR #836, opened 2025-05-21
Update the operator description as it's no longer experimental

FilipB:


chart/values.yaml
```diff
@@ -12,9 +12,9 @@ csv:
   displayName: Sail Operator
   categories: OpenShift Optional, Integration & Delivery, Networking, Security
   capabilities: Seamless Upgrades
-  description: Experimental operator for installing Istio service mesh
+  description: The Sail Operator manages the lifecycle of your Istio control plane. It provides custom resources for you to deploy and manage your control plane components.
```
sridhargaddam, line 15: The description and longDescription seems to be the same - there is no harm in it, just sharing the observation.

sridhargaddam (approved):

dgn (approved):

chart/values.yaml
FilipB: Long description also contains list of supported versions and a link to our documentation. Not sure if we want to extend this or this is fine.

dgn (approved):

PR #843, opened 2025-05-22
add network policy securing operator traffic

cam-garrison:
#### What type of PR is this?

- [ ] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [ ] Test
- [ ] Documentation Update

This PR adds a network policy to the operator  which locks down traffic to and from our operator. Specifically, it allows ingress to port 8443 for metrics, and allows egress to DNS and to the API server. 

It also adds an e2e test ensuring the NP is created with the correct configuration. Tested by running `GINKGO_FLAGS="-v --label-filter=operator" make test.e2e.ocp` 
**NOTE: the final piece metrics test above my test was failing, not sure if this is a known issue but it was unrelated to my changes and I had to disable that test and the corresponding part of the cleanup steps as well.  

The network policy is implemented following internal RH guidance NP doc linked in the below issue. 

Since OLM currently does not support adding a NetworkPolicy into the bundle, we have to create it programatically. I hook the creation logic into main.go, and currently only create the NP if the platform is OCP. 

Also tested by deploying on local cluster and ensuring operator can still deploy a basic istio setup.

Fixes # https://issues.redhat.com/browse/OSSM-9155

Related Issue/PR #


nrfox (commented):
I don't have much context around this but this is just a network policy that needs to be included with the operator? Is this an openshift specific thing? Can we add the net policy YAML to the CSV bundle that the operator installs with?

cam-garrison:
Not sure why the multicluster tests are failing, they are passing locally on my machine :/ adding more test coverage tomorrow, I was .35% off :sweat_smile:

pkg/kube/networkpolicy.go
```diff
@@ -0,0 +1,206 @@
...
+
+var log = ctrl.Log.WithName("netpol")
+
+// CreateNetworkPolicy creates a NetworkPolicy for the sail-operator during startup
+// Uses a direct clientset since this runs before the manager cache is started
+func CreateNetworkPolicy(cfg *rest.Config, namespace string, platform config.Platform) error {
+	clientset, err := kubernetes.NewForConfig(cfg)
+	if err != nil {
+		return err
+	}
```
nrfox, line 39: I don't think we are using the underlying kube client anywhere else. Can we pass the controller-runtime client from the manager? Might be easier to pass the manager itself since you can use the APIReader to check if the object exists and then use the client to create it. That way you don't establish watches on NetworkPolicies when doing a `client.Get(...)`.

pkg/kube/networkpolicy.go
```diff
@@ -0,0 +1,206 @@
...
+
+	networkPolicy := buildNetworkPolicy(namespace, platform)
+
+	ctx := context.Background()
```
nrfox, line 43: Does this need a timeout? I think these get/create requests will block the operator from reconciling resources until these requests complete.

Actually I think you'll want to inherit from the context returned by: `ctrl.SetupSignalHandler` so that these requests will be canceled when interrupt is given. So can you pass that context in here and add a reasonable timeout to it?

tests/e2e/operator/operator_install_test.go
```diff
@@ -185,6 +189,93 @@ spec:
 			))
 		})
 
+		It("creates the network policy", func(ctx SpecContext) {
+			isOpenShift := isOpenShiftCluster()
+
+			if !isOpenShift {
+				Skip("Skipping NetworkPolicy test - not running on OpenShift platform")
+			}
```
nrfox, line 197: Do we have test tags for `isOpenshift` instead of detecting this at runtime @fjglira?

nrfox (commented):
I like the approach of having this as a separate method that gets run on startup rather than a whole separate controller.

pkg/kube/networkpolicy.go
cam-garrison: Unfortunately I don't think I can do this, since these steps are occurring before the manager has been started at the end of the main.go block. So, when I initially tried passing the manager/manager's client to my function, it was erroring as the manager client's cache did not yet exist. So I think using a separate client here is necessary.

tests/e2e/operator/operator_install_test.go
cam-garrison: That works for me. [79759cd](https://github.com/istio-ecosystem/sail-operator/pull/843/commits/79759cd55157cf9adacd84da0b6b586efedd7b7d) I've removed the e2e test for now and can create a new test file in the midstream repo after this is merged.

fjglira: Great, I like that better, with that we are sure that we will run that test only for OCP, and we are not adding any custom setup or validation for OCP clusters

dgn:
If you update the `OPERATOR_SDK_VERSION` to `1.40.0` in the Makefile, you can add the `networkpolicy.yaml` to the chart, and `make bundle` will then automatically add it to the bundle. That should be enough

cam-garrison:
Following discussion on tuesday, I am closing this PR. This (securing operator traffic via NP) will wait until OLM supports adding NetworkPolicy resources into bundles.

PR #846, opened 2025-05-22
Ignore env vars with multiline values. They are failing make targets if present.

libesz:
#### What type of PR is this?

- [x] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [ ] Test
- [ ] Documentation Update

My dev machine contains several shell plugins, like FZF. That plugin has env var settings, where one of them has multiline value. `env` command represents multiline values in a way that they are actually split into multiple lines, which makes parsing the command output hard. When sail-operator make targets are consuming the `env` command output, they can fail when multiline env vars are present. That is, I assume we never need multiline variable values for real, so we can unset them to resolve the issue.

Fixes #

Related Issue/PR #

### DIFF
--- a/common/scripts/setup_env.sh
+++ b/common/scripts/setup_env.sh
@@ -116,6 +116,21 @@ do
   unset -f "${f}"
 done
 
+# Remove variables with multiline values. They are also messing up the output of `env`.
+while IFS= read -r varname; do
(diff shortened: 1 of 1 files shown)


libesz:
Thanks @dgn . There it is: https://github.com/istio/common-files/pull/1185

PR #852, opened 2025-05-27
Fix issue 851: wrong behaviour in ApplyVendorDefaults

### LINKED ISSUE #851: [BUG] Default Vendor does not have expected behaviour

### Bug Description

In the vendor default api, we have `ApplyVendorDefaults`, and the behaviour that is currently having is wrong. We use:
```
mergedValues := helm.Values(mergeOverwrite(helm.FromValues(values), vendorDefaults[version]))
	return helm.ToValues(mergedValues, &v1.Values{})
```

To merge the vendor default values with the user-given values and return the resulting values for the `Istio` resource

**Expected behavior:**
User-given values are kept over the default vendor values set in the default vendors yaml

**Current behavior:**
Vendor default values are kept instead of the user-given values

### Operator Version

latest

### Link to Gist with Logs

_No response_

### Additional Information

_No response_

fjglira:
#### What type of PR is this?

- [x] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [ ] Test
- [ ] Documentation Update

In `ApplyVendorDefaults`, the expected behaviour is to take default vendor values and override them with user-given values. To achieve this in the `mergeOverwrite` call, we need to invert the position of the parameters: the base is the default vendor values, and the override is the user-given values

Fixes #851

Related Issue/PR #

### DIFF
--- a/pkg/istiovalues/vendor_defaults.go
+++ b/pkg/istiovalues/vendor_defaults.go
@@ -45,6 +45,6 @@ func ApplyVendorDefaults(version string, values *v1.Values) (*v1.Values, error) 
 	if len(vendorDefaults) == 0 {
 		return values, nil
 	}
-	mergedValues := helm.Values(mergeOverwrite(helm.FromValues(values), vendorDefaults[version]))
+	mergedValues := helm.Values(mergeOverwrite(vendorDefaults[version], helm.FromValues(values)))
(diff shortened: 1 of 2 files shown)


fjglira:
> Thanks! Could you add a unit test?

Sure

dgn (approved):

PR #864, opened 2025-06-02
[release-1.26] Bump operator to 1.26.1 + add istio 1.26.1 + update deps

FilipB:


dgn (approved):
I guess technically we'll want to add 1.26.1 on the main branch first, but we can also do it separately

FilipB:
> I guess technically we'll want to add 1.26.1 on the main branch first, but we can also do it separately

It should be done automatically but the update deps GH action is failing now - https://github.com/istio-ecosystem/sail-operator/actions/runs/15384229624/job/43279846348

PR #867, opened 2025-06-04
chore(docs): cleans up IstioCNI section

bartoszmajsak:
- [ ] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [ ] Test
- [x] Documentation Update

The paragraph about `IstioCNI` resource mentions that the snippet below consist of `spec.version` which is not the case. Then in the `NOTE` section below it repeats the same.

This change keeps the mention about `spec.version` in the note to avoid confusion.


fjglira (approved):

sridhargaddam (approved):

fjglira:
@bartoszmajsak gencheck is failing because you need to run `make gen` or, to be quicker, `make update-examples-docs`. I already have an issue created to change the behaviour of the doc test run and create the docs on the flight to avoid this issue (Recommended by you  :) )

PR #889, opened 2025-06-09
Improve e2e test cleanup by recording & cleaning up

mkolesnik:
#### What type of PR is this?

- [x] Enhancement / New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [ ] Test
- [ ] Documentation Update

#### What this PR does / why we need it:
This set of commits adds a `Cleaner` type which can record the cluster state and revert to that state by removing resources added via tests.

Currently only pertinent resources are tracked for the sake of simplicity, but this can be further expanded and generalized if necessary.

Cleaning up the operator is handled automatically, and when the operator is deployed externally if would be skipped (as it's recorded at the start of the test).

#### Which issue(s) this PR fixes:

Fixes #833 

Related Issue/PR #

PR #890, opened 2025-06-09
Deleting the Ambient resources

pbajjuri20:
If the first test case fails in the FIPS and IBM P&Z cluster, the resources are still available. However, the second test case is not running properly.  If we cleanup the resources properly then the remaining test cases are running perfectly.

### DIFF
--- a/tests/e2e/ambient/ambient_test.go
+++ b/tests/e2e/ambient/ambient_test.go
@@ -343,6 +343,15 @@ spec:
 				}
 			}
 
+			By("Cleaning up the Istio resource")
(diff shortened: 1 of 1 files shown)


fjglira:
Hey @pbajjuri20, this approach will not work when the test ends correctly without any issue because the resources do not exist anymore and are trying to be deleted. I suggest you wait until this [PR](https://github.com/istio-ecosystem/sail-operator/pull/889) is merged and add a new record for the cleaner to delete any dangling resources there

pbajjuri20:
Thanks @fjglira : 

this approach worked locally, the resources were deleted successfully and created the next test cases run perfectly. Sure will wait for the cleaner PR.

PR #894, opened 2025-06-10
Improve automated docs test run

### LINKED ISSUE #857: Improve run doc test copy docs files
We can improve the doc test by not duplicating the md files in the testing folder. How I see it can be done:

- [x]  Copy the files on the fly while running the make target

- [x] This mean include the logic to convert the comments into code blocks before the execution, copy the files to a tmp folder and delete those files after the execution

- [x] Change the make target and allow the user to execute a specific test by defining the test name directly in the target

fjglira:
#### What type of PR is this?

- [ ] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [ ] Test
- [X] Documentation Update

Changing how the docs tests are being executed:
* Instead of copying the files with make gen and leaving the docs examples that are going to run always in the test dir, we will copy the files before running the test, and after the test is done, we will delete those files
* Enable the possibility to execute only one tag, which will allow users to test locally the documentation easily

Fixes #857

Related Issue/PR #


Makefile.core.mk
```diff
@@ -208,7 +208,6 @@ test.e2e.kind: istioctl ## Deploy a KinD cluster and run the end-to-end tests ag
 .PHONY: test.e2e.describe
 test.e2e.describe: ## Runs ginkgo outline -format indent over the e2e test to show in BDD style the steps and test structure
 	GINKGO_FLAGS="$(GINKGO_FLAGS)" ${SOURCE_DIR}/tests/e2e/common-operator-integ-suite.sh --describe
-##@ Build
```
mkolesnik, line 211: Logically it makes sort of sense to move it, but then `runme` and `update-docs-examples` would be under the "Testing" category, which has only other tests (together with `test.docs`).

Maybe it's more sensible to have `test.docs` under "testing", and these targets should move under "Generated Code & Resources" (or maybe "Build Dependencies", though it seems less appropriate)?

fjglira: Yeah, I just figured out that they where under the wrong one while checking using make help. I'll move those

Makefile.core.mk
```diff
@@ -222,18 +221,21 @@ runme: $(RUNME) ## Download runme to bin directory. If wrong version is installe
...
 update-docs-examples: ## Copy the documentation files and generate the resulting md files to be executed by the runme tool.
 	@echo "Executing copy script to generate the documentation examples md files"
-	@echo "The script will copy the files from the source folder to the destination folder and add the runme suffix to the file names"
+	@echo "The script will copy the files from the source folder to test folder"
 	@tests/documentation_tests/scripts/update-docs-examples.sh
 	@echo "Documentation examples updated successfully"
 
-
+# Declare EXECUTE_DOC_TAG as an optional variable that can be set to a specific tag for the documentation examples.
+EXECUTE_DOC_TAG ?= ""
```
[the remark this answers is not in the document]
fjglira: Changed

tests/documentation_tests/scripts/update-docs-examples.sh
```diff
@@ -28,14 +28,15 @@ SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
...
-# Set the test directory to the test/documentation_tests/ directory
-TEST_DIR="$ROOT_DIR/tests/documentation_tests"
-# Set the output directory to the test/documentation_tests/ directory
-OUTPUT_DIR="$TEST_DIR"
 # Set the files to exclude from the copy
 EXCLUDE_FILES=(
   "sailoperator.io.md" "guidelines.md"
 )
+# Directory to store the output files
+TEST_DIR="${TEST_DIR:-$(mktemp -d)}"
```
mkolesnik, line 36:
```suggestion
ARTIFACTS="${ARTIFACTS:-$(mktemp -d)}"
TEST_DIR="${ARTIFACTS}/docs.test"
```

Shouldn't it default to the same values as in the run script?

Also since it's not really expected to run this script directly, you can alternatively just check that the variable is set which makes more sense IMHO.

fjglira:
Hey @mkolesnik, I made some changes from your review. I would like to check the update script if the variable exists, and if not, we are assuming that it is being tried to executed standalone, so we exit the script with a failure

PR #905, opened 2025-06-12
Revert: Disable imageDigest support in ztunnel controller

sridhargaddam:
ztunnel image can now be overridden using the imageOverrides.


FilipB:
Not sure what this means for downstream konflux build. cc @ferhoyos, @zmiklank

sridhargaddam:
> Not sure what this means for downstream konflux build. cc @ferhoyos, @zmiklank

This PR is required for supporting downstream `ztunnel` using OSSM 3.1 operator.  
CC: @MaxBab

FilipB:
@ferhoyos Downstream build for ztunnel was not available earlier so we just put there pilot [image](https://github.com/openshift-service-mesh/sail-operator/blob/release-3.0/bundle/manifests/servicemeshoperator3.clusterserviceversion.yaml#L670) which was ignored and upstream ztunnel image was used instead. Now it can be updated to use a proper downstream ztunnel build.

dgn (approved):

PR #907, opened 2025-06-13
Adding Eventually to Helm operator install

fjglira:
#### What type of PR is this?

- [ ] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [x] Test
- [ ] Documentation Update

All of our Helm installations of the operator are done using Expected; when Expected fails once time it fails the test run directly. What is happening is that when we run the test over the OCP cluster, it takes more time to clean the previous resources, and it directly fails the Helm installation with the error: `INSTALLATION FAILED: cannot re-use a name that is still in use`. This PR changes the Expected for eventually adding retries to the installation of the operator. This error is only happening over OCP cluster on kind runs without any issues

Fixes #

Related Issue/PR #

### DIFF
--- a/tests/e2e/ambient/ambient_test.go
+++ b/tests/e2e/ambient/ambient_test.go
@@ -54,7 +54,7 @@ var _ = Describe("Ambient configuration ", Label("smoke", "ambient"), Ordered, f
 		if skipDeploy {
 			Success("Skipping operator installation because it was deployed externally")
 		} else {
-			Expect(common.InstallOperatorViaHelm()).
+			Eventually(common.InstallOperatorViaHelm).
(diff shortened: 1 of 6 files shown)


dgn (approved):
works for me, but is there no way to make the cleanup blocking so we don't hit this?

sridhargaddam (approved):

sridhargaddam:
> works for me, but is there no way to make the cleanup blocking so we don't hit this?

+1 (would be more cleaner if we block the cleanup)

fjglira:
> works for me, but is there no way to make the cleanup blocking so we don't hit this?

I'll take a look.

PR #914, opened 2025-06-16
Fix helm issue during e2e test run  operator installation

fjglira:
#### What type of PR is this?

- [ ] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [x] Test
- [ ] Documentation Update

The operator namespace was being created during the image build/push only for OCP test runs, leading to a 'helm name error' in subsequent tests case. This happened because the cleaner record was incorrectly preventing the deletion of the sail-operator namespace.

Fixes #

Related Issue/PR #


tests/e2e/setup/build-and-push-operator.sh
```diff
@@ -81,3 +81,15 @@ if [ "${OCP}" == "true" ]; then
 fi
 
 build_and_push_operator_image
+
+# Workaround for OCP helm operator installation issues:
+# We create the operator namespace before the tests run
+# To avoid any cleanup issues, after we build annd push the image we check if the namespace exists and delete it if it does.
```
mkolesnik, line 87:
```suggestion
# To avoid any cleanup issues, after we build and push the image we check if the namespace exists and delete it if it does.
```

Small typo

tests/e2e/setup/build-and-push-operator.sh
```diff
@@ -81,3 +81,15 @@ if [ "${OCP}" == "true" ]; then
...
+# The test logic already handle the namespace creation and deletion during the test run.
```
mkolesnik, line 88:
```suggestion
# The test logic already handles the namespace creation and deletion during the test run. 
```
Small typo

mkolesnik (commented):
Couple of typos.

Also I'm not sure I understand why this needs to be specifically in the `build-and-push-operator` script and not, for example, the scripts that actually run the tests?
e.g. `integ-suite-ocp.sh`

fjglira:
> Couple of typos.
> 
> Also I'm not sure I understand why this needs to be specifically in the `build-and-push-operator` script and not, for example, the scripts that actually run the tests? e.g. `integ-suite-ocp.sh`

Hey, I did some changes:

- Create a new namespace to push the operator image. This will avoid any issues while deleting the operator namespace. Fun fact, this was working before because once we pulled the first time the image from the operator namespace, it was in cache, no matter that we deleted and created again the operator namespace (Started to fail to me today after I started to delete the operator namespace before starting the test)
- Set the proper roles for the new namespace and the operator namespace to be able to pull and push the image
- Moved the deletion of the operator namespace to the common script that actually runs the test after validating that we are running for OCP. Note: we still need to create the operator namespace to create the roles to pull the image

tests/e2e/setup/build-and-push-operator.sh
```diff
@@ -37,9 +37,12 @@ get_internal_registry() {
   fi
 
   URL=$(${COMMAND} get route default-route -n openshift-image-registry --template='{{ .spec.host }}')
-  export HUB="${URL}/${NAMESPACE}"
+
+  # Create the istio-images namespace to store the operator image and avoid errors during test run.
```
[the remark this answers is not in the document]
fjglira: I'll rephrase this comment

FilipB (approved):

PR #922, opened 2025-06-18
fix IstioRevisionTag uninstall when revision namespace changes

### LINKED ISSUE #917: [BUG] IstioRevisionTag can't be removed after referencing IstioRevisions in different namespaces

### Bug Description

We have a flaw in the operator where if you create an IstioRevisionTag that references an `Istio` or `IstioRevision` (I will just use `IstioRevision` below as an Istio is always resolved to an `IstioRevision`), it will install the corresponding helm chart in the namespace of the underlying `IstioRevision`. If you now switch the reference to an `IstioRevision` that has a different `spec.namespace` than the previous, the tag will attempt to install its chart into that namespace - which fails as it's already installed elsewhere. But the real problem appears when you now try to remove the `IstioRevisionTag` - it cannot be removed, as the finalizer is never removed due to a helm uninstall error: the operator tries to uninstall the helm chart in the new namespace, but it doesn't exist there.

To fix this, we need to change our approach, probably it makes sense to install the tag's chart into the operator namespace. But we also need a migration strategy for existing deployments.

### Operator Version

1.26.1

### Link to Gist with Logs

_No response_

### Additional Information

_No response_

dgn:
Fixes #917.

This is quite a complex edge case, I'll do my best to explain. If you look at the integration test I wrote (part of this commit), you can follow along.

When a tag was created that referenced a revision in ns1, then patched to reference a revision in ns2, we failed to clean up the old helm install in ns1. That almost never causes problems - except in the case where afterwards, you remove all tags, and then create a new one with the same name that is never reconciled. If that tag is deleted, the operator will attempt to uninstall that lingering helm release, which fails, leading the operator to deadlock, and stopping the user from deleting the tag. The only way to get the operator back into a working state would be to manually delete that lingering helm release. That is a pretty significant bug in my opinion.

### DIFF
--- a/controllers/istiorevisiontag/istiorevisiontag_controller.go
+++ b/controllers/istiorevisiontag/istiorevisiontag_controller.go
@@ -110,6 +110,13 @@ func (r *Reconciler) doReconcile(ctx context.Context, tag *v1.IstioRevisionTag) 
 		return nil, reconciler.NewValidationError("IstioRevisionTag cannot reference a remote IstioRevision")
 	}
 
+	// if the IstioRevision's namespace changes, we need to completely reinstall the tag
+	if tag.Status.IstiodNamespace != "" && tag.Status.IstiodNamespace != rev.Spec.Namespace {
+		if err := r.uninstallHelmCharts(ctx, tag); err != nil {
+			return nil, err
+		}
+	}
+
 	log.Info("Installing Helm chart")
 	return rev, r.installHelmCharts(ctx, tag, rev)
 }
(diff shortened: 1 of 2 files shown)


dgn:
> > When a tag was created that referenced a revision in ns1, then patched to reference a revision in ns2, we failed to clean up the old helm install in ns1.
> 
> Isn't `spec.namespace` immutable for `Istio` and `IstioRevision`? This is effectively moving the tag then from one resource to another? Tags are allowed to float between resources?

Well, we could forbid that (ie make the `targetRef` immutable), but we currently don't, so yes, it's possible. I think it makes sense that it's possible, you might deploy an entirely new control plane and want to switch the tag over

nrfox:
> Well, we could forbid that (ie make the targetRef immutable), but we currently don't, so yes, it's possible.

The only risk I can see is that the `status.istiodNamespace` changes before the old helm release gets uninstalled but that doesn't seem likely with this change. Maybe there's a way to put ownerrefs on the helm release so it's tied to the lifecycle of the Tag resource? Anyways this LGTM.

PR #931, opened 2025-06-19
Consolidate e2e resource creation

mkolesnik:
#### What type of PR is this?

- [ ] Bug Fix
- [x] Refactor
- [ ] Optimization
- [ ] Test
- [ ] Documentation Update

Istio and IstioCNI creation is done in multiple places in the E2E tests, and is very repetitive.
This PR consolidates most og it, making the tests less repetitive and more maintainable.

Fixes #

Related Issue/PR #

### DIFF
--- a/tests/e2e/util/common/e2e_utils.go
+++ b/tests/e2e/util/common/e2e_utils.go
@@ -30,6 +30,7 @@ import (
 	"github.com/istio-ecosystem/sail-operator/pkg/env"
 	"github.com/istio-ecosystem/sail-operator/pkg/istioversion"
 	"github.com/istio-ecosystem/sail-operator/pkg/test/project"
+	. "github.com/istio-ecosystem/sail-operator/pkg/test/util/ginkgo"
 	"github.com/istio-ecosystem/sail-operator/tests/e2e/util/helm"
 	"github.com/istio-ecosystem/sail-operator/tests/e2e/util/istioctl"
 	"github.com/istio-ecosystem/sail-operator/tests/e2e/util/kubectl"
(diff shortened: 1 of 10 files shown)


tests/e2e/util/common/e2e_utils.go
```diff
@@ -409,3 +410,54 @@ func ResolveHostDomainToIP(hostDomain string) (string, error) {
...
+	yaml = fmt.Sprintf(yaml, istioCniName, version, istioCniNamespace)
+	Log("IstioCNI YAML:", indent(yaml))
+	Expect(k.CreateFromString(yaml)).To(Succeed(), withClusterName("IstioCNI creation failed", k))
+	Success(withClusterName("IstioCNI created", k))
+}
+
+func indent(str string) string {
+	indent := strings.Repeat(" ", 2)
+	return indent + strings.ReplaceAll(str, "\n", "\n"+indent)
+}
```
fjglira, line 455: Small comment only here, we already have this same function declared [here](https://github.com/istio-ecosystem/sail-operator/blob/9d0b023272b150fb2cb07f2e348606f99209ae3e/tests/e2e/controlplane/control_plane_test.go#L311) and being used in the same file. Can you please make it available in some way to use it in both places

tests/e2e/util/common/e2e_utils.go
```diff
@@ -409,3 +410,54 @@ func ResolveHostDomainToIP(hostDomain string) (string, error) {
...
+
+func withClusterName(m string, k kubectl.Kubectl) string {
+	if k.ClusterName == "" {
+		return m
+	}
+
+	return m + " on " + k.ClusterName
+}
```
fjglira, line 463: kubectl util already has a definition for this [here](https://github.com/istio-ecosystem/sail-operator/blob/9d0b023272b150fb2cb07f2e348606f99209ae3e/tests/e2e/util/kubectl/kubectl.go#L58-L59) it can be confusing in my opinion

fjglira (commented):
only two small comments, for the rest LGTM. I think is a great improve

tests/e2e/util/common/e2e_utils.go
mkolesnik: That one is a public function on the `Kubectl` type, whereas this one is private and won't be used outside the package.
I think it should be fine but if you think differently let me know and I'll change it

PR #936, opened 2025-06-19
Stabilize e2e dual-stack test

unsortedhashsets:
#### What type of PR is this?

- [ ] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [x] Test
- [ ] Documentation Update

Allows wait inside CheckPodsReady func till pods will be ready (avoid errors during dual-stack tests like 

```
[38;5;9m[FAILED] No pods found in dual-stack namespace
  Expected
      <[]v1.Pod | len:0, cap:0>: nil
  not to be empty[0m
```

Same principle as used here: https://github.com/istio-ecosystem/sail-operator/blob/main/tests/e2e/controlplane/control_plane_test.go#L221

Fixes #

Related Issue/PR #

### DIFF
--- a/tests/e2e/controlplane/control_plane_update_test.go
+++ b/tests/e2e/controlplane/control_plane_update_test.go
@@ -146,16 +146,9 @@ spec:
 					Success("sample deployed")
 
 					samplePods := &corev1.PodList{}
+					Eventually(common.CheckPodsReady).WithArguments(ctx, cl, sampleNamespace).Should(Succeed(), "Error checking status of sample pods")
+					Expect(cl.List(ctx, samplePods, client.InNamespace(sampleNamespace))).To(Succeed(), "Error getting the pods in sample namespace")
 
-					Eventually(func() bool {
-						Expect(cl.List(ctx, samplePods, client.InNamespace(sampleNamespace))).To(Succeed())
-						return len(samplePods.Items) > 0
-					}).Should(BeTrue(), "No sample pods found")
-
-					for _, pod := range samplePods.Items {
-						Eventually(common.GetObject).WithArguments(ctx, cl, kube.Key(pod.Name, sampleNamespace), &corev1.Pod{}).
-							Should(HaveConditionStatus(corev1.PodReady, metav1.ConditionTrue), "Pod is not Ready")
-					}
 					Success("sample pods are ready")
 
 					for _, pod := range samplePods.Items {
--- a/tests/e2e/util/common/e2e_utils.go
+++ b/tests/e2e/util/common/e2e_utils.go
@@ -29,9 +29,7 @@ import (
 	"github.com/Masterminds/semver/v3"
 	"github.com/istio-ecosystem/sail-operator/pkg/env"
 	"github.com/istio-ecosystem/sail-operator/pkg/istioversion"
-	"github.com/istio-ecosystem/sail-operator/pkg/kube"
 	"github.com/istio-ecosystem/sail-operator/pkg/test/project"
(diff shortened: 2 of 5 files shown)


sridhargaddam:
@unsortedhashsets did you see the error with Kind setup or OCP deployments?

unsortedhashsets:
> @unsortedhashsets did you see the error with Kind setup or OCP deployments?
With OCP, in general problem was in too fast check of pods existing

tests/e2e/util/common/e2e_utils.go
```diff
@@ -294,16 +294,15 @@ func GetVersionFromIstiod() (*semver.Version, error) {
 func CheckPodsReady(ctx SpecContext, cl client.Client, namespace string) (*corev1.PodList, error) {
```
fjglira, line 294: What if we remove all the eventually inside the function and we call the eventually from outside, I mean. Something like this:
* Currently, we are using the function:
```
_, err = common.CheckPodsReady(ctx, cl, DualStackNamespace)
						Expect(err).ToNot(HaveOccurred(), fmt.Sprintf("Error checking status of dual-stack pods: %v", err))
```
* Change the use for this:
````
Eventually(common.CheckPodsReady).WithArguments(ctx, cl, DualStackNamespace).Should(Succeed(), "Error checking status of dual-stack pods")
```

With this, we will use only one eventually, and we will simplify the code in the test. The only downside of this is that the function CheckPodsReady currently returns `*corev1.PodList` and `error`, and my suggestion will not work with those returns. But in my opinion, the function should only check that the pods in the namespace are ready, and after we check that the pods are ready and not return a list of pods, we can get the pods of the namespace in a following step to be used in later validations. Wdyt @sridhargaddam ?

fjglira: Yeah, in that case, I suggest avoiding using a function with the name CheckPodsReady to return a list of pods. So, we can check the pods are ready in the ns with a function that only checks the state of the pods, and then we can get the list of the pods from the ns to use it later in the next steps. We already get only the pods like that using the client:
For example:
```
					samplePodsCluster2 := &corev1.PodList{}
					It("updates the pods status to Ready", func(ctx SpecContext) {
						Expect(clRemote.List(ctx, samplePodsCluster2, client.InNamespace(sampleNamespace))).To(Succeed())
						Expect(samplePodsCluster2.Items).ToNot(BeEmpty(), "No pods found in sample namespace")
```

fjglira: @unsortedhashsets I mean this is not something strictly related to your PR, I'm just seeing that this can be improved also at the same time with your PR. Using two eventually inside the same function seems a little redundant for me, we can simplified easily moving our the eventually

unsortedhashsets: Thanks for clarification, will try that approach

tests/e2e/controlplane/control_plane_update_test.go
```diff
@@ -146,16 +146,9 @@ spec:
 					Success("sample deployed")
 
 					samplePods := &corev1.PodList{}
+					Eventually(common.CheckPodsReady).WithArguments(ctx, cl, sampleNamespace).Should(Succeed(), "Error checking status of sample pods")
+					Expect(cl.List(ctx, samplePods, client.InNamespace(sampleNamespace))).To(Succeed(), "Error deploying sample pod")
```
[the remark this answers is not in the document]
unsortedhashsets: Thanks, agree, it's describe operation better

PR #937, opened 2025-06-20
Adding a doc for no downtime migration from Istio CA generated certs to

FilipB:
user generated certs.

 

#### What type of PR is this?

- [ ] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [ ] Test
- [X] Documentation Update

### DIFF
--- a/docs/general/plugin-ca.md
+++ b/docs/general/plugin-ca.md
@@ -1,0 +1,143 @@
+[Return to Project Root](../../README.md)
+
+# Plug in CA Certificates
+Istio upstream [documentation](https://istio.io/latest/docs/tasks/security/cert-management/plugin-ca-cert/) is covering how to plug in user generated certificates to be used by the Istio CA. Here, we are documenting how to switch from Istio CA generated self-signed certificates to user generated certificates without any traffic disruptions even with strict mTLS enabled.
+
+## Switching from Istio CA generated self-signed certificates
+By default the Istio CA generates a self-signed root certificate and key and uses them to sign the workload certificates. This is not recommended in production but there might be cases where it's used already and we need to switch to better certificate management method. It's a simple [task](https://istio.io/latest/docs/tasks/security/cert-management/plugin-ca-cert/) if we are able to do that during a maintenance window where the traffic disruptions are expected. The same task with a no downtime requirement is more complex.
+
+### Problem with the simple workflow
+We can simply add a new `cacerts` secret with our certificates but what does it mean for service to service mTLS traffic? Consider `sleep` and `httpbin` demo applications. Both are sharing the same root of trust which is the Istio CA generated self-signed root certificate:
+```bash
+$ istioctl proxy-config secret deployment/sleep -n sleep -o json | jq -r '.dynamicActiveSecrets[1].secret.validationContext.trustedCa.inlineBytes' | base64 --decode | openssl x509 -text -noout
+Certificate:
+    Data:
+        Version: 3 (0x2)
+        Serial Number:
+            b0:3c:9a:d8:3f:ab:c8:a7:12:23:a8:65:d6:5e:76:c8
+        Signature Algorithm: sha256WithRSAEncryption
+        Issuer: O=cluster.local
+        Validity
+            Not Before: Jun 19 12:39:46 2025 GMT
+            Not After : Jun 17 12:39:46 2035 GMT
+        Subject: O=cluster.local
+        ...
+$ istioctl proxy-config secret deployment/httpbin -n httpbin -o json | jq -r '.dynamicActiveSecrets[1].secret.validationContext.trustedCa.inlineBytes' | base64 --decode | openssl x509 -text -noout
+Certificate:
+    Data:
+        Version: 3 (0x2)
+        Serial Number:
+            b0:3c:9a:d8:3f:ab:c8:a7:12:23:a8:65:d6:5e:76:c8
+        Signature Algorithm: sha256WithRSAEncryption
+        Issuer: O=cluster.local
+        Validity
+            Not Before: Jun 19 12:39:46 2025 GMT
+            Not After : Jun 17 12:39:46 2035 GMT
+        Subject: O=cluster.local
+        ...
+```
+
+Now, when you create your new `cacerts` secret with an intermediate certificate signed by a different root certificate, you need to restart all workloads to pick up the change. There will be a time window where one of the workloads is already using a new certificate signed by the new intermediate certificate while other workload is still using old one so the mTLS communication breaks.
+
+### No downtime update
+To achieve the no downtime update of the certificates, it's necessary to assure that all workloads at any given time are trusting certificates signed by either the old root or the new intermediate (which is signed by the new root). This can be achieved by enabling Istio's multi root support.
+
+1. Enable the multi root support:
+    ```bash
+    kubectl patch Istio default --type='merge' --patch-file=istio-patch.yaml
+    ```
+    Content of `istio-patch.yaml`:
+    ```yaml
+    apiVersion: sailoperator.io/v1
+    kind: Istio
+    spec:
+      values:
+        pilot:
+          env:
+            ISTIO_MULTIROOT_MESH: "true"
+        meshConfig:
+          defaultConfig:
+            proxyMetadata:
+          PROXY_CONFIG_XDS_AGENT: "true"
+    ```
+1. Prepare new root and intermediate certificates (you should be using trusted root CA for issuing the intermediate certificate). Here we are using tooling from istio repository:
+    ```bash
+    mkdir -p certs
+    pushd certs
+    make -f ../tools/certs/Makefile.selfsigned.mk root-ca
+    make -f ../tools/certs/Makefile.selfsigned.mk intermediate-cacerts
+    ```
+1. Create new `cacerts` secrets with old CA certificate, key and chain and combined root certificates:
+
+    Get the certificate and the key from existing Istio CA generated secrets and prepare combined root certificates:
+    ```bash
+    kubectl get secret istio-ca-secret -n istio-system -o jsonpath={.data.'ca-cert\.pem'} | base64 -d > ca-cert.pem
+    kubectl get secret istio-ca-secret -n istio-system -o jsonpath={.data.'ca-key\.pem'} | base64 -d > ca-key.pem
+    kubectl get secret istio-ca-secret -n istio-system -o jsonpath={.data.'ca-cert\.pem'} | base64 -d > cert-chain.pem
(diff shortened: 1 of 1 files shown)


docs/general/plugin-ca.md
```diff
@@ -0,0 +1,143 @@
+[Return to Project Root](../../README.md)
+
+# Plug in CA Certificates
+Istio upstream [documentation](https://istio.io/latest/docs/tasks/security/cert-management/plugin-ca-cert/) is covering how to plug in user generated certificates to be used by the Istio CA. Here, we are documenting how to switch from Istio CA generated self-signed certificates to user generated certificates without any traffic disruptions even with strict mTLS enabled.
+
+## Switching from Istio CA generated self-signed certificates
+By default the Istio CA generates a self-signed root certificate and key and uses them to sign the workload certificates. This is not recommended in production but there might be cases where it's used already and we need to switch to better certificate management method. It's a simple [task](https://istio.io/latest/docs/tasks/security/cert-management/plugin-ca-cert/) if we are able to do that during a maintenance window where the traffic disruptions are expected. The same task with a no downtime requirement is more complex.
```
fjglira, line 7: I suggest rephrase this: 
"It's a simple [task](https://istio.io/latest/docs/tasks/security/cert-management/plugin-ca-cert/) if we are able to do that during a maintenance window where the traffic disruptions are expected. The same task with a no downtime requirement is more complex."
to
"It can be a simple [task](https://istio.io/latest/docs/tasks/security/cert-management/plugin-ca-cert/) if traffic disruptions can be allowed, for example, during a maintenance window. The same task with no downtime can be more complex."

docs/general/plugin-ca.md
```diff
@@ -0,0 +1,143 @@
...
+
+### Problem with the simple workflow
```
fjglira, line 9:
```suggestion
### Change the certificate with traffic disruption
```

docs/general/plugin-ca.md
```diff
@@ -0,0 +1,143 @@
...
+        Signature Algorithm: sha256WithRSAEncryption
+        Issuer: O=cluster.local
+        Validity
+            Not Before: Jun 19 12:39:46 2025 GMT
+            Not After : Jun 17 12:39:46 2035 GMT
+        Subject: O=cluster.local
+        ...
+```
+
+Now, when you create your new `cacerts` secret with an intermediate certificate signed by a different root certificate, you need to restart all workloads to pick up the change. There will be a time window where one of the workloads is already using a new certificate signed by the new intermediate certificate while other workload is still using old one so the mTLS communication breaks.
```
fjglira, line 40:
```suggestion
Now, when you create your new `cacerts` secret with an intermediate certificate signed by a different root certificate, you need to restart all workloads to pick up the change. There will be a time window where one of the workloads is already using a new certificate signed by the new intermediate certificate while the other workload is still using the old one, so the mTLS communication breaks, causing traffic disruption.
```

docs/general/plugin-ca.md
```diff
@@ -0,0 +1,143 @@
...
+
+### No downtime update
```
fjglira, line 42:
```suggestion
### Change the certificate with  no-downtime update
```

docs/general/plugin-ca.md
```diff
@@ -0,0 +1,143 @@
...
+To achieve the no downtime update of the certificates, it's necessary to assure that all workloads at any given time are trusting certificates signed by either the old root or the new intermediate (which is signed by the new root). This can be achieved by enabling Istio's multi root support.
+
+1. Enable the multi root support:
+    ```bash
+    kubectl patch Istio default --type='merge' --patch-file=istio-patch.yaml
+    ```
+    Content of `istio-patch.yaml`:
```
fjglira, line 49: Better to first create the file with the content and then apply the patch. It will first show the content of the file. Or just do something like
````
cat <<EOF | kubectl apply -f- 
apiVersion: sailoperator.io/v1
kind: Istio
spec:
  values:
    pilot:
      env:
        ISTIO_MULTIROOT_MESH: "true"
    meshConfig:
      defaultConfig:
        proxyMetadata:
      PROXY_CONFIG_XDS_AGENT: "true"
EOF
```

docs/general/plugin-ca.md
```diff
@@ -0,0 +1,143 @@
...
+      values:
+        pilot:
+          env:
+            ISTIO_MULTIROOT_MESH: "true"
+        meshConfig:
+          defaultConfig:
+            proxyMetadata:
+          PROXY_CONFIG_XDS_AGENT: "true"
+    ```
+1. Prepare new root and intermediate certificates (you should be using trusted root CA for issuing the intermediate certificate). Here we are using tooling from istio repository:
```
fjglira, line 63: Add the link to the upstream repo to allow user to check directly there the content

docs/general/plugin-ca.md
```diff
@@ -0,0 +1,143 @@
...
+        --from-file=root-cert.pem=combined-root.pem \
+        --from-file=cert-chain.pem
+    ```
+1. Restart istiod to pick up new certificates:
+    ```bash
+    kubectl rollout restart deployment/istiod -n istio-system
+    ```
+1. Verify that all workloads are trusting both old and new roots, e.g. for httpbin:
+    ```bash
+    istioctl proxy-config secret deployment/httpbin -n httpbin -o json | jq -r '.dynamicActiveSecrets[1].secret.validationContext.trustedCa.inlineBytes' | base64 --decode
```
fjglira, line 94: Can we show an example output?

docs/general/plugin-ca.md
```diff
@@ -0,0 +1,143 @@
...
+    ```
+    > **_NOTE:_** It might be necessary to restart the workload if you only see one certificate.
+1. Update `combined-root.pem` by adding the new root certificate again, this will assure that workload certificates are regenerated after next step as the root certificate is updated:
```
fjglira, line 97:
```suggestion
1. Update `combined-root.pem` by adding the new root certificate again, this will ensure that workload certificates are regenerated after the next step as the root certificate is updated:
```

docs/general/plugin-ca.md
```diff
@@ -0,0 +1,143 @@
+[Return to Project Root](../../README.md)
```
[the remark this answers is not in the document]
FilipB: Do you have any suggestion where this content should go? Maybe we will need a new directory as this is more related to Istio than sail operator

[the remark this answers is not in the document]
fjglira: Yes, I think it can be in a new directory like Istio related it will be good

docs/general/plugin-ca.md
FilipB: We are not describing how to change the cert with traffic disruptions. We are describing what is actually causing the traffic disruption here. Maybe `Cause of the traffic disruptions`?

docs/README.md
```diff
@@ -42,6 +42,7 @@
   - [Validation](deployment-models/multiple-mesh.md#validation)
     - [Checking application to control plane mapping](deployment-models/multiple-mesh.md#checking-application-to-control-plane-mapping)
     - [Checking application connectivity](deployment-models/multiple-mesh.md#checking-application-connectivity)
+  - [Plug in CA Certificates](general/plugin-ca.md)
```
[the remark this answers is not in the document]
fjglira: Yeah, I think it can be a little confusing, only reading the title. Maybe move it to a new complete section, it can help

fjglira (approved):
lgtm

docs/general/plugin-ca.md
```diff
@@ -0,0 +1,194 @@
...
+
+# Plug in CA Certificates
+Istio [documentation](https://istio.io/latest/docs/tasks/security/cert-management/plugin-ca-cert/) is covering how to plug in user generated certificates to be used by the Istio CA but it's not describing a use case where it's necessary to switch from Istio CA generated self-signed certificates to user generated certificates without any traffic disruptions even with strict mTLS enabled. This missing use case is covered here.
+
+## Switching from Istio CA generated self-signed certificates
+By default the Istio CA generates a self-signed root certificate and key and uses them to sign the workload certificates. This is not recommended in production but there might be cases where it's used already and we need to switch to better certificate management method. It's a simple [task](https://istio.io/latest/docs/tasks/security/cert-management/plugin-ca-cert/) to be done during a maintenance window where the traffic disruptions are allowed. If the same task must be done outside of the maintenance window without any traffic disruptions, the procedure is more complex.
```
nrfox, line 7: > This is not recommended in production

Why is it not recommended for production?

docs/general/plugin-ca.md
```diff
@@ -0,0 +1,194 @@
...
+
+    Prepare`istio-patch.yaml`:
+    ```yaml
+    apiVersion: sailoperator.io/v1
+    kind: Istio
+    spec:
+      values:
+        pilot:
+          env:
+            ISTIO_MULTIROOT_MESH: "true"
```
nrfox, line 55: Do you need this setting if we are combining both roots into a single PEM?

docs/general/plugin-ca.md
```diff
@@ -0,0 +1,194 @@
...
+    ```
+1. Update `cacerts` secrets to use the new intermediate certificate, key and chain and updated combined root certificates:
+    ```bash
+    kubectl delete secret cacerts -n istio-system --ignore-not-found && \
+    kubectl create secret generic cacerts -n istio-system \
+        --from-file=intermediate/ca-cert.pem \
+        --from-file=intermediate/ca-key.pem \
+        --from-file=root-cert.pem=combined-root.pem \
+        --from-file=intermediate/cert-chain.pem
+    ```
```
nrfox, line 160: Why do we delete and recreate `cacerts`? Can we create `cacerts` once with the intermediate?

nrfox (commented):
Also late for review...

docs/general/plugin-ca.md
FilipB: For security reason, root CA private key should be stored on externally. With this default configuration, it's possible to extract it from istio-ca-secret secret. Should I extend this with explanation?

docs/general/plugin-ca.md
FilipB: `If enabled, mesh will support certificates signed by more than one trustAnchor for ISTIO_MUTUAL mTLS`
Not sure if it makes sense to reference Istio upstream doc for clarity here, I can add it.

docs/general/plugin-ca.md
nrfox: > Problem with doing that in one step is that there will be a time period where some workloads are already updated (using combined root-cert.pem and new intermediate cert) while others are not yet updated and using old root-cert.pem with a single root. 

This is the point of the multiroot setting though I thought? Before updating any workloads to be signed by the new CA, all workloads should already trust both CAs. Then when you update the workloads to be signed by the new CA, old workloads will already trust it.

PR #941, opened 2025-06-23
Issue 666

### LINKED ISSUE #666: Speed up test by grouping common setup on Before and After Suite
We have some test suites with multiple specs; these specs had the same setup as beforeall and afterall. For example: multicluster test. I'll do this to save some time in the e2e test execution after I finish https://github.com/istio-ecosystem/sail-operator/issues/360

fjglira:
#### What type of PR is this?

- [ ] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [ ] Test
- [ ] Documentation Update

Small refactor to move the operator installation to outside the test file because operator installation is not part of the test it's part of the setup. With this if we increase the test files under the same suite they will share setup like is already done in the control plane suite

Fixes #666

Related Issue/PR #


tests/e2e/ambient/ambient_suite_test.go
```diff
@@ -64,4 +71,29 @@ func setup() {
...
+
+	if skipDeploy {
+		Success("Skipping operator installation because it was deployed externally")
+	} else {
+		Expect(common.InstallOperatorViaHelm()).
+			To(Succeed(), "Operator failed to be deployed")
+	}
+
+	Eventually(common.GetObject).WithArguments(ctx, cl, kube.Key(deploymentName, operatorNamespace), &appsv1.Deployment{}).
+		Should(HaveConditionStatus(appsv1.DeploymentAvailable, metav1.ConditionTrue), "Error getting Istio CRD")
```
mkolesnik, line 89:
```suggestion
		Should(HaveConditionStatus(appsv1.DeploymentAvailable, metav1.ConditionTrue), "Operator not deployed")
```

I know it's not quite part of your change, but since you're already touching this code shouldn't it have the correct error message in case this fails?

mkolesnik (commented):
I think it's a good change, left some comments.

Also generally the code is very repetitive, in fact it's almost identical (only difference I see is the `namespace` variable name, which the value of is taken from `common.OperatorNamespace`).

I'm wondering if it's not better to put the actual functions passed to `BeforeSuite` and `ReportAfterSuite` in named functions in a common place, and pass these named functions?

fjglira:
> I think it's a good change, left some comments.
> 
> Also generally the code is very repetitive, in fact it's almost identical (only difference I see is the `namespace` variable name, which the value of is taken from `common.OperatorNamespace`).
> 
> I'm wondering if it's not better to put the actual functions passed to `BeforeSuite` and `ReportAfterSuite` in named functions in a common place, and pass these named functions?

I like this. Let me take a look and do the change

PR #949, opened 2025-06-26
Container: only use required packages

skitt:
#### What type of PR is this?
- [ ] Bug Fix
- [ ] Refactor
- [x] Optimization
- [ ] Test
- [ ] Documentation Update

This uses a two-stage build to produce a UBI-based image with only the required packages, in this case only the setup package. buildinfo is copied so that Clair has enough information to accurately scan the image for vulnerable UBI packages.

The result is a minimal image with only the operator as the attack surface (no shell etc.).

Fixes #

Related Issue/PR #

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,17 @@
-FROM registry.access.redhat.com/ubi10/ubi-micro:latest 
-
+FROM registry.access.redhat.com/ubi10/ubi:latest AS packager
 ARG TARGETOS TARGETARCH
+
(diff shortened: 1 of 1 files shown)


skitt:
Here are the scan results on the minimised image: https://quay.io/repository/skitt/sail-operator/manifest/sha256:7e5852e800cbfc4789dd5f0392ed86d8b08a5c38997cbbad323e14eba398f7d5?tab=packages

PR #956, opened 2025-06-29
Deploy operator in the E2E scripts

mkolesnik:
Given that we have the `Cleaner` and each test should be cleaning after itself, deploying the operator just once per test run will save some test run time.
An added value is the tests are simplified and can focus on what they need to achieve, instead of bothering with the operator setup.

Fixes: https://github.com/istio-ecosystem/sail-operator/issues/666

 

#### What type of PR is this?

- [ ] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [ ] Test
- [ ] Documentation Update

Fixes #

Related Issue/PR #


fjglira (commented):
lgtm, only a comment

Issue #920, opened 2025-06-18
[BUG] Rarely, an istiod helm install will fail, and then it'll also fail uninstall


mkolesnik:
### Is this the right place to submit this?

- [x] This is not a question about how to use the sail-operator

### Bug Description

While investigating #866 , some of the failures seemed to stem from `unexpected helm release status uninstalling` when the operator is trying to install/upgrade a helm release.

After managing to reproduce it, the actual error seems to stem from the operator first failing to *install* istiod:
`2025-06-18T05:57:24Z    INFO    ctrlr.istiorev  Resource not found. Retrying... {"IstioRevision": "default", "reconcileID": "9dbce8c4-5e2c-49ea-9435-40ab7df618f1", "error": "failed to install/update Helm chart \"istiod\": failed to install helm chart istiod: 1 error occurred:\n\t* the server could not find the requested resource\n\n"}`

Which gets followed up by a reconciliation which tries to *uninstall* this chart, but fails with:
`2025-06-18T05:57:25Z    LEVEL(-2)       ctrlr.istiorev  uninstall: Failed to delete release: [configmaps.r
bac.authorization.k8s.io "istio-sidecar-injector" is forbidden: User "system:serviceaccount:sail-operator:
sail-operator" cannot delete resource "configmaps" in API group "rbac.authorization.k8s.io" in the namespa
ce "istio-system"]      {"IstioRevision": "default", "reconcileID": "cc897094-60b0-4089-99dd-8ea26dbe230d"
}`

It seems that this then set the helm release into `uninstalling` status, which keeps failing the reconciliation with:
`2025-06-18T05:57:25Z    ERROR   ctrlr.istiorev  Reconciler error        {"IstioRevision": "default", "reconcileID": "5b9abc45-4a68-4f54-9a13-358f556ef9dd", "error": "failed to install/update Helm chart \"istiod\": unexpected helm release status uninstalling"}`

I'm adding the sail pod log, the tests output, and info from the system:

[sail.pod.log](https://github.com/user-attachments/files/20793407/sail.pod.log)
[failed.tests.log](https://github.com/user-attachments/files/20793432/failed.tests.log)
[sh.helm.release.v1.default-istiod.v1.secret.txt](https://github.com/user-attachments/files/20793448/sh.helm.release.v1.default-istiod.v1.secret.txt)

### Operator Version

main

PR #971, opened 2025-07-02
Try to uninstall releases in `uninstalling` status

mkolesnik:
When installing with helm, sometimes the intallation fails and the operator issues an uninstall which can also fail.
In such a case, the release will get stuck perpetually in `uninstalling` status.

It should be safe to retry uninstalling it in such a case.

Fixes: https://github.com/istio-ecosystem/sail-operator/issues/920

### DIFF
--- a/pkg/helm/chartmanager.go
+++ b/pkg/helm/chartmanager.go
@@ -94,8 +94,10 @@ func (h *ChartManager) UpgradeOrInstallChart(
 			return nil, fmt.Errorf("failed to roll back helm release %s: %w", releaseName, err)
 		}
 		releaseExists = true
-	} else if rel.Info.Status == release.StatusPendingInstall || (rel.Info.Status == release.StatusFailed && rel.Version <= 1) {
-		log.V(2).Info("Performing helm uninstall", "release", releaseName)
+	} else if rel.Info.Status == release.StatusPendingInstall ||
+		rel.Info.Status == release.StatusUninstalling ||
+		(rel.Info.Status == release.StatusFailed && rel.Version <= 1) {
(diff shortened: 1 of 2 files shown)


mkolesnik:
> Are you sure this works? This [stackoverflow thread](https://stackoverflow.com/questions/74050419/helm-release-stuck-in-uninstalling-state) suggests that you'd need to rollback the release

I'm running the tests locally in a loop and this manages to recreate it sometimes (very rarely).
So far with the fix I see it did manage to delete the release this way.

If you think `helm rollback` would be better I don't mind trying that instead, though it seems from the discussion the "release" secret will still be left behind and you'll have to delete it manually either way.
I rather use this approach (seeing that it works so far) since helm should do the cleanup, we don't want to go around poking under the hood.

mkolesnik:
I ran this for days in a loop (testing in kind) and didn't get any problem due to the fix.
I do see it happening very rarely in the log, and with this fix the operator is able to recover and reinstall Istio properly.

Attaching a log to show the new behavior.
It can be found by looking for a line such as `Performing helm uninstall	{"IstioRevision": "default", "reconcileID": "d351213e-0a25-430b-b064-88e5d62be86a", "release": "default-istiod", "status": "uninstalling"}`

[sail-operator-6f6b8677d6-h2whk_sail-operator_sail-operator-3d6cadafb8d8b790301bb6aa54d060538695dcfbf2aa8968d21eb141bdf7bcc8.log](https://github.com/user-attachments/files/21087308/sail-operator-6f6b8677d6-h2whk_sail-operator_sail-operator-3d6cadafb8d8b790301bb6aa54d060538695dcfbf2aa8968d21eb141bdf7bcc8.log)

dgn (approved):
Thank you @mkolesnik, nice work

PR #990, opened 2025-07-07
Always use an image from values when generating bundle

FilipB:
Using values.yaml as a source of truth for bundle generation.

 This allows to easily use `${OSSM_OPERATOR_3_1}` string in https://github.com/openshift-service-mesh/sail-operator/pull/382


mkolesnik:
I guess the original intention of this variable was for local testing with kind?

Though it doesn't seem to use it - when running with `OLM=true` it modifies the `values.yaml` directly though IMHO that's a bad approach since it leaves residual changes which have to be manually undone.

PR #1009, opened 2025-07-10
watch EndpointSlices instead of Endpoints

### LINKED ISSUE #811: Watch EndpointSlice instead of Endpoint
The [`Endpoint` API has been deprecated in k8s 1.33+](https://kubernetes.io/blog/2025/04/24/endpoints-deprecation/), so we need to migrate to using `EndpointSlice`.

skriss:
#### What type of PR is this?

- [ ] Bug Fix
- [x] Refactor
- [ ] Optimization
- [ ] Test
- [ ] Documentation Update

Switches to watching EndpointSlices instead of Endpoints due to [Endpoints deprecation in k8s 1.33](https://kubernetes.io/blog/2025/04/24/endpoints-deprecation/).

Fixes #811

#### Additional information:
Note, the istiod chart still creates an Endpoints resource for remote installs (ref. https://github.com/istio/istio/blob/master/manifests/charts/istio-control/istio-discovery/templates/remote-istiod-endpoints.yaml) so need to use `+lint-watches:ignore`  for now. I can look at making an upstream PR to switch to EndpointSlices there as well.

I did not see any codegen for updating the Role definition from the kubebuilder markers, looks like it is updated manually, is that correct?

Is there anything else I'm missing for this change?


bundle/manifests/sailoperator.clusterserviceversion.yaml
```diff
@@ -472,6 +472,12 @@ spec:
                 - horizontalpodautoscalers
               verbs:
                 - '*'
+            - apiGroups:
+                - discovery.k8s.io
+              resources:
+                - endpointslices
+              verbs:
+                - '*'
```
mkolesnik, line 480: Don't we want to limit this to just get, list and watch?

chart/templates/rbac/role.yaml
```diff
@@ -40,6 +40,12 @@ rules:
   - horizontalpodautoscalers
   verbs:
   - '*'
+- apiGroups:
+    - discovery.k8s.io
+  resources:
+    - endpointslices
+  verbs:
+    - '*'
```
mkolesnik, line 48: Same here

bundle/manifests/sailoperator.clusterserviceversion.yaml
skriss: Good call, that will be sufficient for now

skriss:
Looks like it won't be quite this simple, unfortunately the mirrored EndpointSlice doesn't get the same IstioRevision owner ref as the Endpoints resource, instead the EndpointSlice is owned by the Endpoint, so the IstioRevision reconciler won't be triggered. Need to look into this some more.

skriss:
Latest commit additionally supports the case where the ownership chain is EndpointSlice -> Endpoint -> IstioRevision (i.e. what is happening currently). This should both support the current situation, and future-proof things for if/when the istiod Helm chart starts creating EndpointSlices directly.

mkolesnik (approved):
LGTM

PR #1035, opened 2025-07-17
README.md: add some badges for our github actions

dgn:
Our update-deps job was broken for a while - maybe this will increase visibility.


README.md
```diff
@@ -1,3 +1,8 @@
+![unit test badge](https://github.com/istio-ecosystem/sail-operator/actions/workflows/unit-tests.yaml/badge.svg)
```
fjglira, line 1: My bad here, this is static right now. We need to add the link to the action that we are going to monitor. For example, something like this:
```
[![unit-test](https://github.com/istio-ecosystem/sail-operator/actions/workflows/unit-tests.yaml/badge.svg)](https://github.com/istio-ecosystem/sail-operator/blob/main/.github/workflows/unit-tests.yaml)
```

I'll create new PR to fix this @dgn

PR #1048, opened 2025-07-22
Simplify samples files

### LINKED ISSUE #696: Automate copy and build of test images to our repo
We need to automate the build and push to our repo for all the testing images that are going to be used in the testing stages

fjglira:
#### What type of PR is this?

- [ ] Bug Fix
- [ ] Refactor
- [x] Optimization
- [ ] Test
- [ ] Documentation Update

This PR removes the local copies of Istio sample YAMLs to eliminate duplication with the `openshift-service-mesh/istio` repository.

To avoid CI rate-limiting, we currently modify these samples to use images from quay.io/sail-dev in both repositories. Centralizing these files in the midstream repo streamlines maintenance.

The test will now fetch samples from the midstream repo directly by using `SAMPLE_YAML_BASE_URL` logic that already exists for the test. An automated job will be added to the openshift-service-mesh repo to keep the files and image tags updated from upstream.

Fixes #696

Related Issue/PR #


mkolesnik:
If anything, shouldn't this be the other way around?

The upstream should be source of truth, and the midstream building upon it with the vendor specific stuff.

This way it flips the order, which can be quire confusing for any contributor outside RH

fjglira:
> If anything, shouldn't this be the other way around?
> 
> The upstream should be source of truth, and the midstream building upon it with the vendor specific stuff.
> 
> This way it flips the order, which can be quire confusing for any contributor outside RH

The thing is that we already do it in this way, but in this repo. Currently, we have a sample folder that we need to maintain, which is a copy of upstream sample files, but we are changing the image to target the quay.io repository. This means that currently we do not use upstream sample files by default, and then in the openshift-service-mesh/istio repo, we have the same sample files, also with the change of the image to target quay.io.

I know that this can be confusing, but the other solution that I see is to pull the images from upstream and replace them with images from Quay. But this will require parsing the yaml file to override the image field. If we do something like this, we will ensure that we use the upstream samples

Note: We changed the image to target quay.io because we were facing ratelimit issues pulling from upstream images while running the test

PR #1053, opened 2025-07-23
[release-1.26] Allow to use --use-image-digests flag when building a bundle

FilipB:
Operator SDK is able to automatically replace all image pullspecs with tags in the base CSV by digests.
Digests are required in disconnected OCP environments.


FilipB:
Do not merge before https://github.com/openshift-service-mesh/sail-operator/pull/414

PR #1065, opened 2025-07-29
Fix CNI dependency logic for OpenShift platform

sridhargaddam:
This PR fixes an issue with the Istio resource `Status` field.

When we install an Istio CR with `values.pilot.Cni.Enabled` set to `true`, the Istio revision controller expects the Istio CNI pods to be deployed and running, regardless of the platform (including `openshift`).
On the other hand, if `values.pilot.Cni.Enabled` is set to `false` (or left unset), the controller checks the platform and behaves accordingly:
1. If the platform is `openshift`, it expects the IstioCNI pods to be deployed and running.
2. For all other platforms, the controller ignores IstioCNI as part of its dependency checks.

### DIFF
--- a/tests/integration/api/cni_dependency_test.go
+++ b/tests/integration/api/cni_dependency_test.go
@@ -126,7 +126,7 @@ var _ = Describe("CNI Dependency", Ordered, func() {
 
 			AfterAll(func() {
 				Step("Cleaning up Istio resource")
-				Expect(k8sClient.Delete(ctx, istio)).To(Succeed())
+				deleteAllIstiosAndRevisions(ctx)
 			})
 		})
 
@@ -140,12 +140,8 @@ var _ = Describe("CNI Dependency", Ordered, func() {
 					Spec: v1.IstioSpec{
 						Version:   istioversion.Default,
 						Namespace: istioNamespace,
-						Values: &v1.Values{
-							Pilot: &v1.PilotConfig{
-								Cni: &v1.CNIUsageConfig{
-									Enabled: ptr.Of(false),
-								},
-							},
+						Values:    &v1.Values{
+							// not setting anything as CNI should be disabled by default on non-OCP platforms
 						},
 					},
 				}
@@ -171,7 +167,7 @@ var _ = Describe("CNI Dependency", Ordered, func() {
 
 			AfterAll(func() {
 				Step("Cleaning up Istio resource")
-				Expect(k8sClient.Delete(ctx, istio)).To(Succeed())
+				deleteAllIstiosAndRevisions(ctx)
 			})
 		})
 	})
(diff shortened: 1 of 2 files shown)


sridhargaddam:
With the fix when values.pilot.cni.enabled is explicitly set to false.
```
$: k get istios
NAME                NAMESPACE           PROFILE   REVISIONS   READY   IN USE   ACTIVE REVISION     STATUS    VERSION   AGE
openshift-gateway   openshift-ingress             1           1       0        openshift-gateway   Healthy   v1.24.4   38s
$: 

```

When values.pilot.cni.enabled is either not specified or explicitly set to true.
```
$: k get istios
NAME                NAMESPACE           PROFILE   REVISIONS   READY   IN USE   ACTIVE REVISION     STATUS             VERSION   AGE
openshift-gateway   openshift-ingress             1           1       0        openshift-gateway   IstioCNINotFound   v1.24.4   2m4s
$:
```

controllers/istiorevision/istiorevision_controller.go
```diff
@@ -64,17 +65,19 @@ const istioCniName = "default"
 // Reconciler reconciles an IstioRevision object
 type Reconciler struct {
 	client.Client
-	Config       config.ReconcilerConfig
-	Scheme       *runtime.Scheme
-	ChartManager *helm.ChartManager
+	Config             config.ReconcilerConfig
+	Scheme             *runtime.Scheme
+	ChartManager       *helm.ChartManager
+	cniDependencyCache sync.Map
```
mkolesnik, line 71: Why is a cache needed here?

sridhargaddam: Since the `DependsOnIstioCNI` method is invoked as part of `Reconcile` method, I want to avoid invoking the `ComputeValues` method unless there is any change in the IstioRevision.

mkolesnik: OK but does `ComputeValues` take a long time or access any resources on the cluster?

AFAICT it just computes stuff in memory, which should be relatively fast..

The reason I ask is because caches introduce complexity, so unless they save us substantial time/resources then I think we best avoid them.

Also if you believe it's necessary I'd document the reason why it's needed.

sridhargaddam: Right, on further testing, I found one edge case with Cache when users create, delete and re-creates the same resource. Let me remove the cache.

pkg/revision/dependency.go
```diff
@@ -14,17 +14,33 @@
...
 
 // DependsOnIstioCNI returns true if CNI is enabled in the revision
-func DependsOnIstioCNI(rev *v1.IstioRevision) bool {
-	// TODO: get actual final values and inspect pilot.cni.enabled
-	values := rev.Spec.Values
-	if values == nil {
+func DependsOnIstioCNI(rev *v1.IstioRevision, cfg config.ReconcilerConfig) bool {
+	values, err := defaultComputeValues(rev.Spec.Values, rev.Spec.Namespace, rev.Spec.Version,
+		cfg.Platform, cfg.DefaultProfile, "", cfg.ResourceDirectory, rev.Name)
+	if err != nil || values == nil {
```
mkolesnik, line 30: Is it a good idea to squash the error here?

It might be masking a bug..

I'd at least log this in case of error, if not cancel reconciliation attempt altogether.

WDYT?

sridhargaddam: As this PR is fixing a blocker bug. Is it okay if I handle it in a follow up PR?

mkolesnik: @sridhargaddam I was going to say "sure" but I guess I have no choice in the matter..

Anyway might be a good idea to do something if it errors out, unless you think it makes sense to squash it (in which case maybe its good to add a comment why it's squashed).

PR #1069, opened 2025-07-30
Add cleaner to operator install test

mkralik3:
Add a cleaner to clean up resources (`clusterrolebindings metrics-reader-rolebinding` and `job curl-metrics`) which are created in the operator install test. (to be able to run the sail e2e test more times in a row on the same cluster)


fjglira (approved):
lgtm, this will avoid dangling resources while reusing clusters to run test

Issue #1089, opened 2025-08-01
[BUG] Istio CNI fails to upgrade revision


libesz:
### Is this the right place to submit this?

- [x] This is not a question about how to use the sail-operator

### Bug Description

Sometimes bumping `IstioCNI` revision fails. The issue appears to be similar to https://github.com/istio-ecosystem/sail-operator/issues/920, but with CNI.

Resource dump:
```
$ k get istiocnis.sailoperator.io -A -o yaml                                                                                                                                                                                
apiVersion: v1
items:
- apiVersion: sailoperator.io/v1
  kind: IstioCNI
  metadata:
    annotations:
[...]
    creationTimestamp: "2025-03-27T07:17:30Z"
    finalizers:
    - sailoperator.io/sail-operator
    generation: 3
    name: default
    resourceVersion: "421312591"
    uid: cfdfbed9-3856-4e54-b0fe-9f58d383fd9a
  spec:
    namespace: istio-operator
    version: v1.24.6
  status:
    conditions:
    - lastTransitionTime: "2025-05-15T04:05:44Z"
      message: 'error reconciling resource: failed to install/update Helm chart "cni":
        unrecoverable helm release status pending-rollback'
      reason: ReconcileError
      status: "False"
      type: Reconciled
    - lastTransitionTime: "2025-08-01T05:48:08Z"
      status: "True"
      type: Ready
    observedGeneration: 3
    state: ReconcileError
kind: List
metadata:
  resourceVersion: ""
```

Operator log:
```
2025-08-01T11:18:06Z	INFO	ctrlr.istio	Reconciling	{"Istio": "default", "reconcileID": "492ec0b2-4b32-4002-957e-1f9a1027c5b0"}
2025-08-01T11:18:06Z	INFO	ctrlr.istio	Updating IstioRevision	{"Istio": "default", "reconcileID": "492ec0b2-4b32-4002-957e-1f9a1027c5b0", "IstioRevision": "default-v1-24-6"}
2025-08-01T11:18:06Z	INFO	ctrlr.istio	Reconciliation done. Updating status.	{"Istio": "default", "reconcileID": "492ec0b2-4b32-4002-957e-1f9a1027c5b0"}
2025-08-01T11:18:06Z	INFO	ctrlr.istio	Reconciling	{"Istio": "default", "reconcileID": "c5e89a41-4461-4425-8ae8-3a0288c063ea"}
[...]

PR #1104, opened 2025-08-06
Increase mem limit of sail operator.

libesz:
With 512Mi it often gets OOMKilled during chart upgrades, leading to other issues.

 

#### What type of PR is this?

- [x] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [ ] Test
- [ ] Documentation Update

Attempting to fix the rootcause of this: https://github.com/istio-ecosystem/sail-operator/issues/1089

Fixes #

Related Issue/PR #

### DIFF
--- a/bundle/manifests/sailoperator.clusterserviceversion.yaml
+++ b/bundle/manifests/sailoperator.clusterserviceversion.yaml
@@ -34,7 +34,7 @@ metadata:
     capabilities: Seamless Upgrades
     categories: OpenShift Optional, Integration & Delivery, Networking, Security
     containerImage: quay.io/sail-dev/sail-operator:1.27-latest
-    createdAt: "2025-08-06T05:12:00Z"
+    createdAt: "2025-08-06T12:36:35Z"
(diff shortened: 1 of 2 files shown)


libesz:
@fjglira thanks for merging! could you please confirm this change will be effective in OSSM 3.0 as well, with its next release?

mkralik3:
@libesz it will not. 
If you want to have this in the next OSSM 3.0 and 3.1 patch releases, first, the change must be backported/cherry-pick to the `release-1.26`(this is sync with OSSM 3.1 midstream) and `release-1.0`(this is sync with OSSM 3.0 midstream) upstream branches

mkralik3:
> @mkralik3 thanks! Should I do that backporting? Asking because earlier my changes got picked without me doing anything.

no need, when I set cherrypick/release-1.0 and cherrypick/release-1.26 labels, the automation takes care of that. 

However, as @FilipB mentioned, this change is slightly different since, in midstream, we have a separate file for these values (which I had forgotten about). 
So the additional change in midstream is needed even though we add a change to `release-1.0` and `release-1.26` upstream branches.

PR #1107, opened 2025-08-07
Adding support for configuring the max concurrent reconciles

### LINKED ISSUE #991: [FEATURE] Investigate increasing `MaxConcurrentReconciles`

### Feature Description

To enable the operator to handle a larger volume of reconciles, we should consider exposing `MaxConcurrentReconciles` as a config option. Bumping this to greater than 1 may have implications for how the reconcile loops are currently implemented so it needs to be tested first. For more context, see [this thread](https://github.com/istio-ecosystem/sail-operator/pull/802#discussion_r2190091764). 

### Additional Information

_No response_

Dimss:
- [ ] Bug Fix
- [ ] Refactor
- [x] Optimization
- [ ] Test
- [ ] Documentation Update

Fixes #991

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -56,6 +56,8 @@ func main() {
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.StringVar(&configFile, "config-file", "/etc/sail-operator/config.properties", "Location of the config file, propagated by k8s downward APIs")
 	flag.StringVar(&reconcilerCfg.ResourceDirectory, "resource-directory", "/var/lib/sail-operator/resources", "Where to find resources (e.g. charts)")
+	flag.IntVar(&reconcilerCfg.MaxConcurrentReconciles, "max-concurrent-reconciles", 5,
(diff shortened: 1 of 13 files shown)


Dimss:
@nrfox @dgn 
I didn't find any evidence that increasing the max concurrent reconciles can lead to errors or endless reconciles loops.
The base chart indeed reconciled in two different controllers, which perhaps not the best way to handle it, however, this should not affect the concurrent reconciles either.
The `IstioRevision` controller watch for `IstioRevisionTag` , for updating the `InUse` filed, but this logic also seems to me working as well with concurrent reconciles.
Local tests didn't showed any unusual, the CI tests are seems to be ok either.
For all the unit tests, I statically configured the max concurrent reconciles to `5` not sure if this should be managed somehow different or we are ok to keep it as `5` for unit tests. 
Please review the PR and let's discuss how to proceed.

nrfox (commented):
This looks good and thanks for investigating this. Can we update the value in the integration tests as well? https://github.com/istio-ecosystem/sail-operator/blob/d0c03f043ba96b2568585a03d4f9ea13f11700e4/tests/integration/api/suite_test.go#L87-L92

nrfox (approved):

PR #1156, opened 2025-08-21
Fix istiorevision int test for new PDB default behavior

FilipB:
After https://github.com/istio/istio/pull/57202 the PDB is not created by default. This change is setting AutoscaleMin to 2 so the PDB is created and the tests is passing

### DIFF
--- a/tests/integration/api/istiorevision_test.go
+++ b/tests/integration/api/istiorevision_test.go
@@ -406,7 +406,8 @@ var _ = Describe("IstioRevision resource", Label("istiorevision"), Ordered, func
 						},
 						Revision: ptr.Of(revName),
 						Pilot: &v1.PilotConfig{
-							Image: ptr.Of(pilotImage),
+							Image:        ptr.Of(pilotImage),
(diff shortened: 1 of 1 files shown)


sridhargaddam (approved):

FilipB:
@dgn could you please review? I'm not sure if it's ok to watch for the PDB if it does not exist [here](https://github.com/istio-ecosystem/sail-operator/blob/main/controllers/istiorevision/istiorevision_controller.go#L304).

MaxBab:
> lint is not happy with `_internal_defaults_do_not_set` in chart values. Not related to this change.

No, it's not related.
This error pops up in every PR now.

The problem is that I'm unable to reproduce it locally.

FilipB:
> > lint is not happy with `_internal_defaults_do_not_set` in chart values. Not related to this change.
> 
> No, it's not related. This error pops up in every PR now.
> 
> The problem is that I'm unable to reproduce it locally.

It's happening only in the new gcr.io/istio-testing/build-tools:master-84ec1491b45cfdc32689a5f875ffcdd7ae0ba260

PR #1248, opened 2025-10-01
Adding CI env var to e2e test workflow

fjglira:
#### What type of PR is this?

- [ ] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [X] Test
- [ ] Documentation Update

This variable controls the creation of the internal image registry route for OCP cluster while running the test. This means that we need to set the creation of the route during a separated ci steps, this will add stability and control for the test workflow. Also the env var can be used to set others CI specific features later on

Fixes #

Related Issue/PR #

### DIFF
--- a/tests/e2e/setup/build-and-push-operator.sh
+++ b/tests/e2e/setup/build-and-push-operator.sh
@@ -36,11 +36,11 @@ get_internal_registry() {
       "until ${COMMAND} get route default-route -n openshift-image-registry &>/dev/null; do sleep 5; done"
   fi
 
-  setup_ci_environment
+  setup_environment
 }
 
-setup_ci_environment() {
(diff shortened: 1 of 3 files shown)


tests/e2e/setup/build-and-push-operator.sh
```diff
@@ -86,7 +97,13 @@ build_and_push_operator_image() {
 
 # Main logic
 if [ "${OCP}" == "true" ]; then
-  get_internal_registry
+  if [ "${CI}" == "false" ]; then
```
FilipB, line 100: I'm not sure if CI is a good name, here it only controls if we are going to skip internal registry config which must happen later anyway no matter if it's CI run or localhost run.

fjglira:
Not needed anymore, it will be handled during the CI

Issue #1045, opened 2025-07-22
[EPIC] Manifest Customization


dgn:
### Epic Description

We should allow users to customize how the manifests are applied, ideally through a separate CRD, so that it can be access-controlled through Kubernetes RBAC.

Acceptance Criteria:
* Users can define specific fields that should never be overwritten by the operator (e.g. "don't overwrite changes to my ServiceAccount's pullSecret field")
* Users can override specific fields on their resources (e.g. "add the label 'managed-by: platform-team' to all resources created by the operator) even if the helm chart define a different value
* Users can target specific resources, resource kinds or use wildcards to apply these rules (e.g. "add label to all Deployments", "set spec.type to NodePort on the istiod service")
* We should be able to use the same logic to define internal defaults, e.g. "never update the failurePolicy on a webhook" as those are managed by istiod. This doesn't necessarily have to happen through a CRD but maybe just an internal representation (manifest customization defaults)

PR #1256, opened 2025-10-06
Skip reconciliation when using sailoperator.io/ignore annotation

### LINKED ISSUE #1242: [FEATURE] Pause reconciliation

### Feature Description

Please add support for pausing reconciliation on the selected resources. This would help with issues like #1148 as the only way currently to stop recurring update every few sec for managed resource (Istio/IstioRevision/IstioCNI) is to scale down the operator to 0.

### Additional Information

For the reference, similar functionality exists in:
- prometheus operator as part of resource [CRD](https://prometheus-operator.dev/docs/api-reference/api/?#monitoring.coreos.com/v1.Prometheus) - `paused`
- keda as part of resource [annotation](https://keda.sh/docs/2.14/concepts/scaling-jobs/#pause-autoscaling) - `autoscaling.keda.sh/paused: true`

bmangoen:
#### What type of PR is this?

- [x] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [ ] Test
- [ ] Documentation Update

With this fix, setting `sailoperator.io/ignore: "true"` on a resource will skip its reconciliation. 
This is a temporary hack until we implement the [Manifest Customization feature](https://github.com/istio-ecosystem/sail-operator/issues/1045)

Fixes #1242

### DIFF
--- a/pkg/predicate/predicate_test.go
+++ b/pkg/predicate/predicate_test.go
@@ -33,6 +33,38 @@ func TestIgnoreUpdate(t *testing.T) {
 
 func TestIgnoreUpdateWhenAnnotation(t *testing.T) {
 	predicate := IgnoreUpdateWhenAnnotation()
+	// Object does not contain sailoperator.io/ignore annotation
+	// so reconciliation should be done and both objects should be equal
+	assert.Equal(t, true, predicate.Update(event.UpdateEvent{
+		ObjectOld: &corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{},
+		},
+		ObjectNew: &corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{},
+			Data: map[string]string{
+				"foo": "bar",
+			},
+		},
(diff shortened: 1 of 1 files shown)


controllers/istiorevision/istiorevision_controller.go
```diff
@@ -289,40 +289,50 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Watches(&v1.IstioRevision{}, mainObjectHandler).
 		Named("istiorevision").
 
+		// We use predicate.IgnoreUpdateWhenAnnotation() so that we skip the reconciliation
+		// when the sailoperator.io/ignore annotation is set to "true" on the resource.
+		// TODO: this is a temporary hack until we implement the correct solution to ignore specific fields in the resource.
```
dgn, line 294: I don't know if it's a temporary hack- we can keep the annotation around IMO

dgn (commented):
lgtm so far, but an integration test would be good

PR #1297, opened 2025-10-22
Make sure a correct helm values file is used

FilipB:
It's possible to define a helm values file via HELM_VALUES_FILE env variable so we have to obey that when patching the file. We also need an option to disable the patching for ossm.


dgn:
Have you tested this against the nightly build? I'm not sure but I think that's where the `--version` flag originated

FilipB:
> Have you tested this against the nightly build? I'm not sure but I think that's where the `--version` flag originated

Thank you, I missed that. It's there really because of nightly builds. I put it back.

PR #1334, opened 2025-11-07
Enhance IstioRevision dependency check to include ztunnel status

sridhargaddam:
The IstioRevision controller was missing health validation for ZTunnel while using the ambient mode. Because of this, even when ztunnel was missing/unhealthy, the IstioRevision was showing the status as Ready. This PR fixes it.


dgn (approved):
Thank you @sridhargaddam

PR #1353, opened 2025-11-17
Add Ambient multi cluster E2E tests for Multi Primary deployments

mkolesnik:
#### What type of PR is this?

- [ ] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [x] Test
- [ ] Documentation Update

Istio 1.27+ supports running Multi-Primary deployments in Abmient mode, so tests were adapted to run both on Sidecar and Ambient mode.
Both modes run the same tests, sequentialy & independently of each other.

Refactors were made in the multi cluster tests to consolidate and reuse code, as part of the work to add the Ambient mode support.

Fixes #

Related Issue/PR #

### DIFF
--- a/tests/e2e/setup/setup-kind.sh
+++ b/tests/e2e/setup/setup-kind.sh
@@ -18,6 +18,7 @@ set -eux -o pipefail
 
 SCRIPTPATH="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 ROOT=$(cd "${SCRIPTPATH}/../../.." && pwd)
+GW_API_VERSION=${GW_API_VERSION:-v1.4.0}
 
 # shellcheck source=common/scripts/kind_provisioner.sh
 source "${ROOT}/common/scripts/kind_provisioner.sh"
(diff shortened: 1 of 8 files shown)


tests/e2e/setup/setup-kind.sh
```diff
@@ -70,6 +70,11 @@ if [ "${MULTICLUSTER}" == "true" ]; then
     setup_kind_clusters "${KIND_IMAGE}" ""
     setup_kind_registry "${CLUSTER_NAMES[@]}"
 
+    # Apply Gateway API CRDs which are needed for Multi-Cluster Ambient testing
+    for config in "${KUBECONFIGS[@]}"; do
+        kubectl apply --kubeconfig="$config" --server-side -f https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.4.0/experimental-install.yaml
```
fjglira: We should set a way to keep this up to date instead of hardcoding a specific version

fjglira (commented):
lgtm, just two small comments

PR #1365, opened 2025-11-18
api: add ZTunnel v1 CRD version

dgn:
This adds the v1 CRD version of ZTunnel which we missed earlier. The only change is the removal of the `profile` field which had only a valid value of `ambient` which was set by default. I moved setting the profile to `ambient` into the controller code instead.


sridhargaddam (commented):
Thanks @dgn for working on it.

mkolesnik (approved):
LGTM

mkolesnik:
> > @dgn do we need a conversion webhook to handle automatic conversion between `v1alpha1` and `v1` versions?
> 
> nope, not at this time, as they're identical. and as we're changing the `storedversion` to v1, we shouldn't have to in the future, either

Since you removed `profile` does this still make sense or is a conversion needed to remove that field?

PR #1424, opened 2025-12-05
Running automator to update deps also for release branches

FilipB:
Scheduled workflows run on the latest commit on the default branch so we need to run it for all branches here.

Related Issue/PR #1421


FilipB:
Do not merge until the changes in update_deps.sh are propagated to all branches.

PR #1474, opened 2025-12-19
[claude] add slash command for pre-commit steps

cam-garrison:
Adding a simple helper slash command for use with Claude, `/pre-commit` to run pre-commit steps before adding changes to the project. Runs linter, unit tests, gen-check. Gives helpful output if individual steps fail.

### DIFF
--- a/.claude/commands/pre-commit.md
+++ b/.claude/commands/pre-commit.md
@@ -29,7 +29,17 @@ This runs all linters including:
 - Secret scanning (gitleaks)
 - Spell checking
 
-If linting fails, show the errors and stop. Provide guidance on fixing common lint issues.
+**If linting fails:**
+
+1. First, analyze the lint errors to understand what needs fixing
+2. Attempt to fix the issues:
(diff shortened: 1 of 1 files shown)


.claude/commands/pre-commit.md
```diff
@@ -0,0 +1,118 @@
...
+This runs all linters including:
+- Go linting (golangci-lint)
+- YAML linting
+- Helm chart linting
+- Script linting
+- License and copyright checks
+- Secret scanning (gitleaks)
+- Spell checking
+
+If linting fails, show the errors and stop. Provide guidance on fixing common lint issues.
```
bartoszmajsak: Or maybe try to fix it first? :) I find it quite effective to fix linter errors when golangci-lint does not provide autofix issue

fjglira: +1

cam-garrison: f0715f6 good idea!

PR #1547, opened 2026-01-27
Enable TLSv1.2 for ZTunnel when in FIPS mode

dgn:
This change builds on https://github.com/istio/ztunnel/pull/1711 which adds TLSv1.2 support to ZTunnel when `TLS12_ENABLED` is set to `true`. This patch will always set the env var when in FIPS mode, for all versions of ZTunnel, even though it is only supported from 1.29+, but the env var will simply be ignored by versions that don't support it.

### DIFF
--- a/pkg/istiovalues/fips_test.go
+++ b/pkg/istiovalues/fips_test.go
@@ -87,6 +87,8 @@ func TestApplyFipsValues(t *testing.T) {
 	values := helm.Values{}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
+			originalFipsEnabled := FipsEnabled
+			t.Cleanup(func() { FipsEnabled = originalFipsEnabled })
 			FipsEnabled = tt.fipsEnabled
 			actual, err := ApplyFipsValues(values)
 			if (err != nil) != tt.expectErr {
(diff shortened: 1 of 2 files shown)


controllers/ztunnel/ztunnel_controller.go
```diff
@@ -157,6 +157,12 @@ func (r *Reconciler) installHelmChart(ctx context.Context, ztunnel *v1.ZTunnel)
 		return fmt.Errorf("failed to apply profile: %w", err)
 	}
 
+	// apply fips values
+	mergedHelmValues, err = istiovalues.ApplyZTunnelFipsValues(mergedHelmValues)
+	if err != nil {
+		return fmt.Errorf("failed to apply profile: %w", err)
```
sridhargaddam:
```suggestion
		return fmt.Errorf("failed to apply FIPS values: %w", err)
```

pkg/istiovalues/fips.go
```diff
@@ -53,3 +53,13 @@ func ApplyFipsValues(values helm.Values) (helm.Values, error) {
 	}
 	return values, nil
 }
+
+// ApplyFipsValues sets value pilot.env.COMPLIANCE_POLICY if FIPS mode is enabled in the system.
```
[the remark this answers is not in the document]
dgn: fixed!

pkg/istiovalues/fips_test.go
```diff
@@ -101,3 +101,44 @@ func TestApplyFipsValues(t *testing.T) {
...
+					"env": map[string]any{"TLS12_ENABLED": string("true")},
+				},
+			},
+		},
+	}
+
+	values := helm.Values{}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			FipsEnabled = tt.fipsEnabled
```
sridhargaddam: `FipsEnabled` seems to be a global variable and we are mutating its state. I think we should restore its state to the original value after calling ApplyZTunnelFipsValues.

sridhargaddam (approved):

PR #1706, opened 2026-03-17
refactor: move watches and predicates into separate pkg

dgn:
this will help trim down the pkg/install library by making the watches easily accessible from outside the controller pkgs.

cc @aslakknutsen


dgn:
Closing this for now as we should address it as part of #1758

PR #1788, opened 2026-04-07
Add the possibility to override the test timeout

mkralik3:
Added DEFAULT_TEST_TIMEOUT to be able to override default 180 seconds timer across test suite
Added tests for `env.GetInt` 

Assisted-by: Cursor (claude-4.5-opus-high)


fjglira (approved):
lgtm

PR #1807, opened 2026-04-09
refactor error and status condition handling

### LINKED ISSUE #996: Unify validation error handling in controllers
In the IstioRevisionTag controller, the reconciler returns custom error types in order to properly write to the resource status as well as to preserve the original error, should there be one. We should use this pattern throughout all controllers.

We should aim to keep the `Is<ErrorType>` functions around as those are standard in k8s and well-readable.

dgn:
this removes a bunch of duplicated code and aligns and simplifies error and status condition handling across the controllers

Fixes #996


api/v1/condition.go
nrfox, line 1: If you use [metav1.Condition](https://pkg.go.dev/k8s.io/apimachinery@v0.35.3/pkg/apis/meta/v1#Condition) you can then use [all the helpers](https://pkg.go.dev/k8s.io/apimachinery@v0.35.3/pkg/api/meta) like `SetStatusCondition(...)`

pkg/reconciler/condition.go
```diff
@@ -0,0 +1,69 @@
...
+)
+
+// ErrorConditionMapping maps an error predicate to a condition reason.
+// Mappings are evaluated in order; the first match wins.
+type ErrorConditionMapping struct {
+	// Matches returns true if this mapping applies to the given error.
+	Matches func(err error) bool
+	// Reason is the condition reason to use when this error matches.
+	Reason v1.ConditionReason
+}
```
nrfox: This is a bit convoluted. Can we move the generic parts like `updateStatus(...)` into the `StandardReconciler` and keep the non-generic parts like setting specific reason/messages on the status inside the specific controllers?

nrfox:
@dgn does this PR address this issue now: https://github.com/istio-ecosystem/sail-operator/issues/996?

dgn:
> @dgn does this PR address this issue now: #996?

it's my stab at it, yes

api/v1/condition.go
dgn: yeah that would be nice - my only issue with that is that it would also pull in all the kubebuilder validation of that package into our CRDs which is technically not a valid API change (as it restricts what's allowed further) - but maybe for a status field we can make an exception

nrfox: With [CRD ratcheting](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation-ratcheting) changes to validation should no longer be breaking changes. Old CRs with newly invalid fields will continue to be reconciled and patched but if that old field (now invalid) is touched it must be updated inline with the new validation.

nrfox (approved):
This is a good optimization and my comments can be addressed in a followup.

pkg/reconciler/condition.go
dgn: I got rid of the mapping, it's a bit simpler now

PR #1829, opened 2026-04-15
Update kustomization files with registry.istio.io

sridhargaddam:
- [x] Bug Fix


mkralik3 (approved):
AFAIK, we can have also both (but probably not needed)

PR #1840, opened 2026-04-15
Fix license linter after Helm v3→v4 migration

fjglira:
## Problem

After the Helm v4 migration (#1644), `make update-common` syncs `common/config/license-lint.yml` from `istio/common-files`, which only has `helm.sh/helm/v3` in the allowlist. This caused the Automator update-deps PR (#1837) to drop the `helm.sh/helm/v4` entry, making the license linter fail with 23 unrecognized licenses:

```
ERROR: Some modules have unrecognized licenses:
  helm.sh/helm/v4: path '.../testdata/frobnitz/LICENSE'
  ...
```

Related #1837


yxun (approved):

nrfox (approved):
Looks like even though Istio now supports helm v4 it still uses helm/v3 internally

PR #1854, opened 2026-04-20
Add 1.30.0-alpha.2 charts

MaxBab:
Add 1.30.0-alpha.2 charts for early testing phase.


MaxBab:
The failure of the "gencheck" job waits for the following PR to be merged - https://github.com/istio-ecosystem/sail-operator/pull/1852

api/v1/istio_types.go
```diff
@@ -35,10 +35,10 @@ const (
...
+	// +kubebuilder:validation:Enum=v1.30-latest;v1.30.0-alpha.2;v1.29-latest;v1.29.2;v1.29.1;v1.29.0;v1.28-latest;v1.28.6;v1.28.5;v1.28.4;v1.28.3;v1.28.2;v1.28.1;v1.28.0;v1.27-latest;v1.27.9;v1.27.8;v1.27.7;v1.27.6;v1.27.5;v1.27.4;v1.27.3;v1.27.2;v1.27.1;v1.27.0;v1.26-latest;v1.26.8;v1.26.7;v1.26.6;v1.26.5;v1.26.4;v1.26.3;v1.26.2;v1.26.1;v1.26.0;v1.25-latest;v1.25.5;v1.25.4;v1.25.3;v1.25.2;v1.25.1;v1.24-latest;v1.24.6;v1.24.5;v1.24.4;v1.24.3;v1.24.2;v1.24.1;v1.24.0;v1.23-latest;v1.23.6;v1.23.5;v1.23.4;v1.23.3;v1.23.2;v1.22-latest;v1.22.8;v1.22.7;v1.22.6;v1.22.5;v1.21.6;master;v1.31-alpha.4cc49d2e
+	// +kubebuilder:default=v1.30.0-alpha.2
```
dgn, line 41: should we really default to an alpha release?

MaxBab: @dgn 
That is a default generation output for `make gen` after adding the new version (standard or alpha) to `pkg/istioversion/versions.yaml` file.

Would you like to change the gen flow for alpha/beta versions?

dgn (approved):

api/v1/istio_types.go
dgn: as this is only on main, I think it's fine for now, we can do it in a follow-up if needed

PR #1857, opened 2026-04-21
pkg/install library

### LINKED ISSUE #1758: Add Sail Operator library

### Feature Description

Incorporate the [Sail Operator library](https://github.com/openshift-service-mesh/sail-operator/pull/721) from `openshift-service-mesh/sail-operator` into `istio-ecosystem/sail-operator.

We can address some of the review comments raised during the review and possibly split out the functionality into separate parts like what is being done here: https://github.com/istio-ecosystem/sail-operator/pull/1706

### Additional Information

_No response_

dgn:
This is a rewrite of @aslak's install library effort (https://github.com/openshift-service-mesh/sail-operator/pull/721).
    
Notable improvements:
- watches are the same as for the regular controllers
- required rbac is derived from watches
- typed informers/clients everywhere
- reuses ComputeValues/Reconciler logic whereever possible
- image digests are no longer pulled from CSV YAML but are generated and stored in pkg/install/images.gen.go
- multiple e2e tests


pkg/install/informers.go
nrfox, line 1: Why is this and the `lifecyle` part of the library necessary? Can't library users either import `watches` and setup the watches on their own controller or else our library exposes a controller that takes as input `Opts` and runs `Apply` from it?

If the Istio CRDs may not exist on startup then we may be able to provide a "lazy informer" that handles this case. Would need to look at the controller-runtime primitives.

dgn: Effectively this and lifecycle.go are the controller. We could put it in its own sub pkg `controller` and have users pass in their library instance I guess. No strong preference on my part

nrfox: It doesn't matter to me where it goes but it would be nice to avoid re-implementing all the controller-runtime internals if possible.

dgn: good point. I have a bunch of possibilities for simplication that I didn't touch yet.. I also want to explore generating rbac.go

dgn:
~Going through the threads on https://github.com/openshift-service-mesh/sail-operator/pull/721, I will probably have to revert my change to use the CRDs from `bundle/` instead of `chart/`~

Update: fixed

pkg/config/config_test.go
```diff
@@ -28,6 +28,41 @@ var testImages = IstioImageConfig{
 	ZTunnelImage: "ztunnel-test",
 }
 
+func TestReadBytes(t *testing.T) {
+	testCases := []struct {
+		name    string
+		data    string
+		success bool
```
nrfox: This only makes assertions about whether the function returned an error or not. It doesn't test the `Config` is what is expected.

pkg/install/crds.go
```diff
@@ -0,0 +1,235 @@
...
+func mergeStringMap(base, overlay map[string]string) map[string]string {
+	merged := make(map[string]string, len(base)+len(overlay))
+	for k, v := range base {
+		merged[k] = v
+	}
+	for k, v := range overlay {
+		merged[k] = v
+	}
+	return merged
+}
```
nrfox: Use https://pkg.go.dev/maps#Copy

pkg/install/crds.go
```diff
@@ -0,0 +1,235 @@
...
+	for _, entry := range entries {
+		if entry.IsDir() || !strings.HasSuffix(entry.Name(), ".yaml") {
+			continue
+		}
+		data, err := fs.ReadFile(m.crdFS, entry.Name())
+		if err != nil {
+			return nil, fmt.Errorf("failed to read CRD file %s: %w", entry.Name(), err)
+		}
+		allManifests.Write(data)
+		allManifests.WriteString("\n---\n")
```
nrfox: I don't think we need to rewrite the documents into a string. Doing `fs.WalkDir(ms.crdFS)` and using something `k8s.io/apimachinery/pkg/util/yaml.NewYAMLOrJSONDecoder(...).Decode(&crd)` should get us out of having to deal with any of the YAML bits.

pkg/install/library.go
```diff
@@ -0,0 +1,254 @@
...
+
+// WithQPS sets the maximum sustained queries-per-second to the API server.
+func WithQPS(qps float32) LibraryOption {
+	return func(o *libraryOptions) { o.qps = qps }
+}
+
+// WithBurst sets the maximum burst of requests to the API server.
+func WithBurst(burst int) LibraryOption {
+	return func(o *libraryOptions) { o.burst = burst }
+}
```
nrfox, line 67: Are these currently used by cluster ingress? If not, do we need them?

pkg/install/library.go
```diff
@@ -0,0 +1,254 @@
...
+			status := l.Status()
+			if status.Error == nil && status.Installed {
+				return nil
+			}
+			if status.Error != nil {
+				return fmt.Errorf("reconciliation failed: %w", status.Error)
+			}
+		}
+	}
+}
```
nrfox: This won't shutdown if `Stop()` is called.

nrfox: This doesn't look like it's currently used by cluster ingress: https://github.com/openshift/cluster-ingress-operator/blob/961ac218364fb5f6de970ade1161ece3b8195aed/pkg/operator/controller/gatewayclass/controller.go#L267-L275

Can we leave this off for now?

nrfox (commented):
Mostly just some nits. Still working through the rest of it. Much smaller diff which is great.

pkg/install/library.go
dgn: I ran into issues when they were not properly configured, so I upped the defaults and added these helpers. We don't need to keep them necessarily as the defaults are different, but if CIO ever runs into ratelimiting issues they at least have knobs and don't need to wait for the next version

pkg/install/library.go
dgn: It should- Stop() cancels the context

dgn: good point. will remove

dgn: removing the func entirely so this is moot

pkg/config/config_test.go
dgn: ack. will add

pkg/install/images.gen.go
```diff
@@ -0,0 +1,96 @@
...
+// limitations under the License.
+
+// Code generated by hack/patch-values.sh. DO NOT EDIT.
+
+package install
+
+import "github.com/istio-ecosystem/sail-operator/pkg/config"
+
+func init() {
+	config.Config.ImageDigests = map[string]config.IstioImageConfig{
```
nrfox, line 22: Do we want the global config var to be changed as a side effect of importing this package?

pkg/install/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+| Method | Description |
+|---|---|
+| `Start(ctx)` | Starts the controller manager and drift-detection watches; returns notification channel |
+| `Apply(opts)` | Validates and sets desired state; enqueues reconciliation |
+| `Stop()` | Cancels the reconciliation loop and waits for the manager to shut down |
+| `Enqueue()` | Forces re-reconciliation without changing desired state |
+| `Status()` | Returns the latest reconciliation result |
+| `Uninstall(ctx, ns, rev)` | Performs Helm uninstall of istiod |
+| `WaitForDesiredState(ctx)` | Blocks until install succeeds or fails |
```
nrfox:
```suggestion
```

nrfox (commented):
LGTM but waiting for @aslakknutsen to review before approving

pkg/install/images.gen.go
dgn: That was the simplest way -- the install code expects it to be set in that global config var, as the operator loads it when starting up. I suspect we could also initiate it when calling `library.New()`

pkg/install/README.md
dgn: fixed!

pkg/install/images.gen.go
nrfox: Would it be a problem if another part of the operator imported `install`?

dgn: it would be, yeah. but that would be wrong anyway :slightly_smiling_face: the library should import operator code, not the other way around

nrfox (approved):
LGTM but putting a hold so @aslakknutsen can sign off as well

pkg/install/crds.go
```diff
@@ -0,0 +1,222 @@
...
+	v1 "github.com/istio-ecosystem/sail-operator/api/v1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	utilyaml "k8s.io/apimachinery/pkg/util/yaml"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	olmManagedLabel = "operators.coreos.com/managed-by"
```
gcs278: This doesn't work for me - OLM sets `olm.managed` - I don't see `operators.coreos.com/managed-by`. Is this a new OLM label? 

```suggestion
	olmManagedLabel = "olm.managed"
```

pkg/install/crds.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+
+	ownership := m.classifyCRD(existing)
+	if ownership == crdManagedByOLM {
+		if overwriteOLM == nil || !overwriteOLM(ctx, existing) {
+			info.Managed = false
+			info.Ready = m.isCRDReady(existing)
+			return info, nil
+		}
+	}
```
gcs278, line 100: Just an observation - the behavior now overwrites unrecognized CRDs (no OLM or library labels) CRDs whereas previously (downstream) it left those CRDs along and set `CRDUnknownManagement`.

If there's ever a case where someone installed Istio CRDs manually via third party, it will get overwritten by this. I believe OLM overwrites any CRD update anyways, so I suppose it's mirroring that existing behavior. That's probably what we want, but mentioning it here for visibilty.

pkg/install/crds.go
```diff
@@ -0,0 +1,222 @@
...
+
+	crd.SetResourceVersion(existing.ResourceVersion)
+	mergedLabels := maps.Clone(existing.Labels)
```
gcs278: what was the reasoning behind preserving some of the metadata like labels? This also means we keep around `olm.managed: true` even after sail library has added `app.kubernetes.io/managed-by: sail-library` to it.

pkg/install/reconciler.go
```diff
@@ -0,0 +1,242 @@
...
+func (r *libraryReconciler) Reconcile(ctx context.Context, _ ctrlreconcile.Request) (ctrlreconcile.Result, error) {
+	r.lib.mu.Lock()
+	opts := r.lib.desiredOpts
+	gen := r.lib.generation
+	r.lib.mu.Unlock()
+
+	if opts == nil {
+		return ctrlreconcile.Result{}, nil
+	}
+
```
gcs278, line 58: My test is stuck in a drift detection reconcile loop every 4 seconds.

I see the helm rev climbing non-stop, and the only differences I see on the helm metadata are `last_deployed` timestamp, `status`, and `version` number.

Should there be a condition to make sure the current helm generation installed successfully?

gcs278 (commented):
Nice work - I tested this by vendoring the library into CIO and running it on a cluster, and found a few things. Hope this helps.

nrfox:
Hey @gcs278 thanks for taking a look at this and kicking the tires. Since @dgn will be out for a bit I'll try and address your feedback so we can move this along. I'll probably need to close this PR and open a separate one with additional commits on top.

pkg/install/crds.go
nrfox: To my knowledge, this PR is a rewrite of the existing library but the behavior should remain the same as it is in downstream. I'll change this back to leaving unrecognized CRDs alone.

pkg/install/crds.go
nrfox: Same comment as above, wrt keeping the previous behavior. I'll update this as well.

pkg/install/reconciler.go
nrfox: I wasn't able to reproduce this but I did add a test to try and catch this state. [These](https://github.com/istio-ecosystem/sail-operator/blob/f8f03773cf36fb2fed301a8aeee111454a6cf825/tests/e2e/crdownership/crd_ownership_test.go#L364-L371) are the `Apply` `Options` it uses. Are those similar to what y'all are setting?

The library should treat repeated calls to `Apply` as a noop if the previous options == new options.

nrfox: @gcs278 would you mind trying this PR again? If the issue still persists then would you share the `Options` you are passing to `Apply` so I can more easily reproduce it?

gcs278:
@nrfox yes sorry I haven't gotten to it yet, but I'm going to try to kick the tires again soon.

pkg/install/reconciler.go
aslakknutsen: This is resolved in https://github.com/istio-ecosystem/sail-operator/pull/1857/changes/8137671b4549f9131f38c6a7fb975dffecec0936

aslakknutsen (approved):
Looks good and tested via a backport to Sail 1.30 to be used in CIO

PR #1874, opened 2026-04-23
test: Modify await_operator to dinamically get deployment name from csv when OLM is true

fjglira:
If the bundle name is different from the sail operator, and OLM is set to true in the test, we got failures because await_operator never finds a sail-operator deployment in the namespace

 

#### What type of PR is this?

- [ ] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [x] Test
- [ ] Documentation Update

Fixes #

Related Issue/PR #

### DIFF
--- a/tests/e2e/common-operator-integ-suite.sh
+++ b/tests/e2e/common-operator-integ-suite.sh
@@ -230,10 +230,11 @@ await_operator() {
   local name="${DEPLOYMENT_NAME}"
   if [ "${OLM}" == "true" ]; then
     local csv_name
-    csv_name=$("${COMMAND}" get csv -n "${NAMESPACE}" \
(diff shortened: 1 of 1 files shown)


FilipB (commented):
Maybe it would be easier to parse the ClusterServiceVersion file directly in the repo instead of getting it dynamically from the cluster? There might be other CSV which will break this approach

fjglira:
> Maybe it would be easier to parse the ClusterServiceVersion file directly in the repo instead of getting it dynamically from the cluster? There might be other CSV which will break this approach



> Maybe it would be easier to parse the ClusterServiceVersion file directly in the repo instead of getting it dynamically from the cluster? There might be other CSV which will break this approach

Did some change to get the name from the file in the bundle folder

FilipB (approved):

PR #1882, opened 2026-04-24
Fixing two problems with our hack/update-istio.sh script

FilipB:
1. After https://github.com/istio/istio/pull/59509 it is necessary to correctly parse the version
2. After https://github.com/istio-ecosystem/sail-operator/pull/1854 the script must handle when multiple pre-release versions are listed.

Manually fixed problematic version in versions.yaml and run make gen.


dgn (approved):
Thank you 🙏

PR #1891, opened 2026-04-27
Ensure base validator is created for default rev

nrfox:
Clears the `defaultRevision` field in case a user sets this but does not set it to `""` which will cause the webhook to not be created. The validator should be created when the default rev exists. Adds some integration tests for these cases.

 

#### What type of PR is this?

- [x] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [ ] Test
- [ ] Documentation Update

Fixes #

Related Issue/PR #

### DIFF
--- a/tests/integration/api/istio_test.go
+++ b/tests/integration/api/istio_test.go
@@ -32,6 +32,7 @@ import (
 	"helm.sh/helm/v4/pkg/release"
 	releasecommon "helm.sh/helm/v4/pkg/release/common"
 	releasev1 "helm.sh/helm/v4/pkg/release/v1"
+	admissionv1 "k8s.io/api/admissionregistration/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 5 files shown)


nrfox:
@jewertow would you mind taking a look at this to ensure this behavior is correct?

tests/integration/api/istio_test.go
```diff
@@ -654,6 +655,28 @@ var _ = Describe("Istio resource", Ordered, func() {
 			}).Should(Succeed())
 		})
 	})
+
+	Describe("ValidatingWebhookConfiguration", func() {
+		It("creates istiod-default-validator when the Istio is named 'default' with InPlace strategy", func() {
```
jewertow, line 661: Could you add test cases that verify that the default validating webhook is not created in the following scenarios?
1. InPlace + name "custom"
2. RevisionBased

tests/integration/api/istio_test.go
```diff
@@ -654,6 +655,63 @@ var _ = Describe("Istio resource", Ordered, func() {
...
+					Version:   istioversion.Default,
+					Namespace: istioNamespace,
+					UpdateStrategy: &v1.IstioUpdateStrategy{
+						Type: v1.UpdateStrategyTypeInPlace,
+					},
+				},
+			}
+			createIstioWithCleanup(ctx, istio)
+
+			Consistently(k8sClient.Get).WithArguments(ctx, validatorWebhookKey, &admissionv1.ValidatingWebhookConfiguration{}).Should(ReturnNotFoundError())
```
jewertow, line 694: Could we also check that `istiod-custom-validator` was created?

jewertow (changes requested):

jewertow (approved):

fjglira (approved):

PR #1897, opened 2026-04-29
fix: handle repos with different release branch naming in crd-schema-checker

fjglira:
We are getting failures in a fork after the changes were made to the linter on this [PR](http://github.com/istio-ecosystem/sail-operator/pull/1055)

## Issue
When not on a release branch, the script previously derived the comparison branch from `PREVIOUS_VERSION` (e.g. `release-1.29`). This broke in repos that use a different naming convention (e.g. forks of repos using `release-3.x`), and also meant the pinned value could go stale as new releases were cut.

This change aligns the non-release-branch case with the release-branch case, which already auto-detects the relevant branch via `sort -V`. When not on a release branch, the script now always uses the latest available `release-*` branch. A `PREVIOUS_BRANCH` env var is also added as an explicit override when needed.

## Behavior by scenario

  | Scenario | Before | After |
  |---|---|---|
  | Upstream, on `release-1.30` | auto-detects `release-1.29` ✓ | unchanged ✓ |
  | Upstream, on `main` | uses `release-1.29` (pinned, potentially stale) | auto-detects `release-1.30` (latest) ✓ |
  | Downstream, on `main` | fails — `release-1.29` not found | auto-detects `release-3.3.2` ✓ |
  | Any repo, no release branches | skipped | skipped ✓ |

Assisted by @claude

### DIFF
--- a/tools/crd-schema-checker.sh
+++ b/tools/crd-schema-checker.sh
@@ -107,8 +107,8 @@ else
     previous_branch=$(git branch -r | grep -E 'origin/release-[0-9]+(\.[0-9]+)*$' | \
         sed 's|.*origin/||' | sort -V | tail -1)
     if [[ -z "$previous_branch" ]]; then
-        echo "No release branches found. Skipping."
-        exit 0
+        echo "No release branches found. Please check branch naming or specify PREVIOUS_BRANCH explicitly."
+        exit 1
     fi
 fi
 
--- a/tools/crd-schema-checker.sh
+++ b/tools/crd-schema-checker.sh
@@ -18,6 +18,12 @@
 # Compares CRDs between the current release branch and the previous one to catch breaking changes.
 # Uses OpenShift's crd-schema-checker to detect issues like removed fields or stricter validation.
 # Run 'make crd-schema-checker' first to install the dependency, then 'make lint-crds' to check.
+#
+# Environment variables:
+#   PREVIOUS_BRANCH  Optional. Explicitly set the release branch to compare against.
+#                    If not set, the script auto-detects the appropriate branch:
+#                    - On a release branch: uses the previous release branch (via sort -V).
+#                    - On any other branch: uses the latest available release branch (via sort -V).
 
 set -euo pipefail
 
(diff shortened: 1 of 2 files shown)


tools/crd-schema-checker.sh
```diff
@@ -93,11 +93,17 @@ if [[ "$current_branch" =~ ^release-[0-9]+\.[0-9]+$ ]]; then
     previous_branch=$(git branch -r | grep -E 'origin/release-[0-9]+\.[0-9]+$' | 
         sed 's|.*origin/||' | sort -V | 
         awk -v target="$current_branch" '$0 == target { print prev; exit } { prev = $0 }')
-elif [[ -n "${PREVIOUS_VERSION:-}" ]]; then
```
FilipB, line 96: As PREVIOUS_VERSION is not used anymore anywhere in this script it probably does not make sense to pass this variable to this script in the Makefile.

tools/crd-schema-checker.sh
```diff
@@ -93,11 +93,17 @@ if [[ "$current_branch" =~ ^release-[0-9]+\.[0-9]+$ ]]; then
...
-    previous_branch="release-$(echo "${PREVIOUS_VERSION}" | cut -f1,2 -d'.')"
+elif [[ -n "${PREVIOUS_BRANCH:-}" ]]; then
```
FilipB, line 102: PREVIOUS_BRANCH is not described anywhere, maybe a short help would be good for this script.

FilipB (changes requested):
I don't understand why this was not visible earlier.

Edit: nevermind I do now

fjglira:
> I don't understand why this was not visible earlier.

Because the change was merged 3 weeks ago and we have been dealing with the sync error since then

tools/crd-schema-checker.sh
fjglira: Added a note about `PREVIOUS_BRANCH` use

tools/crd-schema-checker.sh
fjglira: Removed

tools/crd-schema-checker.sh
```diff
@@ -93,11 +99,17 @@ if [[ "$current_branch" =~ ^release-[0-9]+\.[0-9]+$ ]]; then
...
+    previous_branch="${PREVIOUS_BRANCH}"
 else
-    echo "Not on a release branch and PREVIOUS_VERSION not set. Skipping."
-    exit 0
+    # Not on a release branch: compare against the latest available release branch.
+    previous_branch=$(git branch -r | grep -E 'origin/release-[0-9]+(\.[0-9]+)*$' | \
+        sed 's|.*origin/||' | sort -V | tail -1)
+    if [[ -z "$previous_branch" ]]; then
+        echo "No release branches found. Skipping."
+        exit 0
```
FilipB: Maybe we should exit with error? If this should catch problems, we should not silently skip the test if the branch is not found

fjglira: Done

FilipB (approved):

PR #1902, opened 2026-05-03
Fix automator update deps sync job failure

MaxBab:
#### What type of PR is this?

- [X] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [ ] Test
- [ ] Documentation Update

The automator update deps job sync failed.

- As part of update-deps flow, "controller-runtime" has been upgraded to "v0.24.0", which required manual update of k8s.io/* packages to "v0.36.0" version.

- Version update of "controller-runtime" updated go version from "1.25.7" to "1.26.0"

- In additional, "controller-runtime v0.24.0 deprecated "scheme.Builder" and that error popped up in the linter check error: SA1019: scheme.Builder is deprecated

  Replaced "&scheme.Builder{GroupVersion: GroupVersion}" with "runtime.NewSchemeBuilder(addKnownTypes)" and added "addKnownTypes()" function to register API types. Removed "init()" function from type files to align with the new structure.

- Added "return" after "t.Fatal()" in api/v1/istiorevision_types_test.go to satisfy staticcheck SA5011.

Fixes #

Related Issue/PR #


go.mod
```diff
@@ -27,14 +27,14 @@ require (
...
-	k8s.io/client-go v0.35.3
-	sigs.k8s.io/controller-runtime v0.23.3
+	istio.io/client-go v1.30.0-alpha.1.0.20260502131957-7307deff563c
+	istio.io/istio v0.0.0-20260502185226-0424593ecc7a
+	k8s.io/api v0.36.0
+	k8s.io/apiextensions-apiserver v0.36.0
+	k8s.io/apimachinery v0.36.0
+	k8s.io/cli-runtime v0.36.0
+	k8s.io/client-go v0.36.0
+	sigs.k8s.io/controller-runtime v0.24.0
```
sridhargaddam, line 37: Does upgrading the controller-runtime create any issues to the minimum k8s version that can be supported?

sridhargaddam (commented):
Have a question, but I'll let @nrfox take a closer look at the PR.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/istio-ecosystem/sail-operator
 
-go 1.25.7
+go 1.26.0
```
FilipB, line 3: Upstream istio is still on 1.25.7. Is it safe to upgrade before upstream? I'm not sure if this will be overwritten by automator or not.

go.mod
MaxBab: Yes, unfortunately, I think it would bump the version up.
client-go v0.36.0 officially supports N-2 (1.34-1.36)

go.mod
MaxBab: No, I don't think we should move ahead of Istio upstream version.

MaxBab:
@sridhargaddam @FilipB @nrfox 

The `update-deps` job tried to bump the `controller-runtime` version and failed because of dependencies.
Probably better for us to stay on the latest `v0.23.x` `controller-runtime` version.

But I think that for the next update-deps job execution, the job will try again to upgrade the version of "controller-runtime" at the current flow.
So, the question is, what are we doing with that.

MaxBab:
Closing the PR, as the fix should be make in a different way.
We should not increase go version ahead of upstream istio.

MaxBab:
The update-deps job failure has been fixed by the following:
https://github.com/istio-ecosystem/sail-operator/pull/1909

PR #1908, opened 2026-05-04
Add Ambient e2e dependency tests

MaxBab:
#### What type of PR is this?

- [ ] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [X] Test
- [ ] Documentation Update

Add dependency testing for Ambient test suite.
Add the following tests:
- Dependency detection and health propagation. Test the dependency between Istio, IstioCNI and ZTunnel resources.
- Order independence. Test that ambient resources could be created in any order and will reach a health state at the end.

Fixes #

Related Issue/PR #

### DIFF
--- a/tests/e2e/ambient/ambient_dependency_test.go
+++ b/tests/e2e/ambient/ambient_dependency_test.go
@@ -1,0 +1,357 @@
+//go:build e2e
+
+// Copyright Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ambient
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/Masterminds/semver/v3"
+	v1 "github.com/istio-ecosystem/sail-operator/api/v1"
+	"github.com/istio-ecosystem/sail-operator/pkg/istioversion"
+	"github.com/istio-ecosystem/sail-operator/pkg/kube"
+	. "github.com/istio-ecosystem/sail-operator/pkg/test/util/ginkgo"
+	"github.com/istio-ecosystem/sail-operator/tests/e2e/util/cleaner"
+	"github.com/istio-ecosystem/sail-operator/tests/e2e/util/common"
+	. "github.com/istio-ecosystem/sail-operator/tests/e2e/util/gomega"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	appsv1 "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ = Describe("Ambient Dependency Management", Label("ambient", "ambient-dependency", "smoke"), Ordered, func() {
+	SetDefaultEventuallyTimeout(time.Duration(defaultTimeout) * time.Second)
+	SetDefaultEventuallyPollingInterval(time.Second)
+
+	Describe("Dependency Detection and Health Propagation", func() {
+		for _, version := range istioversion.GetLatestPatchVersions() {
+			// The minimum supported version for ambient is 1.24 (and above)
+			if version.Version.LessThan(semver.MustParse("1.24.0")) {
+				continue
+			}
+
+			// FIPS clusters do not support ambient mode for versions below 1.28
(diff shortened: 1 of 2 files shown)


tests/e2e/ambient/ambient_dependency_test.go
```diff
@@ -0,0 +1,442 @@
...
+	"github.com/istio-ecosystem/sail-operator/tests/e2e/util/cleaner"
+	"github.com/istio-ecosystem/sail-operator/tests/e2e/util/common"
+	. "github.com/istio-ecosystem/sail-operator/tests/e2e/util/gomega"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	appsv1 "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ = Describe("Ambient Dependency Management", Label("ambient-dependency"), Ordered, func() {
```
fjglira: It should be part of the smoke test also? Can we add also the ambient regular label. The idea is to gather test by type to have easier executions using those labels

tests/e2e/ambient/ambient_dependency_test.go
```diff
@@ -0,0 +1,442 @@
...
+	SetDefaultEventuallyTimeout(time.Duration(defaultTimeout) * time.Second)
+	SetDefaultEventuallyPollingInterval(time.Second)
+
+	// Use the latest supported ambient version for dependency tests
+	version := getLatestAmbientVersion()
```
fjglira: This means that this test will only run against the latest version available in the versions.yaml file?

tests/e2e/ambient/ambient_dependency_test.go
```diff
@@ -0,0 +1,442 @@
...
+			AfterAll(func(ctx SpecContext) {
+				if CurrentSpecReport().Failed() && keepOnFailure {
+					return
+				}
+				clr.Cleanup(ctx)
+			})
+		})
+	})
+
+	Describe("Order Independence", func() {
```
fjglira: Is this one really needed? Can we modify the actual ambient test to fit this? Because checking the ambient regular test, I see that we create on this order: CNI > Istio > Ztunnel. Maybe we can save some test time by doing it in this order, so testing both things at the same time, because the dependency status is already covered in the new test that you are adding

fjglira (commented):
Thanks for adding this test, I think we should expand more around Health status and this kind of thing that can catch regressions

tests/e2e/ambient/ambient_dependency_test.go
MaxBab: Yes, sure. I'll add "ambient" and "smoke" labels in additional to the existing.

tests/e2e/ambient/ambient_dependency_test.go
MaxBab: Yes.
I'm not sure we need to test these kind of tests across all versions.

tests/e2e/ambient/ambient_dependency_test.go
MaxBab: Yep, that's a good point. I'll refactor the existing ambient test and merge this one into the existing one.

tests/e2e/ambient/ambient_dependency_test.go
MaxBab: Added labels.

tests/e2e/ambient/ambient_dependency_test.go
MaxBab: I've looked over other tests in our e2e and all of them are looping over multiple sail versions for testing.
So, on the second thought, I've added the same multiple versions testing here as well.

tests/e2e/ambient/ambient_dependency_test.go
MaxBab: I've refactored the main ambient test to include reverse ambient deployment to not duplicate the tests.

tests/e2e/ambient/ambient_dependency_test.go
fjglira: Sorry that I didn't answer before, I think is ok to go over only latest version. Because what it tested on this test is specific behaviour of the operator, don't have anything to do with the Istio version. So, for me is ok to keep only latest

MaxBab: I moved it to a version loop, to align with the existing tests flow all over the sail operator tests suites.
Do you prefer to move it back to the latest version only?

fjglira: I prefer to latest version only. It can be really time consuming running over all the versions

fjglira (commented):
Besides not run over all the Istio versions (The dependency test) lgtm. After that change, I'll approve. We don't need to run over all the Istio versions because those dependencies and states are set by the operator and not by the specific Istio versions

fjglira (approved):

MaxBab:
> Besides not run over all the Istio versions (The dependency test) lgtm. After that change, I'll approve. We don't need to run over all the Istio versions because those dependencies and states are set by the operator and not by the specific Istio versions

You approved before I made the change. :)
I'll mae a follow up PR to modify it to a latest version only.

PR #1931, opened 2026-05-10
Add 1.30.0-rc.0 charts

MaxBab:
Add 1.30.0-rc.0 charts for early testing phase.


FilipB (commented):
Don't we need this in `release-1.30`? Or you want to add this to both main and release-1.30?

MaxBab:
> Don't we need this in `release-1.30`? Or you want to add this to both main and release-1.30?

Yes, I would like to add it both to "main" and "release-1.30" branches.

FilipB (approved):

PR #1938, opened 2026-05-12
Add Ambient "targetref" and "validation" tests

MaxBab:
#### What type of PR is this?

- [ ] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [X] Test
- [ ] Documentation Update

- Add ambient targetref test
  - Value propagation from TargetRef
  - User value override precedence
  - Missing TargetRef resource
- Add ambient validation test
  - Validate naming enforcement for "ztunnel" and "istiocni" CRs

Fixes #

Related Issue/PR #

### DIFF
--- a/tests/e2e/ambient/ambient_targetref_test.go
+++ b/tests/e2e/ambient/ambient_targetref_test.go
@@ -1,0 +1,618 @@
+//go:build e2e
+
+// Copyright Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ambient
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/Masterminds/semver/v3"
+	v1 "github.com/istio-ecosystem/sail-operator/api/v1"
+	"github.com/istio-ecosystem/sail-operator/pkg/istioversion"
+	"github.com/istio-ecosystem/sail-operator/pkg/kube"
+	. "github.com/istio-ecosystem/sail-operator/pkg/test/util/ginkgo"
+	"github.com/istio-ecosystem/sail-operator/tests/e2e/util/cleaner"
+	"github.com/istio-ecosystem/sail-operator/tests/e2e/util/common"
+	. "github.com/istio-ecosystem/sail-operator/tests/e2e/util/gomega"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	appsv1 "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ = Describe("Ambient TargetRef Behavior", Label("ambient", "ambient-targetref"), Ordered, func() {
+	SetDefaultEventuallyTimeout(time.Duration(defaultTimeout) * time.Second)
+	SetDefaultEventuallyPollingInterval(time.Second)
+
+	// Use the latest supported ambient version for these tests
+	version := getLatestAmbientVersion()
+
+	var clr cleaner.Cleaner
+
+	BeforeAll(func(ctx SpecContext) {
+		clr = cleaner.New(cl)
+		clr.Record(ctx)
+
+		Expect(k.CreateNamespace(controlPlaneNamespace)).To(Succeed())
+		Expect(k.CreateNamespace(istioCniNamespace)).To(Succeed())
+		Expect(k.CreateNamespace(ztunnelNamespace)).To(Succeed())
+
+		// Create IstioCNI CR for shared use across scenarios
+		common.CreateIstioCNI(k, version.Name, `
+profile: ambient`)
+		Success("IstioCNI created")
+	})
+
+	AfterAll(func(ctx SpecContext) {
+		if CurrentSpecReport().Failed() {
+			common.LogDebugInfo(common.Ambient, k)
+			if keepOnFailure {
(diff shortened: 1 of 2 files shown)


tests/e2e/ambient/ambient_targetref_test.go
```diff
@@ -0,0 +1,518 @@
...
+				Success("Value propagation working - ZTunnel DaemonSet is deployed")
+			})
+		})
+	})
+})
+
+// getLatestAmbientVersion returns the latest supported version for ambient mode
+func getLatestAmbientVersion() istioversion.VersionInfo {
+	versions := istioversion.GetLatestPatchVersions()
+	for i := len(versions) - 1; i >= 0; i-- {
```
sridhargaddam: If we want to validate against the latest version, shouldn’t the loop be the following?

for i := 0; i < len(versions); i++ {...}

tests/e2e/ambient/ambient_targetref_test.go
```diff
@@ -0,0 +1,518 @@
...
+	Context("User Value Override Precedence", Ordered, func() {
+		inheritedNetwork := "inherited-net"
+		overrideNetwork := "override-net"
+		customEnvVar := "CUSTOM_TEST_VAR"
+		customEnvValue := "custom-value"
+
+		BeforeAll(func(ctx SpecContext) {
+			// Delete the ZTunnel from the previous context since only one ZTunnel named 'default' can exist
+			// Ignore error if it doesn't exist
+			_ = k.Delete("ztunnel", "default")
```
sridhargaddam, line 331: We are calling k.Delete and the code immediately proceeds with creating the ztunnel at L259. If the old ZTunnel is still terminating, the creation can fail or produce unexpected behavior. Its better to use Eventually and ensure that its deleted before moving forward.

tests/e2e/ambient/ambient_targetref_test.go
```diff
@@ -0,0 +1,518 @@
...
+				Success(fmt.Sprintf("ZTunnel now uses inherited NETWORK=%s", inheritedNetwork))
+			})
+		})
+	})
+
+	Context("Invalid TargetRef Handling", Ordered, func() {
+		BeforeAll(func(ctx SpecContext) {
+			// Delete the ZTunnel from the previous context since only one ZTunnel named 'default' can exist
+			// Ignore error if it doesn't exist
+			_ = k.Delete("ztunnel", "default")
```
sridhargaddam, line 468: same as above...

tests/e2e/ambient/ambient_targetref_test.go
```diff
@@ -0,0 +1,518 @@
...
+			common.LogDebugInfo(common.Ambient, k)
+			debugInfoLogged = true
+		}
+		clr.Cleanup(ctx)
+	})
+
+	Context("Value Propagation from TargetRef", Ordered, func() {
+		networkValue := "test-network"
+
+		When("Istio CR is created with custom global.network value", func() {
```
sridhargaddam, line 73: Can you add a test case to validate the propagation of `spec.values.meshConfig` field as well.

tests/e2e/ambient/ambient_targetref_test.go
```diff
@@ -0,0 +1,518 @@
...
+
+					found := false
+					for _, container := range containers {
+						for _, env := range container.Env {
+							if env.Name == "NETWORK" && env.Value == networkValue {
+								found = true
+								break
+							}
+						}
+					}
```
sridhargaddam: Can you extract this to some helper API as its used across multiple places in the code.

tests/e2e/ambient/ambient_targetref_test.go
```diff
@@ -0,0 +1,518 @@
...
+		if version.Version.LessThan(semver.MustParse("1.24.0")) {
+			continue
+		}
+		// FIPS clusters require v1.28+
+		if fipsCluster && version.Version.LessThan(semver.MustParse("1.28.0")) {
+			continue
+		}
+		return version
+	}
+	// Fallback to the last version if no suitable version found
```
sridhargaddam, line 604: I think the comment is not correct. Based on the current logic, it actually falls back to the oldest version rather than the latest one.

sridhargaddam (commented):
quick pass....

tests/e2e/ambient/ambient_targetref_test.go
MaxBab: You're correct.
Added "Eventually" check to validate deletion.

tests/e2e/ambient/ambient_targetref_test.go
MaxBab: Added.

tests/e2e/ambient/ambient_targetref_test.go
MaxBab: Added a new test case covering "meshConfig" configuration overrides.

tests/e2e/ambient/ambient_targetref_test.go
MaxBab: Created a separate function to eliminate the duplication.
That's for pointing on this one.

tests/e2e/ambient/ambient_targetref_test.go
MaxBab: This loop is a versions order reverse loop to catch the latest version earlier.

tests/e2e/ambient/ambient_targetref_test.go
MaxBab: This loop is a versions order reverse loop to catch the latest version earlier.
So the fallback returns the latest version as expected.

tests/e2e/ambient/ambient_targetref_test.go
sridhargaddam: Looking at the code it does not appear to be the case. Are you really sure that the API `getLatestAmbientVersion` returns the latest supported version. To me it appears like it will return the oldest version and not the latest version.

MaxBab: @sridhargaddam 

Hi Sridhar,

Thank you. You were right.
The loop was wrong.
I've modified it to a simple range loop.

sridhargaddam (approved):
LGTM, thanks @MaxBab

PR #1959, opened 2026-05-18
Add monitoring controller for metrics 

mayleighnmyers:
## Summary

Implements [SEP8471](https://github.com/istio-ecosystem/sail-operator/blob/main/enhancements/SEP8471-metrics-integration.md): a monitoring controller that automatically creates Prometheus Operator `ServiceMonitor` and `PodMonitor` resources so Istio control-plane and sidecar metrics can be scraped without manual Prometheus ConfigMap edits.

Opt-in via annotation on the `Istio` CR:

```
apiVersion: sailoperator.io/v1
kind: Istio
metadata:
  name: default
  annotations:
    sailoperator.io/monitoring: enabled
spec:
  namespace: istio-system
```

Relates to: OSSM-15085

 

#### What type of PR is this?

- [X ] Enhancement / New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [ ] Test
- [ ] Documentation Update

Fixes #

Related Issue/PR # OSSM-15085


nrfox:
Hi @mayleighnmyers,

How would users control this if they wanted to opt in or out? What will the API be for this? You can either open an issue and we can discuss it there or better yet create a [SEP](https://github.com/istio-ecosystem/sail-operator/tree/main/enhancements) for this. We need to discuss how this will integrate with the operator before we can move forward with this.

mayleighnmyers:
> Hi @mayleighnmyers,
> 
> How would users control this if they wanted to opt in or out? What will the API be for this? You can either open an issue and we can discuss it there or better yet create a [SEP](https://github.com/istio-ecosystem/sail-operator/tree/main/enhancements) for this. We need to discuss how this will integrate with the operator before we can move forward with this.

So this is still in progress. Good call for the opt in/out. Something we can consider is adding a field either in the Istio CR or the istioRevision CR to allow users to opt in since right now it is currently always enabled. Yuanlin and I will be talking about this more in some future meeting when we have all ducks in a row. :)

enhancements/SEP3-observability-integration.md
```diff
@@ -0,0 +1,43 @@
...
+## Non-goals
+- Deployment of the Observability metrics components are not goals. We will implement validation methods for checking required CRD groups and CRDs. We assume the Observability components such Observability operator(s), Prometheus monitoring stacks have been deployed in a Kubernetes environment. 
+
+## Design
+
+### User Stories
+1. As a user of Istio's metrics monitoring using Prometheus components, I want to automatically configure Prometheus metrics scraping jobs for Istio generated telemetry metrics.
+
+### API Changes
+We will add a boolean field `spec.monitoring.enabled` in the `Istio` and `IstioRevision` CRDs. When a user enables it, a monitoring controller will reconcile a `ServiceMonitor` CR for the mesh control plane and `PodMonitor` CRs for each namespace labeled by the Istio sidecar injection label.
```
nrfox: The implementation is missing some of the relabeling rules that exist in the [upstream samples](https://github.com/istio/istio/blob/709b8bfb91c2f44016c76e35036e5e2e57352edc/samples/addons/extras/prometheus-operator.yaml#L19-L46). Do we anticipate users needing to customize those and does that change the API?

enhancements/SEP3-observability-integration.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Architecture
+We assume the related `ServiceMonitor` and `PodMonitor` CRDs and the API groups such as "monitoring.coreos.com/v1" , "monitoring.rhobs/v1" are available. We will add required RBAC permissions for creating, deleting and updating those two resources in the Sail Operator ClusterRole.
```
nrfox: We can't assume this since the Sail Operator is supported on non-openshift clusters that may not have these APIs installed. Even on OpenShift clusters, how will the operator know whether to use `monitoring.coreos.com/v1` or `monitoring.rhobs/v1`? Will we need to support both?

enhancements/SEP3-observability-integration.md
nrfox: I think we need to mention that it's an openshift specific requirement that we create a `PodMonitor` in each namespace otherwise non-openshift users would benefit from just using the `NamspaceSelector` on the [PodMonitor](https://prometheus-operator.dev/docs/api-reference/api/#monitoring.coreos.com/v1.PodMonitorSpec). [This](https://github.com/istio/istio/blob/208b4ce65dfc6979595d0ee4d4e7f41365a76105/samples/addons/extras/prometheus-operator.yaml#L10-L14) is how the upstream sample is configured.

nrfox: When you are using COO, is this label different for every prom stack: `monitored-by: coo-prometheus`? Will the API need to include the ability to customize this label?

nrfox:
The implementation of this broadly looks fine but I think the API will need to be larger than a boolean and will require the sail operator to have additional permissions which may not be relevant for non-openshift users. With `ZTunnel` and `IstioRevisionTag` we have `targetRef` fields that reference the `Istio` resource. Maybe we could do something similar with this integration.

There could be a standalone `Integration` resource that has 1. an `Istio` ref 2. refs to metrics or tracing resources. Something like:
```
  apiVersion: integration.ossm/v1alpha1
  kind: Integration
  metadata:
    name: my-integration
    namespace: istio-system
  spec:
    istioRef:
      name: default
    metrics:
      type: ClusterObservability
      clusterObservability:
        monitoringStackRef:
          name: my-stack
          namespace: monitoring
    tracing:
      type: TempoStack
      tempoStackRef:
        name: my-tempo
        namespace: tracing
```
Here `istioRef` is a reference to an `Istio` resource, `monitoringStackRef` is a reference to a `MonitoringStack`, and `tempoStackRef` is a reference to a `TempoStack`. For each of these references the controller would lookup the relevant fields in the reference and configure the `Istio` accordingly. For example, the controller would use the `monitoringStackRef` to find the resource label for the `PodMonitor`/`ServiceMonitor` resources.

This could even be its own deployment with its own ServiceAccount so that the Sail Operator wouldn't need to be granted additional permissions. Since a lot of the integrations may be OSSM specific, it may even make more sense to only have this available in `ossm/sail-operator`. Not sure.

Server side apply can ensure that any updates that the integrations controller makes to the `Istio` resource do not conflict with user supplied values.

The implementation included here would largely stay the same. The main change would be in the API to manage these integrations.

enhancements/SEP3-observability-integration.md
yxun: > Do we anticipate users needing to customize those and does that change the API?

No, we will add a method about appending those relabeling spec as default value in the following commits.

> When you are using COO, is this label different for every prom stack: monitored-by: coo-prometheus? 

Yes, that label value depends on the prom stack name created by user.

enhancements/SEP3-observability-integration.md
yxun: I will add validation methods. It will query API server and validate if those two CRDs , APIs are installed.
I suggest using and supporting the `monitoring.coreos.com/v1`  group. The rhobs is specific from COO operator side and no difference in the two CRDs we use.

tests/e2e/setup/setup-kind.sh
```diff
@@ -72,6 +72,7 @@ if [ "${MULTICLUSTER}" == "true" ]; then
     # Apply Gateway API CRDs which are needed for Multi-Cluster Ambient testing
     for config in "${KUBECONFIGS[@]}"; do
         kubectl apply --kubeconfig="$config" --server-side -f "${SCRIPTPATH}/testdata/gateway-api/experimental-install.yaml"
+        kubectl apply --kubeconfig="$config" --server-side -f "${SCRIPTPATH}/testdata/monitoring-rhobs/"
```
yxun: hey @mayleighnmyers , I think we don't plan to install the CRDs of `PodMonitor` and `ServiceMonitor` in e2e . They are included in Prometheus Operator and `kube-prometheus-stack` chart. 
When we setup KinD, those dependencies should be installed there. 

When we are only adding integration tests, we may install CRDs but that's not how users would setup a cluster.

tests/e2e/setup/install-kube-prometheus-stack.sh
```diff
@@ -0,0 +1,46 @@
...
+helm repo add prometheus-community https://prometheus-community.github.io/helm-charts
+helm repo update
+
+helm install "${PROM_RELEASE}" prometheus-community/kube-prometheus-stack \
+  --namespace "${PROM_NAMESPACE}" \
+  --create-namespace \
+  --set grafana.enabled=false \
+  --set alertmanager.enabled=false \
+  --set prometheus.prometheusSpec.serviceMonitorSelectorNilUsesHelmValues=false \
+  --set prometheus.prometheusSpec.podMonitorSelectorNilUsesHelmValues=false \
```
yxun:
```
--set prometheus.prometheusSpec.serviceMonitorSelectorNilUsesHelmValues=false \
--set prometheus.prometheusSpec.podMonitorSelectorNilUsesHelmValues=false \
```

By default, those two `MonitorSelectorNilUsesHelmValues` are true in `kube-prometheus-stack` chart. Most likely, users will not explicitly set those two to false when apply that chart on KinD.

We want to remove those two lines in our e2e setup. And add an additional label `release: <helm chart release name>` in both `ServiceMointor` and `PodMonitor` objects when they are created by the monitoring controller.

Ref: https://github.com/istio-ecosystem/sail-operator/pull/2028#issuecomment-4904401578

controllers/monitoring/monitoring_controller.go
```diff
@@ -0,0 +1,483 @@
...
+func (r *Reconciler) buildServiceMonitor(rev *v1.IstioRevision) *monitoringv1.ServiceMonitor {
+	name := rev.Name + serviceMonitorNameSuffix
+	namespace := rev.Spec.Namespace
+	relabelCfg := relabeling.ForPlatform(r.Config.Platform, istioOwnerName(rev), tuningEnabled)
+
+	sm := &monitoringv1.ServiceMonitor{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+			Labels:    r.monitorLabels("istiod"),
```
yxun: Could we add an owner label for all `ServiceMonitor` and `PodMonitor` objects created by the monitoring controller ?  
We can start from using a label such as `managed-by: sail-operator`. This will be useful when we need to select all `ServiceMonitor` and `PodMonitor` objects created by the monitoring controller but not those created manually from other source.

controllers/monitoring/monitoring_controller.go
```diff
@@ -0,0 +1,483 @@
...
+const (
+	serviceMonitorNameSuffix = "-istiod"
+	podMonitorNameSuffix     = "-proxies"
+
+	// COO (Cluster Observability Operator) API group
+	rhobsAPIGroup   = "monitoring.rhobs"
+	rhobsAPIVersion = "v1"
+
+	// Labels
+	cooMonitoredByLabel = "monitored-by"
```
yxun:
```suggestion
	monitoredByLabel = "monitored-by"
```

This label `"monitored-by"` can be used in general. 
When monitor resources are created on regular kube clusters, its default value can be `"kube-prometheus"` . For example, users can deploy Prometheus using `kube-prometheus-stack` helm chart on KinD.
When monitor resources are created on OpenShift, we can set the label value "coo-prometheus" as your method did below.

api/v1/istio_types.go
```diff
@@ -63,6 +63,24 @@ type IstioSpec struct {
...
+	// Defines the configuration for Prometheus monitoring integration.
+	// When enabled, the operator creates ServiceMonitor and PodMonitor resources
+	// for scraping Istio metrics using the Cluster Observability Operator (COO).
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Monitoring"
+	// +optional
+	Monitoring *MonitoringConfig `json:"monitoring,omitempty"`
+}
+
+// MonitoringConfig defines the configuration for Prometheus monitoring integration
+type MonitoringConfig struct {
```
yxun: API changes are discussed in https://github.com/istio-ecosystem/sail-operator/pull/2028

When the enhancement PR is merged, we should revisit this PR's API implementation.

tests/e2e/monitoring/monitoring_test.go
```diff
@@ -0,0 +1,175 @@
...
+
+		It("does not create a PodMonitor in the control plane namespace", func(ctx SpecContext) {
+			pm := &monitoringv1.PodMonitor{}
+			pm.SetGroupVersionKind(monitoringGV.WithKind("PodMonitor"))
+			Consistently(cl.Get).WithArguments(ctx, client.ObjectKey{Name: podMonitorName, Namespace: controlPlaneNamespace}, pm).
+				ShouldNot(Succeed(), "PodMonitor should not be created in istio-system")
+			Success("PodMonitor not created in control plane namespace")
+		})
+	})
+
```
yxun, line 194: We should add test cases about `It("discovers the proxy PodMonitor target in Prometheus" ...` after deploying a sample application in the sidecar injected namespace.

tests/e2e/setup/install-kube-prometheus-stack.sh
```diff
@@ -0,0 +1,46 @@
...
+  --wait --timeout 15m
```
yxun: Do we need wait such a long time as 15 minutes before timeout ?
This is a hard-coded wait condition now. We may investigate some condition or status value when helm install that chart.

tests/e2e/setup/install-kube-prometheus-stack.sh
```diff
@@ -0,0 +1,46 @@
...
+
+kubectl wait --for=condition=available "deployment/${PROM_RELEASE}-operator" -n "${PROM_NAMESPACE}" --timeout=10m
+kubectl rollout status "statefulset/prometheus-${PROM_RELEASE}-prometheus" -n "${PROM_NAMESPACE}" --timeout=10m
```
yxun: Same as comment above.

chart/values.yaml
```diff
@@ -97,7 +97,7 @@ csv:
     features.operators.openshift.io/cnf: "false"
     features.operators.openshift.io/cni: "true"
     features.operators.openshift.io/csi: "false"
-image: quay.io/sail-dev/sail-operator:1.30-latest
+image: localhost:5000/sail-operator:e2e-test
```
yxun: why do you change this image to localhost ? It should use `quay.io/sail-dev/sail-operator:1.30-latest` as it's a top level chart values yaml.

pkg/constants/constants.go
```diff
@@ -64,6 +64,14 @@ const (
 	// ManagedByLabelValue is the ManagedByKey label value the operator sets on all objects it creates
 	ManagedByLabelValue = "sail-operator"
 
+	// MonitoringAnnotationKey is an annotation on the Istio CR that enables the monitoring
+	// controller to create ServiceMonitor and PodMonitor resources.
+	// See enhancements/SEP8471-metrics-integration.md.
+	MonitoringAnnotationKey = MetadataNamespace + "/monitoring"
+
+	// MonitoringAnnotationEnabled is the annotation value that turns monitoring on.
+	MonitoringAnnotationEnabled = "enabled"
```
yxun:
```suggestion
	MonitoringAnnotationEnabledValue = "enabled"
```

Follow naming convention as the `IstioInjectionEnabledValue` . And then update the usages in controller_test, controller.go, and e2e monitoring_test.go

controllers/monitoring/monitoring_controller.go
```diff
@@ -0,0 +1,453 @@
...
+			}
+			return false, fmt.Errorf("failed to get Istio CR: %w", err)
+		}
+		return istio.Annotations[constants.MonitoringAnnotationKey] == constants.MonitoringAnnotationEnabled, nil
+	}
+	return false, nil
+}
+
+// TODO: map tuningEnabled from an Integration API spec field
+var tuningEnabled = false
```
yxun:
```suggestion

```

Remove this var and keep the TODO comment if we are not going to set it by default. We may either enabling tuning relabeling by default or enabling that in next API enhancement PR.

controllers/monitoring/monitoring_controller.go
```diff
@@ -0,0 +1,453 @@
...
+
+	// Skip if the IstioRevision is being deleted
+	if rev.DeletionTimestamp != nil {
+		log.V(2).Info("IstioRevision is being deleted, skipping monitoring reconciliation")
+		return ctrl.Result{}, nil
+	}
+
+	enabled, err := r.parentMonitoringEnabled(ctx, rev)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to check if monitoring is enabled: %w", err)
```
yxun:
```suggestion
		return ctrl.Result{}, fmt.Errorf("failed to check if parent Istio CR monitoring is enabled: %w", err)
```

controllers/monitoring/monitoring_controller.go
```diff
@@ -0,0 +1,453 @@
...
+		return ctrl.Result{}, fmt.Errorf("failed to reconcile ServiceMonitor: %w", err)
+	}
+
+	// Reconcile PodMonitors for istio-proxy sidecars in namespaces with injection enabled
+	if err := r.reconcilePodMonitors(ctx, rev); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to reconcile PodMonitors: %w", err)
+	}
+
+	log.Info("Monitoring resources reconciled successfully")
+	return ctrl.Result{}, nil
```
yxun: According to [istio_controller.go](https://github.com/istio-ecosystem/sail-operator/blob/main/controllers/istio/istio_controller.go) , the controller structure should be :

`func (r *Reconciler) Reconcile` calls `func (r *Reconciler) doReconcile` . `doReconcile` is the function that actually reconciles the resources.

Better to keep consistent structure across controllers. Need a refactoring and moving detailed calls into `doReconcile` method.

controllers/monitoring/monitoring_controller.go
```diff
@@ -0,0 +1,453 @@
...
+	err := r.Client.Get(ctx, client.ObjectKeyFromObject(desired), existing)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Info("Creating ServiceMonitor", "name", desired.GetName(), "namespace", desired.GetNamespace())
+			return r.Client.Create(ctx, desired)
+		}
+		return fmt.Errorf("failed to get ServiceMonitor: %w", err)
+	}
+
+	log.V(2).Info("ServiceMonitor already exists, leaving unchanged", "name", desired.GetName(), "namespace", desired.GetNamespace())
```
yxun:
```suggestion
	log.V(2).Info("ServiceMonitor already exists, leaving unchanged")
```

controllers/monitoring/monitoring_controller.go
```diff
@@ -0,0 +1,453 @@
...
+	}
+
+	return nil
+}
+
+// namespacesForRevision returns namespaces where sidecar injection is enabled for the given revision.
+// A namespace matches when istio-injection=enabled (default revision) or istio.io/rev=<revision>.
+func (r *Reconciler) namespacesForRevision(ctx context.Context, rev *v1.IstioRevision) ([]corev1.Namespace, error) {
+	nsList := &corev1.NamespaceList{}
+	if err := r.Client.List(ctx, nsList); err != nil {
```
yxun: I think there is a much simple way of listing namespaces , for example, 
`r.Client.List(ctx, client.MatchingLabels{"istio-injection": "enabeld"} ,nsList)` 
and `r.Client.List(ctx, client.MatchingLabels{"istio.io/rev": rev.name} ,nsList)` 

Could you explain your thoughts of not using `MatchingLabels` param in the Client.List ?

Because `r.Client.List(ctx, nsList)` is listing all namespaces in a cluster, that's not efficient.
We also don't need to check `if ns.Name == rev.Spec.Namespace` because Istio will not inject sidecar in control plane namespace there, we don't need double check in the monitoring controller.

controllers/monitoring/monitoring_controller.go
```diff
@@ -0,0 +1,453 @@
...
+	}
+	return namespaces, nil
+}
+
+// namespaceReferencesRevision reports whether the namespace should receive a PodMonitor for rev.
+func namespaceReferencesRevision(ns *corev1.Namespace, rev *v1.IstioRevision) bool {
+	if ns.Name == rev.Spec.Namespace {
+		return false
+	}
+	return revision.GetReferencedRevisionFromNamespace(ns.GetLabels()) == rev.Name
```
yxun: I see you want to use this function from `pkg/revision/references.go` . We may ask other's review opinion if that's a better approach here.

controllers/monitoring/monitoring_controller.go
```diff
@@ -0,0 +1,453 @@
...
+	// Labels
+	monitoredByLabel    = "monitored-by"
+	kubePrometheusValue = "kube-prometheus"
+)
+
+func (r *Reconciler) monitorLabels(app string) map[string]string {
+	return map[string]string{
+		"app":                       app,
+		constants.ManagedByLabelKey: constants.ManagedByLabelValue,
+		monitoredByLabel:            kubePrometheusValue,
```
yxun, line 70:
```suggestion
		monitoredByLabel:            kubePrometheusValue,
		"release":                   "istio",
```

If we set kubePrometheusValue = "kube-prometheus" as default const, we need a label "release" for both Monitor resources. 
Reference: https://stackoverflow.com/questions/69466567/why-is-it-that-my-prometheus-operator-servicemonitor-needs-a-release-label-to

controllers/monitoring/monitoring_controller.go
```diff
@@ -0,0 +1,453 @@
...
+	for _, ownerRef := range rev.GetOwnerReferences() {
+		if ownerRef.Kind == v1.IstioKind {
+			return ownerRef.Name
+		}
+	}
+	return rev.Name
+}
+
+// ptr returns a pointer to the given value
+func ptr[T any](v T) *T {
```
yxun: We should import `"istio.io/istio/pkg/ptr"` .

controllers/monitoring/monitoring_controller.go
```diff
@@ -0,0 +1,453 @@
...
+			Namespace: namespace,
+			Labels:    r.monitorLabels("istiod"),
+			OwnerReferences: []metav1.OwnerReference{
+				{
+					APIVersion:         v1.GroupVersion.String(),
+					Kind:               v1.IstioRevisionKind,
+					Name:               rev.Name,
+					UID:                rev.UID,
+					Controller:         ptr(true),
+					BlockOwnerDeletion: ptr(true),
```
yxun:
```suggestion
					
```

We should not block the owner IstioRev deletion here. Why do you need set BlockOwnerDeletion true here ?

bundle/manifests/sailoperator.clusterserviceversion.yaml
```diff
@@ -746,6 +746,19 @@ spec:
                 - get
                 - patch
                 - update
+            - apiGroups:
+                - monitoring.coreos.com
+              resources:
+                - servicemonitors
+                - podmonitors
+              verbs:
```
yxun, line 773: We don't need `watch, update, patch` verbs because we don't overwrite existing Monitor resources.
We can discuss if delete is required for those two resources.

chart/templates/rbac/role.yaml
```diff
@@ -324,3 +324,16 @@ rules:
   - get
   - patch
   - update
+- apiGroups:
+  - monitoring.coreos.com
+  resources:
+  - servicemonitors
+  - podmonitors
+  verbs:
```
yxun, line 332: We don't need watch, update, patch verbs because we don't overwrite existing Monitor resources.
We can discuss if delete is required for those two resources.

pkg/monitoring/relabeling/relabeling.go
```diff
@@ -0,0 +1,200 @@
...
+				PodMonitorRelabelings:     openShiftMetricTuningPodMonitorRelabelings(meshID),
+				ServiceMonitorRelabelings: nil,
+			}
+		}
+
+		return Config{
+			PodMonitorRelabelings:     openshiftPodMonitorRelabelings(meshID),
+			ServiceMonitorRelabelings: nil,
+		}
+	default:
```
yxun, line 47: For default kube platform, we should revisit Istio upstream example :
https://github.com/istio/istio/blob/master/samples/addons/extras/prometheus-operator.yaml

controllers/monitoring/monitoring_controller.go
mayleighnmyers: Yea, you are right. It doesn't make sense listing all of the namespaces. Thanks for that catch!

controllers/monitoring/monitoring_controller.go
mayleighnmyers: So, we don't need the helper here since the label-selective list will keep it more contained and easier to follow. But, more than happy for others to share what they prefer and what might be best.

controllers/monitoring/monitoring_controller.go
mayleighnmyers: I was just being cautious. We will only need an owner ref so the GC cleans up the ServiceMonitor when the istioRevision is deleted. The BlockOwnerDelection is not required. I will go ahead and delete, but keep the Controller: true.

chart/templates/rbac/role.yaml
mayleighnmyers: I’d keep delete for now:
However, 
ServiceMonitors rely on owner-ref GC (operator may not call Delete), but
PodMonitors are cross-namespace (no owner refs), so cleanup when injection labels are removed or monitoring is disabled will need Delete in a follow-up.
If we want least privilege for this PR, we can drop delete and add it when cleanup is implemented. Prefer keeping it so we don’t expand RBAC again soon — WDYT?

tests/e2e/setup/install-kube-prometheus-stack.sh
```diff
@@ -0,0 +1,67 @@
...
+helm install "${PROM_RELEASE}" prometheus-community/kube-prometheus-stack \
+  --namespace "${PROM_NAMESPACE}" \
+  --create-namespace \
+  --timeout "${WAIT_TIMEOUT}" \
+  --set grafana.enabled=false \
+  --set alertmanager.enabled=false \
+  --set prometheusOperator.admissionWebhooks.enabled=false \
+  --set prometheusOperator.tls.enabled=false \
+  --set prometheus.prometheusSpec.serviceMonitorSelectorNilUsesHelmValues=false \
+  --set prometheus.prometheusSpec.podMonitorSelectorNilUsesHelmValues=false
```
yxun:
```suggestion

```

By default, we should test `serviceMonitorSelectorNilUsesHelmValues` and `podMonitorSelectorNilUsesHelmValues` to true. Those two params default are true. We don't need to set explicitly. 

The issue is when those two are true, the `kube-prometheus-stack` step up requires a label of `release:` in `ServiceMonitor` and `PodMonitor` .
We may have a task of implementing that next week.

chart/templates/rbac/role.yaml
yxun: Right, we can keep delete there and remove the watch, update, patch RBAC.

controllers/monitoring/monitoring_controller.go
```diff
@@ -0,0 +1,496 @@
...
+	log := logf.FromContext(ctx)
+
+	log.Info("Reconciling")
+	err := r.doReconcile(ctx, rev)
+	log.Info("Reconciliation done")
+	return ctrl.Result{}, err
+}
+
+// doReconcile is the function that actually reconciles monitoring resources for the IstioRevision.
+func (r *Reconciler) doReconcile(ctx context.Context, rev *v1.IstioRevision) error {
```
nrfox: Why reconcile the `IstioRevision` instead of the `Istio` resource? If there are multiple `IstioRevisions`, for example during an upgrade, will you end up reconciling both and creating duplicate pod monitors / service monitors?

nrfox: The Istio controller has a method to get the active `IstioRevision` for an `Istio`: https://github.com/istio-ecosystem/sail-operator/blob/6c634d7b055390224b2bddaa6b29a28e1dcc65af/controllers/istio/istio_controller.go#L163-L170

This may be what you want here. If it makes sense to reconcile `Istio` rather than `IstioRevision`, you can modify that func to accept a client, remove the pointer receiver, and use it here.

mayleighnmyers: Hmmm, So here we are reconciling the `IstioRevision` because the monitors are following the injection ownership (`istio.io/rev` & `istio-injection`) (which is per-revision), not just the active revisions. During upgrades we will reconcile each revision, but it won't create duplicates. The ServiceMonitor and PodMontior names are revision-scoped ( `{rev}-isitiod-metrics`, and `{rev}-proxies-metrics`), and the PodMonitors only are created in the namespace that matches the revision. Also, Wouldn't using the getActiveRevision drop the monitoring for the non-active but still-serving revisions during any revision based upgrades? If we are wanting to do the active only scraping, I can go ahead and change it to what you suggested.

nrfox: No that makes sense to reconcile all the revisions, not just the active one. I think we should still switch to reconciling `Istio` and finding all the `IstioRevision`s for that `Istio` rather than reconciling `IstioRevisions` directly and tying them back to the parent `Istio`. Controllers work better when you [reconcile the parent resource and not the child](https://book.kubebuilder.io/reference/watching-resources.html).

mayleighnmyers: Okay! That makes sense. I will switch it to reconcile `Istio`, list its owned `IstioRevisions`, and create monitor for each. We will also still cover all of the revisions, not just active ones.

controllers/monitoring/monitoring_controller.go
```diff
@@ -0,0 +1,448 @@
...
+		return fmt.Errorf("failed to get ServiceMonitor: %w", err)
+	}
+
+	log.V(2).Info("ServiceMonitor already exists, leaving unchanged")
+	return nil
+}
+
+// reconcilePodMonitors creates PodMonitors for istio-proxy sidecars in namespaces
+// that reference this IstioRevision via istio-injection=enabled or istio.io/rev labels.
+func (r *Reconciler) reconcilePodMonitors(ctx context.Context, istio *v1.Istio, rev *v1.IstioRevision) error {
```
nrfox, line 176: Do we remove `PodMonitor` resources when the injection label or rev label gets removed from the namespace?

controllers/monitoring/monitoring_controller.go
```diff
@@ -0,0 +1,448 @@
...
+	relabelCfg := relabeling.ForPlatform(r.Config.Platform, istio.Name, false)
+
+	pm := &monitoringv1.PodMonitor{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+			Labels:    r.monitorLabels("istio-proxy", podMonitorMonitoring),
+			// Note: We don't set owner references here because the PodMonitor is in a different
+			// namespace than the IstioRevision (which is cluster-scoped). Cross-namespace owner
+			// references are not supported by Kubernetes.
```
nrfox: From the [kube docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/#owner-references-in-object-specifications) this is allowed.

Relevant section:
> Namespaced dependents can specify cluster-scoped or namespaced owners.

In our case the `PodMonitor` is a namespaced dependent and the `IstioRevision` is a cluster-scoped owner.

controllers/monitoring/monitoring_controller.go
nrfox: What's the behavior when the `PodMonitor` gets removed? Do you no longer see metrics for those pods?

controllers/monitoring/monitoring_controller.go
```diff
@@ -0,0 +1,448 @@
...
+		// Watch owned IstioRevisions so revision create/update/delete requeues the parent Istio.
+		// ServiceMonitor owner references on IstioRevision still handle GC on revision deletion.
+		Watches(&v1.IstioRevision{}, ownedRevisionHandler).
+		// Watch namespaces with sidecar injection labels to create PodMonitors in them
+		Watches(&corev1.Namespace{}, namespaceHandler, builder.WithPredicates(sidecarInjectionNamespacePredicate())).
+		Complete(reconciler.NewStandardReconciler[*v1.Istio](r.Client, r.Reconcile))
+}
+
+// mapNamespaceToReconcileRequest returns reconcile requests for all Istio CRs when a
+// namespace's sidecar injection labels change.
```
nrfox: When a namespace's sidecar injection label changes, we don't want to enqueue every `Istio`. Only the Istio that was referencing the namespace and potentially the new `Istio` that is referencing the namespace.

controllers/monitoring/monitoring_controller.go
mayleighnmyers: As of right now, the `PodMonitor` is not removed when the injection or `istio.io/rev `label is taken off the namespace. Reconcile is create-only: matching namespaces get a` PodMonitor`, and namespaces that no longer match are skipped, so the existing object stays.

If the `PodMonitor `is deleted, Prometheus Operator drops those scrape targets, so new proxy metrics stop (historical samples stay until retention). Removing the namespace label does not strip existing sidecars, so a leftover `PodMonitor` still scrapes those pods until they restart. Relabeling is limited to `istio-proxy` and `prometheus.io/scrape`, so once the sidecars are gone it has no targets. If the `PodMonitor` is deleted while the namespace still matches, the next reconcile recreates it.

We could prune orphans after the create loop: list PodMonitors named `{rev}-proxies-metrics` with` managed-by=sail-operator `and delete any whose namespace no longer matches injection/`istio.io/rev`. That is the same desired-state idea as `IstioRevision` pruning unused revisions. Immediate delete would stop metrics from remaining sidecars until those pods restart; we could also wait until no `istio-proxy` pods are left.

controllers/monitoring/monitoring_controller.go
mayleighnmyers: Fixed. `mapNamespaceToReconcileRequest` now resolves the namespace labels to an IstioRevision and only enqueues that revision’s owning Istio. On update the mapper runs for both the old and new object, so the previous Istio and the newly referenced one both get requeued.

controllers/monitoring/monitoring_controller.go
nrfox: I think we should keep the `PodMonitor` as long as there are pods left in the namespace that are using it. Cleaning up the `PodMonitor` when it's no longer in use and when there's no longer a label on the namespace can be a follow up.

controllers/monitoring/monitoring_controller.go
```diff
@@ -0,0 +1,479 @@
...
+			},
+		},
+	}
+
+	return sm
+}
+
+// buildPodMonitor constructs the PodMonitor for monitoring istio-proxy sidecars
+func (r *Reconciler) buildPodMonitor(istio *v1.Istio, rev *v1.IstioRevision, namespace string) *monitoringv1.PodMonitor {
+	name := rev.Name + podMonitorNameSuffix
```
nrfox, line 313: Should this be `rev.Name` or `istio.Name`? Building a `PodMonitor` per revision will mean we may end up with multiple in a namespace but the label selector for the `PodMonitor` does not discriminate based on rev so we may have two scrape jobs for the same targets.

controllers/monitoring/monitoring_controller.go
```diff
@@ -0,0 +1,479 @@
...
+			Labels:    r.monitorLabels("istio-proxy", podMonitorMonitoring),
+			OwnerReferences: []metav1.OwnerReference{
+				{
+					APIVersion: v1.GroupVersion.String(),
+					Kind:       v1.IstioRevisionKind,
+					Name:       rev.Name,
+					UID:        rev.UID,
+					Controller: ptr.Of(true),
+				},
+			},
```
nrfox, line 330: Same with this ownerref. Since the `PodMonitor` is not tied to a revision, I think the ownerref should be the `Istio` instance. Otherwise when an `IstioRevision` goes away after a revisioned upgrade, the `PodMonitor`s would get deleted as well.

pkg/monitoring/relabeling/relabeling.go
```diff
@@ -0,0 +1,200 @@
...
+	// PodMonitorRelabelings are applied to podMetricsEndpoints.relabelings on PodMonitor resources.
+	PodMonitorRelabelings []monitoringv1.RelabelConfig
+	// ServiceMonitorRelabelings are applied to endpoints.relabelings on ServiceMonitor resources.
+	ServiceMonitorRelabelings []monitoringv1.RelabelConfig
+}
+
+// ForPlatform returns default relabeling configuration for the given platform.
+// meshID is the Istio CR name used as the mesh_id label on OpenShift; it is ignored on Kubernetes.
+// tuningEnabled is an option for metric thinning usages.
+func ForPlatform(platform config.Platform, meshID string, tuningEnabled bool) Config {
```
nrfox, line 33: Why only add `meshID` for openshift? `meshID` can be set in kube as well.

pkg/monitoring/relabeling/relabeling.go
```diff
@@ -0,0 +1,200 @@
...
+			out[i].Separator = strPtr(*cfg.Separator)
+		}
+		if cfg.Replacement != nil {
+			out[i].Replacement = strPtr(*cfg.Replacement)
+		}
+		if len(cfg.SourceLabels) > 0 {
+			out[i].SourceLabels = append([]monitoringv1.LabelName(nil), cfg.SourceLabels...)
+		}
+	}
+	return out
```
nrfox, line 195: All kube objects implement `DeepCopy`: https://pkg.go.dev/github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1#RelabelConfig.DeepCopy

tests/e2e/monitoring/monitoring_test.go
```diff
@@ -0,0 +1,220 @@
...
+	BeforeAll(func(ctx SpecContext) {
+		clr.Record(ctx)
+		Expect(k.CreateNamespace(controlPlaneNamespace)).To(Succeed(), "Istio namespace failed to be created")
+		Expect(k.CreateNamespace(istioCniNamespace)).To(Succeed(), "IstioCNI namespace failed to be created")
+	})
+
+	AfterAll(func(ctx SpecContext) {
+		clr.Cleanup(ctx)
+	})
```
nrfox, line 68: nit: Use `DeferCleanup` instead of `AfterAll` so that it runs even if there's an error.

tests/e2e/monitoring/monitoring_test.go
```diff
@@ -0,0 +1,220 @@
...
+					Operator: metav1.LabelSelectorOpIn,
+					Values:   []string{"pilot"},
+				}))
+				g.Expect(sm.OwnerReferences).NotTo(BeEmpty())
+				g.Expect(sm.OwnerReferences[0].Kind).To(Equal(v1.IstioRevisionKind))
+			}).Should(Succeed())
+			Success("ServiceMonitor for istiod exists")
+		})
+
+		PIt("discovers the istiod ServiceMonitor in Prometheus", func(ctx SpecContext) {
```
nrfox, line 118: typo?
```suggestion
		It("discovers the istiod ServiceMonitor in Prometheus", func(ctx SpecContext) {
```

tests/e2e/monitoring/monitoring_test.go
```diff
@@ -0,0 +1,220 @@
...
+
+		It("does not create a PodMonitor in the control plane namespace", func(ctx SpecContext) {
+			pm := &monitoringv1.PodMonitor{}
+			pm.SetGroupVersionKind(monitoringGV.WithKind("PodMonitor"))
+			Consistently(cl.Get).WithArguments(ctx, client.ObjectKey{Name: podMonitorName, Namespace: controlPlaneNamespace}, pm).
+				ShouldNot(Succeed(), "PodMonitor should not be created in istio-system")
+			Success("PodMonitor not created in control plane namespace")
+		})
+
+		PIt("discovers the proxy PodMonitor target in Prometheus", func(ctx SpecContext) {
```
nrfox, line 173: typo?
```suggestion
		It("discovers the proxy PodMonitor target in Prometheus", func(ctx SpecContext) {
```

tests/e2e/setup/setup-kind.sh
```diff
@@ -83,8 +83,13 @@ else
   KUBECONFIG="${ARTIFACTS}/config" setup_kind_cluster "${KIND_CLUSTER_NAME}" "${KIND_IMAGE}" "" "true" "true"
   setup_kind_registry "$KIND_CLUSTER_NAME"
 
+  export KUBECONFIG="${ARTIFACTS}/config"
+
   # Apply Gateway API CRDs needed for library tests
   KUBECONFIG="${ARTIFACTS}/config" kubectl apply --server-side -f "${SCRIPTPATH}/testdata/gateway-api/experimental-install.yaml"
 
+  # Install kube-prometheus-stack for monitoring controller e2e tests
+  "${SCRIPTPATH}/install-kube-prometheus-stack.sh"
```
nrfox, line 92: I'm wary of adding this to the kind setup. As we add more integrations we will continue to add more resources to the testing/dev setup. If we are going to install these resources into the cluster then it should happen as part of the setup for the monitoring tests and then removed from the cluster as part of the teardown for the monitoring tests. You'll need to exclude OpenShift from this as iirc it comes with its own prom operator instead of the community one.

tests/e2e/setup/install-kube-prometheus-stack.sh
```diff
@@ -0,0 +1,69 @@
...
+  echo "kube-prometheus-stack already installed in ${PROM_NAMESPACE}"
+  exit 0
+fi
+
+helm repo add prometheus-community https://prometheus-community.github.io/helm-charts
+helm repo update
+
+# Install without --wait; readiness is verified via targeted kubectl waits below.
+# Disable admission webhooks so KinD does not depend on the certgen Job (fragile when
+# kube-proxy/service networking is degraded). CRDs and Prometheus scraping still work.
```
nrfox, line 42: Why not use `helm install --wait`? It's not clear from the comment what the issue is with the certgen Job and why the webhooks need to be disabled.

tests/e2e/monitoring/monitoring_test.go
nrfox, line 1: It would be good to add tests for some of the issues discussed above with the namespace label changing and what the operator should do with `PodMonitor` when it does. This will help suss out whether or not we get 2 `PodMonitors` when an upgrade happens or whether a `PodMonitor` gets deleted or not.

tests/e2e/setup/setup-kind.sh
mayleighnmyers: Okay, I changed this so the suite now installs in `BeforeAll `and uninstalls in `DeferCleanup.`

controllers/monitoring/monitoring_controller_test.go
```diff
@@ -0,0 +1,1333 @@
...
+				sm := &monitoringv1.ServiceMonitor{}
+				sm.SetGroupVersionKind(monitoringv1.SchemeGroupVersion.WithKind("ServiceMonitor"))
+				err := cl.Get(ctx, types.NamespacedName{
+					Name:      tt.expectSMRevision + serviceMonitorNameSuffix,
+					Namespace: istioNamespace,
+				}, sm)
+				g.Expect(err).ToNot(HaveOccurred())
+				g.Expect(sm.Name).To(Equal(tt.expectSMRevision + serviceMonitorNameSuffix))
+				expectMonitoringLabels(g, sm.Labels, serviceMonitorMonitoring)
+			}
```
nrfox, line 312: Seems like we need an `else` here to check the negative conditions where we don't expect the service monitor to exist?

controllers/monitoring/monitoring_controller_test.go
```diff
@@ -0,0 +1,1333 @@
...
+				}, pm)
+				g.Expect(err).ToNot(HaveOccurred())
+				g.Expect(pm.Name).To(Equal(istioName + podMonitorNameSuffix))
+				expectMonitoringLabels(g, pm.Labels, podMonitorMonitoring)
+				g.Expect(pm.OwnerReferences).To(HaveLen(1))
+				g.Expect(pm.OwnerReferences[0].Kind).To(Equal(v1.IstioKind))
+				g.Expect(pm.OwnerReferences[0].Name).To(Equal(istioName))
+				g.Expect(pm.OwnerReferences[0].Controller).ToNot(BeNil())
+				g.Expect(*pm.OwnerReferences[0].Controller).To(BeTrue())
+			}
```
nrfox, line 329: Same here.

controllers/monitoring/monitoring_controller_test.go
```diff
@@ -0,0 +1,1333 @@
...
+				Spec: v1.IstioSpec{
+					Version:   "v1.29.2",
+					Namespace: istioNamespace,
+				},
+			},
+			revisions: []*v1.IstioRevision{
+				{
+					ObjectMeta: revisionMeta,
+					Spec: v1.IstioRevisionSpec{
+						Version:   "v1.24.0",
```
nrfox, line 239: Why is the version for `Istio` `v1.29.2` and the version of the `IstioRevision` `v1.24.0`? It may not affect the tests but looks odd. You may not need to specify the version at all. It should get defaulted.

controllers/monitoring/monitoring_controller_test.go
```diff
@@ -0,0 +1,1333 @@
...
+					g.Expect(result.ResourceVersion).To(Equal("123"))
+				} else {
+					expectMonitoringLabels(g, result.Labels, serviceMonitorMonitoring)
+				}
+			}
+		})
+	}
+}
+
+func TestReconcilePodMonitors(t *testing.T) {
```
nrfox, line 582: Are these tests duplicates of `TestReconcile`?

controllers/monitoring/monitoring_controller.go
```diff
@@ -0,0 +1,482 @@
...
+
+	sm := &monitoringv1.ServiceMonitor{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+			Labels:    r.monitorLabels("istiod", serviceMonitorMonitoring),
+			OwnerReferences: []metav1.OwnerReference{
+				{
+					APIVersion: v1.GroupVersion.String(),
+					Kind:       v1.IstioRevisionKind,
```
nrfox, line 279: Shouldn't this be the `Istio` resource and not the `IstioRevision` similar to the `PodMonitor`?

tests/e2e/monitoring/install-kube-prometheus-stack.sh
```diff
@@ -0,0 +1,91 @@
...
+fi
+
+helm repo add prometheus-community https://prometheus-community.github.io/helm-charts --force-update
+helm repo update
+
+# Admission webhooks require a certgen Job that talks to the webhook Service. On KinD
+# that Job frequently times out (the webhook Service is not reachable until kube-proxy
+# is healthy), which would also make helm --wait hang. The monitoring e2e only needs CRDs
+# and scrape targets, so webhooks are disabled. --wait then tracks the operator Deployment.
+helm install "${PROM_RELEASE}" prometheus-community/kube-prometheus-stack \
```
nrfox, line 64: Doing `helm repo add` and `helm repo update` will have side effects on the machine this hack script is run on. Can you install the helm chart using `--repo` and the url directly?

tests/e2e/monitoring/install-kube-prometheus-stack.sh
nrfox, line 1: We'll probably want to expose a util package for installing the integration dependencies and handling the differences between helm vs. OLM but just keep this as is for now.

nrfox (commented):
A few minor comments otherwise LGTM

nrfox (approved):

PR #1974, opened 2026-05-27
Add E2E migration test

MaxBab:
#### What type of PR is this?

- [ ] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [X] Test
- [ ] Documentation Update

Add the following migration test scenarios from sidecar to ambient mode:
- Tests the complete migration workflow from sidecar to ambient mode.
  - Validates L4 connectivity via ztunnel and L7 policy enforcement
    via waypoint in a single namespace.
  - Migration flow:
    sidecar → HBONE coexistence → pure ambient → waypoint activation

- Tests migration resilience from sidecar to ambient mode with
  continuous traffic.
  - Uses a stable external client (sidecar mode, never migrates)
    to send continuous traffic to a server that migrates from
    sidecar → ambient while traffic flows.
  - Validates:
    - Limited disruption during pod restart
      (≤10 503 errors - brief, acceptable)
    - Traffic RECOVERS after disruption
      (≥9/10 consecutive successes post-migration)
    - Cross-mode communication works
      (sidecar client → ambient server)

Fixes #

Related Issue/PR #

### DIFF
--- a/tests/e2e/migration/migration_coexistence_test.go
+++ b/tests/e2e/migration/migration_coexistence_test.go
@@ -1,0 +1,420 @@
+//go:build e2e
+
+// Copyright Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package migration
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	v1 "github.com/istio-ecosystem/sail-operator/api/v1"
+	"github.com/istio-ecosystem/sail-operator/pkg/istioversion"
+	"github.com/istio-ecosystem/sail-operator/pkg/kube"
+	. "github.com/istio-ecosystem/sail-operator/pkg/test/util/ginkgo"
+	"github.com/istio-ecosystem/sail-operator/tests/e2e/util/cleaner"
+	"github.com/istio-ecosystem/sail-operator/tests/e2e/util/common"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	// expectedPodRestartError is the expected HTTP 503 error during pod restart in migration
+	expectedPodRestartError = "unexpected status code: 503"
+)
+
+var _ = Describe("Migration Sidecar-Ambient Coexistence Validation", Ordered,
+	Label("migration", "migration-coexistence", "slow"), func() {
+		// Tests sidecar-ambient coexistence during migration with continuous traffic.
+		// Uses a stable external client (sidecar mode, never migrates) to send traffic
+		// to a server that transitions from sidecar → HBONE → ambient while traffic flows.
+		// This validates cross-mode communication and verifies traffic resilience during migration.
+		// Validates:
+		//   1. Cross-mode communication works (sidecar client → ambient server)
+		//   2. Limited disruption during pod restart (≤10 503 errors - brief, acceptable)
+		//   3. Traffic RECOVERS after disruption (≥9/10 consecutive successes post-migration)
+		SetDefaultEventuallyTimeout(time.Duration(defaultTimeout) * time.Second)
+		SetDefaultEventuallyPollingInterval(time.Second)
+
+		version := istioversion.GetLatestAmbientVersion()
+
+		Context(fmt.Sprintf("Migration testing with Istio %s", version.Version), func() {
+			var clr cleaner.Cleaner
+			workloadNamespace := "workload-traffic"
+			stableClientNamespace := "stable-client"
+
+			BeforeAll(func(ctx SpecContext) {
+				clr = cleaner.New(cl)
+				clr.Record(ctx)
+
+				Expect(k.CreateNamespace(controlPlaneNamespace)).To(Succeed())
+				Expect(k.CreateNamespace(istioCniNamespace)).To(Succeed())
+				Expect(k.CreateNamespace(ztunnelNamespace)).To(Succeed())
+
+				common.CreateIstioCNI(k, version.Name)
+				common.AwaitCondition(ctx, v1.IstioCNIConditionReady, kube.Key(istioCniName), &v1.IstioCNI{}, k, cl)
+				Success("IstioCNI created (no profile - sidecar mode)")
+
+				common.CreateIstio(k, version.Name)
+				common.AwaitCondition(ctx, v1.IstioConditionReady, kube.Key(istioName), &v1.Istio{}, k, cl)
+				Success("Istio created (default profile - sidecar mode)")
+
+				// Wait for sidecar injector webhook to be ready
+				Eventually(func(g Gomega) {
+					webhook := &admissionregistrationv1.MutatingWebhookConfiguration{}
+					g.Expect(cl.Get(ctx, kube.Key("istio-sidecar-injector"), webhook)).To(Succeed())
+				}).Should(Succeed())
+				Success("Sidecar injector webhook is ready")
+			})
+
+			AfterAll(func(ctx SpecContext) {
+				if CurrentSpecReport().Failed() {
+					common.LogDebugInfo(common.Ambient, k)
+					if keepOnFailure {
+						return
+					}
+				}
+				clr.Cleanup(ctx)
+			})
+
+			Context("Sidecar-Ambient Coexistence with Continuous Traffic", Ordered,
+				Label("migration-coexistence-traffic"), func() {
+					BeforeAll(func(ctx SpecContext) {
+						// Create and deploy stable client (never migrates, always stays in sidecar mode)
+						Expect(k.CreateNamespace(stableClientNamespace)).To(Succeed())
+						Expect(k.Label("namespace", stableClientNamespace, "istio-injection", "enabled")).To(Succeed())
+						Expect(k.WithNamespace(stableClientNamespace).ApplyKustomize("sleep")).To(Succeed())
+						Success("Stable sleep client deployed with sidecar injection")
+
+						Eventually(func(g Gomega) {
+							pods := &corev1.PodList{}
+							g.Expect(cl.List(ctx, pods, client.InNamespace(stableClientNamespace))).To(Succeed())
+							g.Expect(pods.Items).To(HaveLen(1))
+
+							pod := pods.Items[0]
+							g.Expect(pod.Status.Phase).To(Equal(corev1.PodRunning))
+
+							for _, containerStatus := range pod.Status.ContainerStatuses {
+								g.Expect(containerStatus.Ready).To(BeTrue(),
+									"Container %s should be ready", containerStatus.Name)
+							}
+
+							g.Expect(common.HasSidecarInjected(pod)).To(BeTrue(),
+								"Stable client pod should have sidecar injected")
+						}).Should(Succeed())
+						Success("Stable client ready with sidecar")
+					})
+
+					When("server workloads are deployed in sidecar mode", func() {
+						It("creates and labels namespace for sidecar injection", func(ctx SpecContext) {
+							Expect(k.CreateNamespace(workloadNamespace)).To(Succeed())
+							Expect(k.Label("namespace", workloadNamespace, "istio-injection", "enabled")).To(Succeed())
+							Success(fmt.Sprintf("Namespace %s created and labeled for sidecar injection", workloadNamespace))
+						})
+
+						It("deploys httpbin server in sidecar mode", func(ctx SpecContext) {
+							// Only deploy httpbin - client is in stable-client namespace
+							Expect(k.WithNamespace(workloadNamespace).ApplyKustomize("httpbin")).To(Succeed())
+							Success("Httpbin server deployed in sidecar mode")
+						})
+
+						It("waits for httpbin pod to be ready with sidecar", func(ctx SpecContext) {
+							Eventually(func(g Gomega) {
+								pods := &corev1.PodList{}
+								g.Expect(cl.List(ctx, pods, client.InNamespace(workloadNamespace))).To(Succeed())
+								g.Expect(pods.Items).To(HaveLen(1), "Expected 1 pod (httpbin)")
+
+								pod := pods.Items[0]
+								g.Expect(pod.Status.Phase).To(Equal(corev1.PodRunning))
+								g.Expect(common.HasSidecarInjected(pod)).To(BeTrue(),
+									"Pod %s should have istio-proxy in either containers or init containers", pod.Name)
+							}).Should(Succeed())
+							Success("Httpbin running with sidecar")
+						})
+
+						It("verifies initial traffic works from stable client", func(ctx SpecContext) {
+							// Get stable client pod (never migrated)
+							pods := &corev1.PodList{}
+							Expect(cl.List(ctx, pods, client.InNamespace(stableClientNamespace), client.MatchingLabels{"app": "sleep"})).To(Succeed())
+							Expect(pods.Items).To(HaveLen(1))
+
+							clientPod := pods.Items[0].Name
+							GinkgoWriter.Printf("Using stable client: %s (in namespace %s)\n", clientPod, stableClientNamespace)
+
+							Eventually(func() error {
+								// Stable client (sidecar mode) → httpbin (sidecar mode)
+								targetURL := fmt.Sprintf("httpbin.%s.svc.cluster.local:8000/get", workloadNamespace)
+								return common.CheckHTTPConnectivity(k, stableClientNamespace, clientPod, "sleep", targetURL, 5, "200")
+							}).Should(Succeed())
+							Success("Traffic flows from stable client to httpbin (both sidecar mode)")
+						})
+					})
+
+					When("namespace is switched to ambient mode during traffic", func() {
+						var (
+							stats      *common.HTTPTrafficStats
+							cancelFunc context.CancelFunc
+						)
+
+						// Ensure continuous traffic is stopped when this When block completes
+						AfterAll(func() {
+							if cancelFunc != nil {
+								GinkgoWriter.Printf("Cleaning up: stopping any remaining continuous traffic...\n")
+								cancelFunc()
+								time.Sleep(500 * time.Millisecond)
+							}
+						})
+
+						It("starts continuous traffic from stable client", func(ctx SpecContext) {
+							pods := &corev1.PodList{}
+							Expect(cl.List(ctx, pods, client.InNamespace(stableClientNamespace), client.MatchingLabels{"app": "sleep"})).To(Succeed())
+							Expect(pods.Items).To(HaveLen(1))
+							clientPod := pods.Items[0].Name
+							GinkgoWriter.Printf("Using stable client pod: %s (in namespace %s)\n", clientPod, stableClientNamespace)
+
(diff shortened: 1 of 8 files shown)


tests/e2e/migration/migration_procedure_test.go
```diff
@@ -0,0 +1,623 @@
...
+	Label("migration", "migration-procedure", "slow"), func() {
+		// Tests the complete migration workflow from sidecar to ambient mode.
+		// Validates L4 connectivity via ztunnel and L7 policy enforcement via waypoint in a single namespace.
+		// Migration flow: sidecar → HBONE coexistence → pure ambient → waypoint activation
+		SetDefaultEventuallyTimeout(time.Duration(defaultTimeout) * time.Second)
+		SetDefaultEventuallyPollingInterval(time.Second)
+
+		version := istioversion.GetLatestAmbientVersion()
+
+		Context(fmt.Sprintf("Migration testing with Istio %s", version.Version), func() {
```
fjglira:
```suggestion
		Context(fmt.Sprintf("Migration from sidecar to ambient with Istio version %s", version.Version), func() {
```

tests/e2e/migration/migration_procedure_test.go
```diff
@@ -0,0 +1,623 @@
...
+				if CurrentSpecReport().Failed() {
+					common.LogDebugInfo(common.Ambient, k)
+					if keepOnFailure {
+						return
+					}
+				}
+				clr.Cleanup(ctx)
+			})
+
+			Context("Sidecar → Ambient Migration", Ordered, func() {
```
fjglira:
```suggestion
			Context("Sidecar to Ambient Migration", Ordered, func() {
```

fjglira: I don't like those rows in the test logs xD

tests/e2e/migration/migration_procedure_test.go
```diff
@@ -0,0 +1,623 @@
...
+						crd := &apiextensionsv1.CustomResourceDefinition{}
+						err := cl.Get(ctx, kube.Key("gateways.gateway.networking.k8s.io"), crd)
+						if err != nil {
+							Success("Gateway API CRDs not found, installing from upstream")
+							Expect(k.Apply("https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.2.0/standard-install.yaml")).
+								To(Succeed(), "Failed to install Gateway API CRDs")
+							Success("Gateway API CRDs installed (standard channel v1.2.0)")
+						} else {
+							Success("Gateway API CRDs already present in cluster")
+						}
```
fjglira: You can move this to the setup of the test. I think is not related and it's a prerequisite of the test to be able to run, so adding a small func in the setup in the main file make sense to me

tests/e2e/migration/migration_procedure_test.go
```diff
@@ -0,0 +1,623 @@
...
+						Expect(cl.List(ctx, pods, client.InNamespace(workloadNamespace),
+							client.MatchingLabels{"app": "sleep"})).To(Succeed())
+						Expect(pods.Items).To(HaveLen(1))
+						sleepPod := pods.Items[0].Name
+
+						Eventually(func(g Gomega) {
+							cmd := fmt.Sprintf("curl -s -o /dev/null -w '%%{http_code}' httpbin:8000/get --max-time 5")
+							output, err := k.WithNamespace(workloadNamespace).Exec(sleepPod, "sleep", cmd)
+							g.Expect(err).NotTo(HaveOccurred())
+							g.Expect(output).To(Equal("200"))
```
fjglira: Can you please check if we do connectivity checks between sleep and httpbin pods. I think this is something that we can move to a func to reuse it easier and keep clean the code

tests/e2e/migration/migration_procedure_test.go
```diff
@@ -0,0 +1,623 @@
...
+						Success("AuthorizationPolicy created for sidecar mode")
+					})
+
+					It("verifies L7 traffic routing works in sidecar mode", func(ctx SpecContext) {
+						pods := &corev1.PodList{}
+						Expect(cl.List(ctx, pods, client.InNamespace(workloadNamespace),
+							client.MatchingLabels{"app": "sleep"})).To(Succeed())
+						sleepPod := pods.Items[0].Name
+
+						// Test that VirtualService routes traffic correctly
```
fjglira, line 210: Add information about what does the current VS and the expected output. It will help people to understand easier the test

tests/e2e/migration/migration_procedure_test.go
```diff
@@ -0,0 +1,623 @@
...
+						Eventually(func(g Gomega) {
+							cmd := fmt.Sprintf("curl -s -o /dev/null -w '%%{http_code}' httpbin:8000/get --max-time 5")
+							output, err := k.WithNamespace(workloadNamespace).Exec(sleepPod, "sleep", cmd)
+							g.Expect(err).NotTo(HaveOccurred())
+							g.Expect(output).To(Equal("200"), "VirtualService should route /get successfully")
```
fjglira: I think now it worth to have a func, you are using the same pattern more than once time in the same file

tests/e2e/migration/migration_procedure_test.go
```diff
@@ -0,0 +1,623 @@
...
+apiVersion: sailoperator.io/v1
+kind: ZTunnel
+metadata:
+  name: default
+spec:
+  version: %s
+  namespace: %s
+  targetRef:
+    kind: Istio
+    name: %s`, version.Name, ztunnelNamespace, istioName)
```
fjglira: maybe it worth to create a common.CreateZtunnel? And use it across all the ambient test

tests/e2e/migration/migration_procedure_test.go
```diff
@@ -0,0 +1,623 @@
...
+							webhook := &admissionregistrationv1.MutatingWebhookConfiguration{}
+							g.Expect(cl.Get(ctx, kube.Key("istio-sidecar-injector"), webhook)).To(Succeed(), "Sidecar injector webhook should exist in coexistence mode")
+						}).Should(Succeed(), "Waiting for istio-sidecar-injector webhook to be recreated")
+						Success("Sidecar injector webhook is ready")
+					})
+
+					It("updates namespace to use revision-based injection for HBONE-aware sidecars", func(ctx SpecContext) {
+						// After switching to ambient profile, use revision-based injection
+						// Remove istio-injection=enabled and add istio.io/rev=default
+						// This enables HBONE-aware sidecars in coexistence mode
```
fjglira: You can have HBONE with the `istio-injection=enabled` this is not mandatory to get the HBONE enable

tests/e2e/migration/migration_procedure_test.go
```diff
@@ -0,0 +1,623 @@
...
+					It("verifies HBONE capability is enabled in sidecars", func(ctx SpecContext) {
+						pods := &corev1.PodList{}
+						Expect(cl.List(ctx, pods, client.InNamespace(workloadNamespace))).To(Succeed())
+						Expect(pods.Items).NotTo(BeEmpty())
+
+						pod := pods.Items[0]
+						Expect(common.HasSidecarInjected(pod)).To(BeTrue(), "istio-proxy container not found in pod %s", pod.Name)
+						Expect(common.HasHBONEEnabled(pod)).To(BeTrue(), "ISTIO_META_ENABLE_HBONE env var not set to 'true' in istio-proxy container")
+						Success("HBONE capability verified in sidecar (ISTIO_META_ENABLE_HBONE=true)")
+					})
```
fjglira, line 368: I think worth to add a step that verify that traffic still flow after HBONE is enable

tests/e2e/migration/migration_procedure_test.go
```diff
@@ -0,0 +1,623 @@
...
+						// Remove revision-based injection label and add ambient label
+						ns := &corev1.Namespace{}
+						Expect(cl.Get(ctx, kube.Key(workloadNamespace), ns)).To(Succeed())
+						delete(ns.Labels, "istio.io/rev")
+						ns.Labels["istio.io/dataplane-mode"] = "ambient"
+						Expect(cl.Update(ctx, ns)).To(Succeed())
+						Success(fmt.Sprintf("Namespace %s switched to ambient mode", workloadNamespace))
+					})
+
+					It("restarts pods to remove sidecars", func(ctx SpecContext) {
```
fjglira:
```suggestion
					It("restarts httpbin pods to remove sidecars", func(ctx SpecContext) {
```

tests/e2e/migration/migration_procedure_test.go
```diff
@@ -0,0 +1,623 @@
...
+						Success("Deployment restart initiated")
+
+						Eventually(func(g Gomega) {
+							_, err := k.WithNamespace(workloadNamespace).RolloutStatus("deployment/httpbin")
+							g.Expect(err).NotTo(HaveOccurred(), "Rollout status check failed")
+						}).Should(Succeed(), "Waiting for deployment rollout to complete")
+						Success("Deployment rollout completed")
+					})
+
+					It("restarts sleep deployment to remove sidecars", func(ctx SpecContext) {
```
fjglira, line 408: In the previous step you used pods and on this deployment, let's keep the same format in both

tests/e2e/migration/migration_procedure_test.go
```diff
@@ -0,0 +1,623 @@
...
+						Expect(err).NotTo(HaveOccurred(), "Failed to restart sleep deployment")
+
+						Eventually(func(g Gomega) {
+							_, err := k.WithNamespace(workloadNamespace).RolloutStatus("deployment/sleep")
+							g.Expect(err).NotTo(HaveOccurred(), "Rollout status check failed")
+						}).Should(Succeed(), "Waiting for sleep deployment rollout to complete")
+						Success("Sleep deployment rollout completed")
+					})
+
+					It("waits for pods to be ready in ambient mode (no sidecar)", func(ctx SpecContext) {
```
fjglira:
```suggestion
					It("waits for pods to be ready in ambient mode without sidecar", func(ctx SpecContext) {
```

tests/e2e/migration/migration_traffic_test.go
```diff
@@ -0,0 +1,535 @@
...
+	errors          []string
+}
+
+var _ = Describe("Migration Traffic Recovery Validation", Ordered,
+	Label("migration", "migration-traffic", "slow"), func() {
+		// Tests migration resilience from sidecar to ambient mode with continuous traffic.
+		// Uses a stable external client (sidecar mode, never migrates) to send continuous traffic
+		// to a server that migrates from sidecar → ambient while traffic flows.
+		// Validates:
+		//   1. Limited disruption during pod restart (≤10 503 errors - brief, acceptable)
```
fjglira: Can we test scaling the target pod to two, that will cause less disruption of the traffic. It make sense for this test?

tests/e2e/migration/migration_traffic_test.go
```diff
@@ -0,0 +1,535 @@
...
+		//   2. Traffic RECOVERS after disruption (≥9/10 consecutive successes post-migration)
+		//   3. Cross-mode communication works (sidecar client → ambient server)
```
fjglira: For me, this is the main goal, verify that coexistence of both mode works, so instad of calling the test migration traffic for me it's a coexistence test with traffic

tests/e2e/migration/migration_coexistence_test.go
```diff
@@ -0,0 +1,535 @@
...
+		// Validates:
+		//   1. Limited disruption during pod restart (≤10 503 errors - brief, acceptable)
+		//   2. Traffic RECOVERS after disruption (≥9/10 consecutive successes post-migration)
+		//   3. Cross-mode communication works (sidecar client → ambient server)
+		SetDefaultEventuallyTimeout(time.Duration(defaultTimeout) * time.Second)
+		SetDefaultEventuallyPollingInterval(time.Second)
+
+		version := istioversion.GetLatestAmbientVersion()
+
+		Context(fmt.Sprintf("Migration testing with Istio %s", version.Version), func() {
```
fjglira, line 57: Change the Context name to the overall objective of the test to differentiate with the other test

tests/e2e/migration/migration_coexistence_test.go
```diff
@@ -0,0 +1,535 @@
...
+							Expect(cl.List(ctx, pods, client.InNamespace(stableClientNamespace), client.MatchingLabels{"app": "sleep"})).To(Succeed())
+							Expect(pods.Items).To(HaveLen(1))
+							clientPod := pods.Items[0].Name
+							GinkgoWriter.Printf("Using stable client pod: %s (in namespace %s)\n", clientPod, stableClientNamespace)
+
+							// Start continuous traffic from stable client → httpbin server
+							// Client stays in sidecar mode, server will migrate to ambient
+							// Background context is used so traffic continues
+							// across multiple It blocks until we explicitly stop it
+							targetService := fmt.Sprintf("httpbin.%s.svc.cluster.local:8000/get", workloadNamespace)
```
fjglira, line 195: This section is good because it does the continuous traffic test but it can be confusing to read. Maybe move and create a helper to test traffic continuously it can be useful for others or future test I think. WDYT?

tests/e2e/migration/migration_coexistence_test.go
```diff
@@ -0,0 +1,535 @@
...
+spec:
+  version: %s
+  namespace: %s
+  targetRef:
+    kind: Istio
+    name: %s`, version.Name, ztunnelNamespace, istioName)
+							Expect(k.CreateFromString(ztunnelYAML)).To(Succeed())
+							common.AwaitCondition(ctx, v1.ZTunnelConditionReady, kube.Key("default"), &v1.ZTunnel{}, k, cl)
+							Success("ZTunnel deployed and ready")
+						})
```
fjglira, line 246: Maybe we can add validation steps to check that traffic continues to flow after each of the step, with this we can catch easier any change that disrupt traffic in case of any failure

tests/e2e/migration/migration_coexistence_test.go
```diff
@@ -0,0 +1,535 @@
...
+								for _, err := range errors {
+									if err == "unexpected status code: 503" && shown < 3 {
+										GinkgoWriter.Printf("  - %s\n", err)
+										shown++
+									}
+								}
+							}
+
+							// Assert: 503 errors should be limited (pod restart is brief)
+							// Allow up to 10 503s (generous for slow OCP pod restart)
```
fjglira, line 372: can we set this with a variable in case we need to adjust it for midstream testing?

tests/e2e/migration/migration_traffic_test.go
```diff
@@ -0,0 +1,535 @@
...
+							Expect(cl.List(ctx, pods, client.InNamespace(stableClientNamespace), client.MatchingLabels{"app": "sleep"})).To(Succeed())
+							Expect(pods.Items).To(HaveLen(1))
+							clientPod := pods.Items[0].Name
+							targetService := fmt.Sprintf("httpbin.%s.svc.cluster.local:8000/get", workloadNamespace)
+
+							GinkgoWriter.Printf("Verifying sustained recovery (10 consecutive requests)...\n")
+							consecutiveSuccesses := 0
+							recoveryFailures := 0
+
+							for i := 0; i < 10; i++ {
```
fjglira: You can use `MustPassRepeatedly` in a step to set how many time needs to pass to be able to accept the step as passed. There is more information in the Ginkgo documentation. https://onsi.github.io/ginkgo/#the-mustpassrepeatedly-decorator

You can check if in any way you can use it to make simpler the validations. I think also it can be passed to an Eventually, but need to check better that

fjglira (commented):
LG in general for me. The only note that I have is all the traffic test it can be confusing to follow and I think we should looks into create some functions than gather repetitive steps to make it easier to go over the entire test

tests/e2e/migration/migration_procedure_test.go
MaxBab: Yeah, good point.
Moved.

tests/e2e/migration/migration_procedure_test.go
MaxBab: Moved the check to a standalone function that called from the tests.

>Can you please check if we do connectivity checks between sleep and httpbin pods.

Why do we need to make a connectivity checks between sleep and httpbin pods?
For that I'm creating a dedicated `workload-migration` NS and pods.

tests/e2e/migration/migration_procedure_test.go
MaxBab: Added

tests/e2e/migration/migration_procedure_test.go
MaxBab: Done

tests/e2e/migration/migration_procedure_test.go
MaxBab: We already have a `common.CreateZTunnel` function.
I missed it.
Moved to use it.

tests/e2e/migration/migration_procedure_test.go
MaxBab: So, only creation of the `ZTunnel` resource makes the transition of the flow to the HBONE, including the workload namespaces?
Is there a way to control the transition to HBONE via the NS?

tests/e2e/migration/migration_procedure_test.go
MaxBab: Added

tests/e2e/migration/migration_procedure_test.go
MaxBab: I'll do it in the followup change, because the deployment change currently hardcoded to `istio-system` NS only.
I don't want to make too many not related changes in this PR.

tests/e2e/migration/migration_traffic_test.go
MaxBab: My main goal is to make sure that the traffic continues to flow after moving from sidecar to ambient.
I'm not focusing on less disruption state here.

tests/e2e/migration/migration_traffic_test.go
MaxBab: Renamed the test and related stuff.

tests/e2e/migration/migration_coexistence_test.go
MaxBab: Renamed.

tests/e2e/migration/migration_coexistence_test.go
MaxBab: Extracted this whole flow into a `common` utils.

tests/e2e/migration/migration_coexistence_test.go
MaxBab: I understand the reason behind it, but I would avoid doing this.
The test is already pretty long and adding traffic validation on each step, would make it endless.

tests/e2e/migration/migration_coexistence_test.go
MaxBab: Converted into variable.

tests/e2e/migration/migration_traffic_test.go
MaxBab: Nice.
I was not aware of this functionality in ginkgo.

Converted to use it.

tests/e2e/migration/migration_procedure_test.go
fjglira: I mena, in others test. I'm think if we can reuse part of this code as a function to use it in a standar way across all the test

tests/e2e/migration/migration_procedure_test.go
fjglira: Activating HBONE in the pods is done by adding the profile to the Istio resource, deploying the Ztunnel, and restarting the pods. This allows users to enable Ambient without doing any actual change in their current deployments

fjglira: So, if you have injection-enable label by default and have already the ambient profile and the Ztunnel deployed the pods are going to get the HBONe enable, there is no way to control this as far I know

MaxBab: I've removed this section.
Indeed, it's not required.

tests/e2e/migration/migration_procedure_test.go
MaxBab: Got it.
I'll look at it as a follow up change.

MaxBab:
@fjglira 

The PR is ready for the final review.

fjglira (approved):
LGTM, let's open issues to avoid forget the enchancements related from discussion

PR #1976, opened 2026-05-28
Fix operator image build failure when using Podman as container runtime

fjglira:
Fix operator image build failure when using Podman as the container runtime. When running inside a Podman VM build container, `docker build .` fails to transfer the build context to Docker Desktop; the workaround uses a minimal temp directory containing only the Dockerfile and the compiled binary, bypassing the context transfer issue.

The existing Docker path (used in CI) is completely unchanged.

### DIFF
--- a/tests/e2e/setup/build-and-push-operator.sh
+++ b/tests/e2e/setup/build-and-push-operator.sh
@@ -77,11 +77,36 @@ build_and_push_operator_image() {
   echo "Docker build flags: ${DOCKER_BUILD_FLAGS}"
   echo "Using image base: ${HUB}/${IMAGE_BASE}:${TAG}"
 
-  BUILD_WITH_CONTAINER=0 \
-    DOCKER_BUILD_FLAGS=${DOCKER_BUILD_FLAGS} \
-    IMAGE=${HUB}/${IMAGE_BASE}:${TAG} \
-    TARGET_ARCH=${TARGET_ARCH} \
-    make docker-push
+  # If using Podman, we need to build the binary locally and then build the image using the local Dockerfile
+  # since Podman doesn't support BuildKit's multi-platform builds with local source context.
+  # For Docker, we keep the existing make target which handles both building the binary and pushing the image in one step.
+  if [[ "${CONTAINER_CLI:-docker}" == "podman" ]]; then
+    repo_root=$(cd "${WD}/../../.." || exit; pwd)
+
+    BUILD_WITH_CONTAINER=0 TARGET_OS=linux TARGET_ARCH=${TARGET_ARCH} \
+      make -C "${repo_root}" build
+
+    tmp_ctx=$(mktemp -d)
+    cp "${repo_root}/Dockerfile" "${tmp_ctx}/Dockerfile"
+    mkdir -p "${tmp_ctx}/out/linux_${TARGET_ARCH}"
+    cp "${repo_root}/out/linux_${TARGET_ARCH}/sail-operator" "${tmp_ctx}/out/linux_${TARGET_ARCH}/sail-operator"
+
+    docker build \
+      ${DOCKER_BUILD_FLAGS} \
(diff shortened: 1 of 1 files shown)


tests/e2e/setup/build-and-push-operator.sh
```diff
@@ -77,11 +77,36 @@ build_and_push_operator_image() {
...
+
+    BUILD_WITH_CONTAINER=0 TARGET_OS=linux TARGET_ARCH=${TARGET_ARCH} \
+      make -C "${repo_root}" build
+
+    tmp_ctx=$(mktemp -d)
+    cp "${repo_root}/Dockerfile" "${tmp_ctx}/Dockerfile"
+    mkdir -p "${tmp_ctx}/out/linux_${TARGET_ARCH}"
+    cp "${repo_root}/out/linux_${TARGET_ARCH}/sail-operator" "${tmp_ctx}/out/linux_${TARGET_ARCH}/sail-operator"
+
+    docker build \
```
nrfox, line 94: Should this be `podman build` or does it not matter?

fjglira: Ohh, the podman configuration that I have already mount docker in the podman socket and allow to execute docker command but it's running at the end podman. Let me check if the behaviour keeps the same (I don't think there is going to be any change)

dgn (commented):
Hey @fjglira is this about `podman + multi-arch` or does it apply to podman on x86 linux as well?

tests/e2e/setup/build-and-push-operator.sh
```diff
@@ -77,11 +77,36 @@ build_and_push_operator_image() {
...
 
-  BUILD_WITH_CONTAINER=0 \
-    DOCKER_BUILD_FLAGS=${DOCKER_BUILD_FLAGS} \
-    IMAGE=${HUB}/${IMAGE_BASE}:${TAG} \
-    TARGET_ARCH=${TARGET_ARCH} \
-    make docker-push
+  # If using Podman, we need to build the binary locally and then build the image using the local Dockerfile
+  # since Podman doesn't support BuildKit's multi-platform builds with local source context.
+  # For Docker, we keep the existing make target which handles both building the binary and pushing the image in one step.
+  if [[ "${CONTAINER_CLI:-docker}" == "podman" ]]; then
```
FilipB, line 83: I thought podman is not working when running in `registry.istio.io/testing/build-tools`? Has something changed in that image?

fjglira: Yeah podman is working with the build-tools image. Most of the failures that we get using podman are privileges or missing configurations and I already added some of those fixes into this repo a while ago. Currently te issue is while building and pushiong but it'0s mostly because we hardcode docker

fjglira:
> Hey @fjglira is this about `podman + multi-arch` or does it apply to podman on x86 linux as well?

It's about podman, it should apply to linux x86 as well. But I asume that almost nobody currently is using podman on this project because seems that I'm the only one hiting this issues xD

tests/e2e/setup/build-and-push-operator.sh
FilipB: It's probably better to explicitly use podman here to avoid confusion

tests/e2e/setup/build-and-push-operator.sh
```diff
@@ -77,11 +77,36 @@ build_and_push_operator_image() {
...
+    repo_root=$(cd "${WD}/../../.." || exit; pwd)
+
+    BUILD_WITH_CONTAINER=0 TARGET_OS=linux TARGET_ARCH=${TARGET_ARCH} \
+      make -C "${repo_root}" build
+
+    tmp_ctx=$(mktemp -d)
```
FilipB, line 89: Why it's necessary to build the image in tmp dir?

PR #1994, opened 2026-06-03
Automate FBC upgrade graph flow during release workflow

### LINKED ISSUE #1982: Automate update of previous version for the OLM upgrade graph in the FBC
Currently we have previous version hardcoded in the [Makefile](https://github.com/istio-ecosystem/sail-operator/blob/main/Makefile.core.mk#L33) and it's easy to forget to update it when a new version is available.

Possible solution: Periodical GH action which will check catalog template as a source of truth for the latest available versions of the operator and it will create a PRs for all affected branches. This will cover cases when e.g. new 1.29.z is added and we have to update release-1.30 branch. This will not cover cases when creating new release. e.g. when creating 1.30.1, previous version must be updated during the release not after by the periodical job. We might want to add new test gating PRs which will check if previous version is up to date so it will prevent adding new version without updating previous version.

MaxBab:
#### What type of PR is this?

- [ ] Bug Fix
- [ ] Refactor
- [X] Optimization
- [ ] Test
- [ ] Documentation Update

During publish of the operator to OperatorHub, "PREVIOUS_VERSION" variable is used to configure upgrade path for FBC. Currently, this variable is maintainer manually and in case of missing update to the relevant version, leads to a gate job failure.

Automate "PREVIOUS_VERSION" variable detection, based on the "VERSION" variable, which is updated as part of the release flow and based on "git tag" detection to catch the latest released version.

Fixes #1982 

Related Issue/PR #

### DIFF
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -22,15 +22,19 @@ OLD_VARS := $(.VARIABLES)
 VERSION ?= 1.30.0
 MINOR_VERSION := $(shell echo "${VERSION}" | cut -f1,2 -d'.')
 
-# This version will be used to generate the OLM upgrade graph in the FBC as a version to be replaced by the new operator version defined in $VERSION.
-# This applies for stable releases, for nightly releases we are getting previous version directly from the FBC.
-# Currently we are pushing the operator to two operator hubs https://github.com/k8s-operatorhub/community-operators and
-# https://github.com/redhat-openshift-ecosystem/community-operators-prod. Nightly builds go only to community-operators-prod which already
-# supports FBC. FBC yaml files and kept in community-operators-prod repo and we only generate a PR with changes using make targets from this Makefile.
-# There are also GH workflows defined to release nightly and stable operators.
-# There is no need to define `replaces` and `skipRange` fields in the CSV as those fields are defined in the FBC and CSV values are ignored.
-# FBC is source of truth for OLM upgrade graph.
-PREVIOUS_VERSION ?= 1.29.1
+# PREVIOUS_VERSION is used to generate the OLM upgrade graph in FBC for stable releases.
+# This variable is auto-detected in hack/operatorhub/publish-bundle.sh by querying the
+# FBC (File-Based Catalog) to get the latest version from the 'stable' channel.
+# The FBC catalog is the authoritative source of truth for OLM upgrade graphs.
+#
+# Auto-detection occurs during 'make bundle-publish' - the script queries:
+# https://raw.githubusercontent.com/redhat-openshift-ecosystem/community-operators-prod/main/operators/sailoperator/catalog-templates/basic.yaml
+#
+# To manually override (if needed):
+#   make bundle-publish -e PREVIOUS_VERSION=X.Y.Z
+#
+# Default: empty (will be auto-detected by publish-bundle.sh)
+PREVIOUS_VERSION ?=
 
 OPERATOR_NAME ?= sailoperator
 VERSIONS_YAML_DIR ?= pkg/istioversion
--- a/hack/operatorhub/publish-bundle.sh
+++ b/hack/operatorhub/publish-bundle.sh
@@ -26,12 +26,6 @@ GIT_USER="${GIT_USER:-}"
 GIT_CONFIG_USER_NAME="${GIT_CONFIG_USER_NAME:-}"
 GIT_CONFIG_USER_EMAIL="${GIT_CONFIG_USER_EMAIL:-}"
 
-# The OPERATOR_NAME is defined in Makefile
-: "${OPERATOR_NAME:?"Missing OPERATOR_NAME variable"}"
-: "${OPERATOR_VERSION:?"Missing OPERATOR_VERSION variable"}"
-: "${CHANNELS:?"Missing CHANNELS variable"}"
-: "${PREVIOUS_VERSION:?"Missing PREVIOUS_VERSION variable"}"
-
 show_help() {
   echo "publish-bundle - raises PR to Operator Hub"
   echo " "
@@ -52,6 +46,7 @@ skipInDryRun() {
   fi
 }
 
+# Parse arguments first so --help works without environment variables
 while test $# -gt 0; do
   case "$1" in
     -h|--help)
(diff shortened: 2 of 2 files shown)


hack/operatorhub/publish-bundle.sh
```diff
@@ -32,6 +32,47 @@ GIT_CONFIG_USER_EMAIL="${GIT_CONFIG_USER_EMAIL:-}"
 : "${CHANNELS:?"Missing CHANNELS variable"}"
 : "${PREVIOUS_VERSION:?"Missing PREVIOUS_VERSION variable"}"
 
+# Validate PREVIOUS_VERSION is a valid semantic version
+if ! validate_semantic_versioning "v${PREVIOUS_VERSION}" 2>/dev/null; then
```
FilipB: Nit: Checking exit value and additional logging is not necessary, see https://github.com/MaxBab/sail-operator/blob/1a2a9aa98e0ac081c5cf78fc07254c7e7f77217f/hack/validate_semver.sh#L35

hack/operatorhub/publish-bundle.sh
```diff
@@ -32,6 +32,47 @@ GIT_CONFIG_USER_EMAIL="${GIT_CONFIG_USER_EMAIL:-}"
...
+
+  if git rev-parse --git-dir > /dev/null 2>&1; then
+    if ! git tag --list | grep -q "^${PREVIOUS_VERSION}$"; then
+      echo "WARNING: PREVIOUS_VERSION (${PREVIOUS_VERSION}) does not exist as a git tag!" >&2
+      echo "  This may indicate the version detection failed or an override was used" >&2
+      echo "  Available recent tags:" >&2
+      git tag --list | grep -E '^[0-9]+\.[0-9]+\.[0-9]+$' | sort -V | tail -5 >&2
+
+      # In CI, this is an error
+      if [ "${CI:-false}" = "true" ]; then
```
FilipB: Where do we specify if we run in CI or not?

Makefile.core.mk
```diff
@@ -22,15 +22,25 @@ OLD_VARS := $(.VARIABLES)
...
 # There are also GH workflows defined to release nightly and stable operators.
 # There is no need to define `replaces` and `skipRange` fields in the CSV as those fields are defined in the FBC and CSV values are ignored.
 # FBC is source of truth for OLM upgrade graph.
-PREVIOUS_VERSION ?= 1.29.1
+#
+# To override: make bundle-publish -e PREVIOUS_VERSION=X.Y.Z
+#
+PREVIOUS_VERSION ?= $(shell \
+	MAJOR=$$(echo "$(VERSION)" | cut -f1 -d'.'); \
+	git tag --list "$${MAJOR}.*.*" 2>/dev/null | sort -V | awk -v ver="$(VERSION)" '$$0 < ver {latest=$$0} END {print latest}')
```
FilipB: `git tag --list` problem with this is that users using `git pull`  will not have all remote tags fetched and it will use local tags which are not up to date. So it will not work when running from local unless users explicitly fetch all tags. I'm not sure what is happening when running in GH action regarding tags

FilipB (changes requested):

hack/operatorhub/publish-bundle.sh
MaxBab: Thx.
Removed.

hack/operatorhub/publish-bundle.sh
MaxBab: I had a logic when added that, but rethinking again, it made no sense to keep it.
Removed it.
Instead, I've slightly modified the logic here.

If the "PREVIOUS_VERSION" var is not found within the "git tag" list, the flow is aborted to prevent a release issue.

Makefile.core.mk
MaxBab: Thanks a lot @FilipB !

You found an issue that for sure would happen.
I've added a command to fetch the git tags.
If execution of this command fails, the flow aborted.

FilipB (approved):

Makefile.core.mk
```diff
@@ -22,15 +22,29 @@ OLD_VARS := $(.VARIABLES)
...
+# supports FBC. FBC yaml files are kept in community-operators-prod repo and we only generate a PR with changes using make targets from this Makefile.
 # There are also GH workflows defined to release nightly and stable operators.
 # There is no need to define `replaces` and `skipRange` fields in the CSV as those fields are defined in the FBC and CSV values are ignored.
 # FBC is source of truth for OLM upgrade graph.
-PREVIOUS_VERSION ?= 1.29.1
+#
+# To override: make bundle-publish -e PREVIOUS_VERSION=X.Y.Z
+#
+PREVIOUS_VERSION ?= $(shell \
+	if ! git fetch --tags >/dev/null 2>&1; then \
```
FilipB: Don't you need to specify remote? If the default remote is your fork, tags will not be there I think

FilipB (changes requested):

Makefile.core.mk
FilipB: Maybe the safest option would be to use https://github.com/redhat-openshift-ecosystem/community-operators-prod as source of truth? And do not specify PREVIOUS_VERSION in the Makefile at all and move the detection to publish_bundle.sh?

MaxBab: I thought of this.
Initially, I set `origin` as remote name.
But then I though that it might not be the name for anyone by default.

Currently, if remote name is not specified, git would use the tracking flow of the branch.
So, if we're running the flow from the `main` branch, it would check to which remote the branch is tied and automatically use it.
I though that it would be enough.

Do you think we still need to specify the remote name as `origin`?

FilipB: I'm not sure how to do this reliably using tags. I think the only source of truth is really the https://github.com/redhat-openshift-ecosystem/community-operators-prod

MaxBab: > I think the only source of truth is really the https://github.com/redhat-openshift-ecosystem/community-operators-prod

Do you mean to search through the `sailoperator` directory within https://github.com/redhat-openshift-ecosystem/community-operators-prod and specify the PREVIOUS_VERSION based on the versions existing there?

FilipB: We can parse this file - https://github.com/redhat-openshift-ecosystem/community-operators-prod/blob/main/operators/sailoperator/catalog-templates/basic.yaml

MaxBab: I've added a check that is done directly with `community-operators-prod` and removed the `PREVIOUS_VERSION` entirely from Makefile.

Makefile.core.mk
```diff
@@ -22,16 +22,6 @@ OLD_VARS := $(.VARIABLES)
...
 
-# This version will be used to generate the OLM upgrade graph in the FBC as a version to be replaced by the new operator version defined in $VERSION.
-# This applies for stable releases, for nightly releases we are getting previous version directly from the FBC.
-# Currently we are pushing the operator to two operator hubs https://github.com/k8s-operatorhub/community-operators and
-# https://github.com/redhat-openshift-ecosystem/community-operators-prod. Nightly builds go only to community-operators-prod which already
-# supports FBC. FBC yaml files and kept in community-operators-prod repo and we only generate a PR with changes using make targets from this Makefile.
-# There are also GH workflows defined to release nightly and stable operators.
-# There is no need to define `replaces` and `skipRange` fields in the CSV as those fields are defined in the FBC and CSV values are ignored.
-# FBC is source of truth for OLM upgrade graph.
-PREVIOUS_VERSION ?= 1.29.1
```
FilipB, line 33: Also please remove this variable from other places in this file

hack/operatorhub/publish-bundle.sh
```diff
@@ -30,7 +30,36 @@ GIT_CONFIG_USER_EMAIL="${GIT_CONFIG_USER_EMAIL:-}"
...
+#
+# The FBC catalog is the authoritative source of truth for OLM upgrade graphs.
+# This queries the 'stable' channel in the community-operators-prod FBC catalog to get the latest released version.
+#
+# To override: export PREVIOUS_VERSION=X.Y.Z before calling this script
+#
+if [ -z "${PREVIOUS_VERSION:-}" ]; then
+  echo "Auto-detecting PREVIOUS_VERSION from FBC catalog..."
+  FBC_URL="https://raw.githubusercontent.com/redhat-openshift-ecosystem/community-operators-prod/main/operators/${OPERATOR_NAME}/catalog-templates/basic.yaml"
+  LATEST=$(curl -sf "${FBC_URL}" 2>/dev/null | yq '.entries[] | select(.schema == "olm.channel" and .name == "stable").entries[-1].name' 2>/dev/null)
```
FilipB: This will work fine for releases of the latest sail version. We don't do releases of older sail versions e.g. 1.29 now but to be safe, we should check if someone is trying to do so and print error or cover this case too. Otherwise it will use the latest available which will be incorrect for e.g. 1.29 version

FilipB (changes requested):

hack/operatorhub/publish-bundle.sh
MaxBab: @FilipB 

I doubt we would ever do that, but...

>but to be safe, we should check if someone is trying to do so

Do you mean to check for open PR on this?
Sounds, like a bit over-complicated effort.

FilipB: No, I mean if you run the publish-bundle.sh from e.g. release-1.29 branch to release older version than the latest

MaxBab: Yeah, that's a valid point.
Will add such a check.

MaxBab: I looked into the logic based on your comment and it's a bit complicated.

In Sail, we do have `1.29` release. And that release even have patch releases. (We're skipped 1.29 release in OSSM).
Now, older releases could and will get patch releases, like `1.28`, so no logic would really be relevant here.

The only think I could check, is the branch from which we're running the `publish_release` flow.
WDYT?

FilipB: Right, it's tricky, that's why we had that variable set manually. I think for now it's sufficient to check that the script runs always only for the latest available version. So I think if you check VERSION  > PREVIOUS_VERSION should be enough.

Makefile.core.mk
MaxBab: The "PREVIOUS_VERSION" variable kept in Makefile only in one place, and it's there to be able to manually set the variable in case it needs to be done by the user.
Otherwise, it would be detected automatically.

hack/operatorhub/publish-bundle.sh
MaxBab: I've added a validation with the following parameters:

* If the version that is added is a new minor version (contains ".0") in the patch section, verify it's older than existsing.
* If the version that is added is a new patch version, validate it's being added to the existing minor release.

Makefile.core.mk
FilipB: Ok. Let's keep it there but we should probably document that variable and set default value? e.g. empty string or `auto`?

hack/operatorhub/publish-bundle.sh
```diff
@@ -30,7 +30,102 @@ GIT_CONFIG_USER_EMAIL="${GIT_CONFIG_USER_EMAIL:-}"
...
+      exit 1
+    fi
+
+    echo "✓ Version check passed: New minor version ${OPERATOR_VERSION} > ${PREVIOUS_VERSION}"
+  else
+    # Rule 2: For patch releases, validate the minor series exists in FBC
+    # It's OK to publish 1.28.4 even when latest is 1.30.0, as long as 1.28 series exists
+    echo "Validating patch release ${OPERATOR_VERSION} for minor series ${OPERATOR_MINOR_VERSION}..."
+
+    FBC_URL="https://raw.githubusercontent.com/redhat-openshift-ecosystem/community-operators-prod/main/operators/${OPERATOR_NAME}/catalog-templates/basic.yaml"
```
FilipB: This is already defined once above.

hack/operatorhub/publish-bundle.sh
FilipB: It's getting complicated but I guess there is no better way.

Makefile.core.mk
MaxBab: Added document explanation to Makefile around the "PREVIOUS_VERSION" value.

hack/operatorhub/publish-bundle.sh
MaxBab: Yes, agree.
We should stop somewhere as this kind of logic validation could be endless.

hack/operatorhub/publish-bundle.sh
MaxBab: Fixed.
Thx.

FilipB (commented):
LGTM, only remaining part is the `yq` comment from copilot which seems to be legit.

MaxBab:
Hi @FilipB 

I've addressed the copilot review as well.
The PR is ready for review.

FilipB (approved):
Thank you!

PR #1999, opened 2026-06-08
[release-1.28] Add 1.28.8 istio version

mkralik3:
SSIA
2 commits:

1. Bump version in `Makefile.core.mk` + add charts to `chart/values.yaml`  + `make gen`
2. `UPDATE_BRANCH=release-1.28 PIN_MINOR=true ./tools/update_deps.sh` + `go get -u istio.io/istio@1.28.4` + `make gen`


go.mod
```diff
@@ -1,8 +1,6 @@
 module github.com/istio-ecosystem/sail-operator
 
-go 1.24.0
-
-toolchain go1.24.1
+go 1.25.5
```
FilipB, line 3: We need to verify it it's ok to bump minor version of go in patch release

Makefile.core.mk
FilipB, line 33: Remember to update PREVIOUS_VERSION too. I'm wondering why it's pointing to 1.28.3

FilipB: Ok, not so relevant for older upstream release branches as we don't release older sail operator versions

mkralik3: yes, for older release branches, we left there previous latest version which was released. For 1.28, the latest released upstream version was 1.28.3

go.mod
mkralik3: downstream already use 1.25 golang builder, if upstream pass, I don't see any issue

FilipB (approved):

PR #2005, opened 2026-06-11
docs: fix inaccurate "ambient mode only supports InPlace" update-strategy note

chinaran:
#### What type of PR is this?
Fixes an inaccurate note in `docs/update-strategy/update-strategy.adoc` that claimed
**only** the `InPlace` strategy is supported in ambient mode.

- [ ] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [ ] Test
- [x] Documentation Update

The note conflated the control plane with the data plane, and its stated reason was
technically wrong:

- The control plane (`istiod`) supports multiple coexisting revisions in ambient mode.
  This is already exercised by the e2e test `tests/e2e/ambient/ambient_update_test.go`
  (`Revision-Based Updates`), where `istiod` and `istiod-canary` run side by side while
  sharing a single ztunnel DaemonSet. No validation/webhook ties `updateStrategy` to
  ambient (`api/v1/istio_types.go` only declares the `InPlace;RevisionBased` enum).
- ztunnel is a per-node DaemonSet, not a "cluster-wide singleton with one instance per
  cluster". Only the ZTunnel custom resource is a singleton (`name: default`). The real
  limitation is that ztunnel/CNI are node-shared components that are always updated in place.
- The old text also contradicted the same document, which states (line 803) that waypoints
  "can function with both revisions during migration" under `RevisionBased`.


MaxBab:
Hi @chinaran 

Could you add the explanation of the PR that is under `What this PR does / why we need it` section to the commit body?
We are willing our commits to contains more info about the changes beside commit subject line.
That's why the `commit-check` gate job failed, BTW.

chinaran:
@MaxBab Done — moved the "What this PR does / why we need it" explanation into the commit body.

cam-garrison (approved):
LGTM - thanks for the fix @chinaran

PR #2026, opened 2026-06-24
test: organization of labels used on e2e test

fjglira:
#### What type of PR is this?

- [ ] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [x] Test
- [ ] Documentation Update

This PR aims to organize the labels used across all the e2e test of the sail operator, so I'm adding and removing some of the current labels used to be able to filter better the test when we need to execute a specific set of test for different kinds of scenarios. Also adding information in the Readme file to give context about the changes, the current status and how can be used

Fixes #

Related Issue/PR #

### DIFF
--- a/tests/e2e/ambient/ambient_test.go
+++ b/tests/e2e/ambient/ambient_test.go
@@ -37,7 +37,7 @@ import (
 
 var defaultTimeout = env.GetInt("DEFAULT_TEST_TIMEOUT", 180)
 
-var _ = Describe("Ambient configuration ", Label("smoke", "ambient"), Ordered, func() {
+var _ = Describe("Ambient configuration ", Label("ambient", "slow"), Ordered, func() {
 	SetDefaultEventuallyTimeout(time.Duration(defaultTimeout) * time.Second)
 	SetDefaultEventuallyPollingInterval(time.Second)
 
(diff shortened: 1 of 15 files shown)


tests/e2e/ambient/ambient_test.go
```diff
@@ -37,7 +37,7 @@ import (
 
 var defaultTimeout = env.GetInt("DEFAULT_TEST_TIMEOUT", 180)
 
-var _ = Describe("Ambient configuration ", Label("smoke", "ambient"), Ordered, func() {
+var _ = Describe("Ambient configuration ", Label("ambient", "slow"), Ordered, func() {
```
MaxBab, line 40: While I understand the reason for the change, I'm looking at the tests and realize that there are actual `slow` tests, like - migration, update, etc..
While Control plane installation, Ambient configuration, are kind in the middle between `smoke` and `slow` tests.
Maybe we could create a middle label and by that separate the really slow test, from middle but functional tests.
And by that create a three level of testing, where the slow tests will be executed last, while middle tests would be a second wave of testing after the smoke.

mkralik3 (approved):

tests/e2e/ambient/ambient_test.go
fjglira: I like this approach. I'll do a small change in a next PR to add this

PR #2028, opened 2026-06-24
enhancements: add SEP for metrics integration

mayleighnmyers:
#### What type of PR is this?

- [ X] Enhancement / New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [ ] Test
- [ ] Documentation Update

Fixes #

Related Issue/PR #
https://github.com/istio-ecosystem/sail-operator/pull/1959


enhancements/SEP3-observability-integration.md
```diff
@@ -0,0 +1,92 @@
...
+## Implementation Plan
+
+v1alpha1
+- [x] Monitoring controller reconciling `ServiceMonitor` and `PodMonitor` for `IstioRevision`
+- [x] `Istio.spec.monitoring.enabled` API
+- [x] Platform-specific PodMonitor relabeling defaults
+- [x] Unit tests for controller and relabeling package
+- [ ] External CRD/API group detection (`monitoring.coreos.com` vs `monitoring.rhobs`)
+- [ ] Kubernetes PodMonitor strategy using `namespaceSelector.any: true`
+- [ ] Validation when monitoring CRDs are unavailable
```
fjglira: Can we add to this document design details for this? From my pov this can be discussed here to set a base. I think we can set a Condition field with `Type: MonitoringAvailable, Status: False, Reason: MissingCRDs, Message: "Prometheus Operator CRDs not found"` something like this will be very helpfull

enhancements/SEP8471-metrics-integration.md
```diff
@@ -0,0 +1,92 @@
...
+
+#### Alternatives API Considered
+
+A broader Observability Integration API with target references to a monitoring and tracing stacks is being discussed in the initial PR comments. We plan to design and implement the monitoring controller and broader Integration API controller separately. This enhancement proposal is focused on the monitoring controller with API change in the `Istio` Custom Resource. We will work toward a broader Integration API design and implementation in next phase.
+
+### Architecture
+
+We assume `ServiceMonitor` and `PodMonitor` CRDs are available under `monitoring.coreos.com/v1` and/or `monitoring.rhobs/v1` group. The Sail Operator ClusterRole grants permissions for both API groups' `ServiceMonitor` and `PodMonitor` custom resources.
+
+A monitoring controller watches `IstioRevision` resources and reconciles `ServiceMonitor` and `PodMonitor` CRs when the boolean field `spec.monitoring.enabled` is set to `true` in the parent `Istio` CR. It also watches namespaces with the `istio-injection=enabled` label and reconciles `PodMonitor` CRs for Istio sidecar injection namespaces.
```
fjglira: This means it will only reconcile the resource for namespaces with stio-injection=enabled`?  What about the revision labels used?

enhancements/SEP8471-metrics-integration.md
```diff
@@ -0,0 +1,92 @@
...
+## Overview
+
+Upstream Istio generates telemetry metrics for the control plane and sidecar proxies. Users can customize Istio metrics with the Telemetry API and create scraping jobs by updating a Prometheus ConfigMap. For example, [prometheus.yaml](https://raw.githubusercontent.com/istio/istio/master/samples/addons/prometheus.yaml). However, that approach requires restarting the Prometheus instance and users may not have enough permission to update the existing Prometheus's ConfigMap.
+
+Alternatively, those can be configured by using two custom resources `ServiceMonitor` and `PodMonitor`. We want to automate the creation of `ServiceMonitor` and `PodMonitor` resources with default values in Sail Operator. And then users can query Istio metrics in Prometheus or Kiali dashboard without manual configuration step.
+
+## Goals
+
+* Provide a monitoring controller that reconciles `ServiceMonitor` and `PodMonitor` resources for Istio control plane and sidecar proxy metrics.
+* Apply platform-appropriate default relabeling rules on Kubernetes and OpenShift.
```
fjglira: Ambient is out of scope?

enhancements/SEP8471-metrics-integration.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Non-goals
+
+* Deploying observability stack components (Prometheus, OpenShift Cluster Observability Operator, OpenShift User-Workload Monitoring, etc.). We assume those are installed separately.
+* Updating `ServiceMonitor` and `PodMonitor` for user customization of scraping paths, ports, or relabeling rules via the Sail Operator API.
```
fjglira, line 22: We will need to explicitly mention this in the docs:
`If users require custom relabeling or scraping configs, they must leave spec.monitoring.enabled: false, and deploy their own independent PodMonitors.`

enhancements/SEP8471-metrics-integration.md
fjglira: If it is out of scope, we should mention this; if not, we need to plan how metrics are gathered from node level ztunnel components and namespace level waypoint proxies

enhancements/SEP8471-metrics-integration.md
```diff
@@ -0,0 +1,92 @@
...
+    enabled: false
+```
+
+This field defaults to `false`. When it is set to `true`, the monitoring controller reconciles `ServiceMonitor` and `PodMonitor` CRs for each owned `IstioRevision`. 
+
+#### Alternatives API Considered
+
+A broader Observability Integration API with target references to a monitoring and tracing stacks is being discussed in the initial PR comments. We plan to design and implement the monitoring controller and broader Integration API controller separately. This enhancement proposal is focused on the monitoring controller with API change in the `Istio` Custom Resource. We will work toward a broader Integration API design and implementation in next phase.
+
+### Architecture
```
fjglira, line 57: How will deletion be handled? I think it worth to mention how we plan to hanlde the deletion of the resources

enhancements/SEP3-observability-integration.md
yxun: Hi @fjglira , yes, I am working on this validation change and having a draft change.
ok, we can add status condition update in the `Istio` custom resources. I will update this detail in SEP later today.

enhancements/SEP8471-metrics-integration.md
yxun: Hi @mayleighnmyers , here is a doc about injection label:
https://istio.io/latest/docs/setup/additional-setup/sidecar-injection/#controlling-the-injection-policy

Currently, we only have the `istio-injection` label watch. There is another label we want to watch. `istio.io/rev` 
when the `istio.io/rev` label has a value , it is same as the `istio-injection=enabled`. 

I can update the SEP part, would you like to add that additional `istio.io/rev` label watch in the monitoring controller code ? thanks.

enhancements/SEP8471-metrics-integration.md
yxun: Right, ambient is out of scope in this SEP. We want to keep changes smaller in each SEP. We can have the next SEP for ambient mode metrics.

enhancements/SEP8471-metrics-integration.md
yxun: There is a concern from existing environments where users have configured their custom ServiceMonitor or PodMontior resources manually. They expect those existing resources should not be deleted from Sail Operator. 
We will add a paragraph about deletion in SEP. Thanks.

enhancements/SEP8471-metrics-integration.md
fjglira: Let's update then the SEP to clearly mention that ambient is out of scope

enhancements/SEP8471-metrics-integration.md
fjglira: Checking ownership can help us to solve this? we can add specific metadata to allow us to easily filter which resources are handle by our operator

enhancements/SEP3-observability-integration.md
```diff
@@ -0,0 +1,142 @@
...
+
+When `spec.monitoring.enabled` is `false`, the monitoring controller does not set this condition.
+
+### Performance Impact
+
+The `ServiceMonitor` reconciliation is only required in the control plane namespace and it has low performance impact. The `PodMonitor` reconciliation watches namespaces with the `istio-injection=enabled` label on OpenShift platform and it may impact performance on clusters when there are large number of existing namespaces.
+
+## Implementation Plan
+
+v1alpha1
```
dgn: what does v1alpha1 mean in this case?

enhancements/SEP8471-metrics-integration.md
dgn, line 1: just a nit on the filename: we have migrated to numbering the SEPs based on the issue that they're about. it appears we don't have an open issue that tracks this work, so that would be the first step, and then we can call this SEP12345 (if 12345 ends up being the issue number)

yxun: I see. This was requested in a JIRA by Nick and we have an epic , tasks. I will update this file name with the JIRA number.

enhancements/SEP3-observability-integration.md
yxun: It refers to API changes we will introduce for enabling this controller. The initial plan was simply adding a `spec.monitoring.enabled` field in the existing Istio CRD so we don't need this `v1alpha1` .

Nick suggested an alternative new API , see his comment in : 
https://github.com/istio-ecosystem/sail-operator/pull/1959#issuecomment-4654948384

If we will introduce a new API CRD such as `Integration` , then we need to put it in a `v1alpha1` version.
We described this alternative API in the section "Alternatives API Considered" above.

That approach will introduce additional discussions about API changes and more general integration controller(s). I feel those need to be in follow up SEPs and this SEP focuses on metric monitoring controller first.  

I will remove this `v1alpha1` here. Thanks.

enhancements/SEP8471-metrics-integration.md
yxun: yes, that helps. Checking labels should also work when the monitoring controller creates those resources with a unique label such as `managed-by`: `mesh-operator`

fjglira (approved):
LGTM, as we discussed in the yesterday's meeting, I'll use some time to take a look into the implementation. Adding a hold label to let @dgn to add any final comment

chinaran:
One question on the generated resources: besides the managed-by: mesh-operator label used for ownership tracking, will the controller let users attach custom labels to the generated ServiceMonitor/PodMonitor?

The reason I ask: a Prometheus CR usually restricts what it scrapes via serviceMonitorSelector/podMonitorSelector (and their namespace-selector counterparts). For example, the widely used kube-prometheus-stack chart defaults to serviceMonitorSelectorNilUsesHelmValues: true, so its Prometheus only picks up monitors labeled release: <release-name>. If the auto-generated monitors don't carry the label(s) the user's Prometheus selector expects, they'll be silently ignored and no Istio metrics get scraped — which defeats the "no manual configuration" goal for exactly those users.

yxun:
> will the controller let users attach custom labels to the generated ServiceMonitor/PodMonitor?

Hello @chinaran , thanks for this concern. 

We tested an integration of kube-prometheus-stack on KinD by setting the `prometheus.prometheusSpec.serviceMonitorSelectorNilUsesHelmValues` and `prometheus.prometheusSpec.podMonitorSelectorNilUsesHelmValues` to false.

Yes, we need to address this issue when setting the `...MonitorSelectorNilUsesHelmValues` to true. 

We may add another string field as the `spec.monitoring.monitored-by` in the Istio CR. Users can set this field for setting Prometheus helm chart release name. And the monitoring controller should automatically map this string field value to a label `release: <helm chart release name>`  when creating monitor resources. I will update this in the SEP.

yxun:
>  will the controller let users attach custom labels to the generated ServiceMonitor/PodMonitor?

@chinaran Is there any case other than the kube-prometheus-stack chart one when users need attach custom labels ? 
We don't expect to attach a long arbitrary list of custom labels in the Istio CR spec. Those are better to be set in their own CRDs or we may introduce a new integration CRD in a future enhancement.

yxun:
Hello @dgn , I have updated the API change according to our last week discussion. The `spec.monitoring.enabled` field type has been updated from boolean to a list of strings in the SEP doc.
When you have time, could you help review this ? Thanks

chinaran:
@yxun Thank you for your reply.
Regarding other use cases—such as the Prometheus CRD from Prometheus Operator—the `spec.podMonitorSelector` and `spec.serviceMonitorSelector` fields support custom labels.
(https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api-reference/api.md#monitoring.coreos.com/v1.Prometheus)
I also agree that configuring this via a new CRD is a better approach. Of course, there is a less elegant alternative: synchronizing the labels from the Istio resources to the `PodMonitor` and `ServiceMonitor` objects.

enhancements/SEP8471-metrics-integration.md
```diff
@@ -0,0 +1,154 @@
...
+
+Upstream Istio generates telemetry metrics for the control plane and sidecar proxies. Users can customize Istio metrics with the Telemetry API and create scraping jobs by updating a Prometheus ConfigMap. For example, The following sample deploys a Prometheus instance and configures Istio telemetry integration in the Prometheus ConfigMap [samples/addons/prometheus.yaml](https://raw.githubusercontent.com/istio/istio/master/samples/addons/prometheus.yaml). However, that approach requires restarting the Prometheus instance and users may not have enough permission to update the existing Prometheus's ConfigMap.
+
+Alternatively, those can be configured by using two custom resources `ServiceMonitor` and `PodMonitor`. We want to automate the creation of `ServiceMonitor` and `PodMonitor` resources with default values in Sail Operator. And then users can query Istio metrics in Prometheus or Kiali dashboard without manual configuration step.
+
+## Goals
+
+* Provide a monitoring controller that reconciles `ServiceMonitor` and `PodMonitor` resources for Istio control plane and sidecar proxy metrics.
+* Apply platform-appropriate default relabeling rules on Kubernetes and OpenShift.
```
nrfox: The Prometheus Operator isn't mentioned here but that is what provides and reconciles the `ServiceMonitor` and `PodMonitor` resources. The upstream prom sample can work with any prometheus instance including those deployed without the Prometheus Operator. This solution requires that you are also running the Prometheus Operator in your cluster.

enhancements/SEP8471-metrics-integration.md
```diff
@@ -0,0 +1,154 @@
...
+* Deploying observability stack components (Prometheus, OpenShift Cluster Observability Operator, OpenShift User-Workload Monitoring, etc.). We assume those are installed separately.
+* Updating `ServiceMonitor` and `PodMonitor` for user customization of scraping paths, ports, or relabeling rules via the Sail Operator API.
+* Observability Distributed Tracing integration.
+* Ambient mode metrics integration is out of scope. Scraping configurations of `ZTunnel`component and Waypoint proxies are non-goals in this SEP.
+
+## Design
+
+### User Stories
+
+1. As a user running Istio with a Prometheus monitoring stack, I want the Sail Operator to configure metrics scraping jobs for Istio-generated telemetry metrics without restart the Prometheus instance.
```
nrfox: iirc the prometheus helm chart includes a hot reloader to automatically pickup changes to the prom config. The istio addon sample doesn't include this but it's not meant for production.

enhancements/SEP8471-metrics-integration.md
```diff
@@ -0,0 +1,154 @@
...
+2. As a user running OpenShift Service Mesh, I want the Sail Operator to create `PodMonitor` resources with platform-appropriate relabeling rules. So those metrics can be formatted and displayed correctly in the OpenShift console and Kiali dashboard.
+
+### API Changes
+
+We will add fields `spec.monitoring.enabled` and `spec.monitoring.monitored-by` in the `Istio` Custom Resource(CR). For example,
```
nrfox: Since ambient and other observability integrations are not included in this SEP, it's not clear to me how the API could grow over time. Non-openshift users will likely want to create a single `PodMonitor` resource and set the `namespaceSelector` field similar to the [upstream prom operator example](https://github.com/istio/istio/blob/208b4ce65dfc6979595d0ee4d4e7f41365a76105/samples/addons/extras/prometheus-operator.yaml#L10-L14). Even some OpenShift users may not be using UWM and won't have the same restrictions on setting this field. There may be other fields that users will want to customize on the `PodMonitor`/`ServiceMonitor` resources that are created by the Sail Operator. Will we need to add these under `spec.monitoring` in the future? Fields that we add to `Istio` are permanent and without knowing how we plan on evolving this part of the API, it's tough to say if we are adding the right fields or not.

Instead of adding fields to `Istio` we could enable this functionality through annotations on the `Istio` resource. This would enable us to move forward with the implementation without committing to the API changes yet. Something like
```
kind: Istio
metadata:
  name: default
  annotations:
    sail.operator.io/montoring: enabled
```

enhancements/SEP8471-metrics-integration.md
yxun: yes, I will add a sentence about relying on Prometheus Operator as a dependency. We will use the the ServiceMonitor and PodMonitor CRDs from Prometheus Operator.

enhancements/SEP8471-metrics-integration.md
yxun: I see. I checked the default UWM Prometheus on OpenShift. It also  doesn't include a hot re-loader for the prom config. This enhancement is targeting the Promethus Operator and  the ServiceMonitor and PodMonitor resources.

enhancements/SEP8471-metrics-integration.md
yxun: I will update this API change section. If we will use an annotation instead, I need to figure out some way of enabling user's custom labels for the Monitor resources.

enhancements/SEP8471-metrics-integration.md
```diff
@@ -0,0 +1,154 @@
...
+
+```yaml
+apiVersion: sailoperator.io/v1
+kind: Istio
+metadata:
+  name: default
+spec:
+  monitoring:
+    enabled: None
```
dgn: what we do in other places is use a `type` field to choose an implementation (see `spec.updateStrategy.type`). Let's stick to that convention and use something like `spec.monitoring.type: None|PrometheusOperator`.

enhancements/SEP8471-metrics-integration.md
```diff
@@ -0,0 +1,154 @@
...
+    monitored-by: prometheus
```
dgn: Let's stick to camelCase monitoredBy.

Also, for `spec.updateStrategy`, when providing additional configuration for a specific type, we use `spec.updateStrategy.<type>.<config>` so in this case this field should go into `spec.updateStrategy.prometheusOperator.monitoredBy`

dgn: btw, are we sure we need this field?

yxun: Yes, we need a field when there are multiple Prometheus instances so that the `ServiceMonitor` or `PodMonitor` need to know which Prometheus scraping job config they are updating.
I'm double checking the label name now. It was came from OpenShift Cluster Observability Monitoring Stack.

Another reason is for integrating with the helm chart kube-prometheus-stack on KinD. 
That came from a comment above in this PR.

```
We may add another string field as the spec.monitoring.monitored-by in the Istio CR. 
Users can set this field for setting Prometheus helm chart release name. 
And the monitoring controller should automatically map this string field value to a label release: <helm chart release name> when creating monitor resources.
```

enhancements/SEP8471-metrics-integration.md
yxun: ok, I will follow that convention and update this section.

yxun:
After review feedback and discussion about this enhancement, we decided starting a simple API change by adding an annotation. 

After initial implementation of a metric integration in Sail Operator, we will revisit users' feedback and solve more use cases with a better API design in the next enhancement. Thanks.

yxun:
Attaching `ServiceMonitor` and `PodMonitor` custom labels for selector use cases is a non-goal in this enhancement. 

We will address common integration use cases in the default label(s) creation such as mapping Prometheus stack helm chart release name to a Monitor resource's `release` label. Other use cases will be added in next enhancement proposal.

enhancements/SEP8471-metrics-integration.md
```diff
@@ -0,0 +1,151 @@
...
+## Design
+
+### User Stories
+
+1. As a user running Istio with a Prometheus production stack, I want the Sail Operator to configure metrics scraping jobs for Istio-generated telemetry metrics.
+2. As a user running OpenShift Service Mesh, I want the Sail Operator to create `PodMonitor` resources with platform-appropriate relabeling rules. So those metrics can be formatted and displayed correctly in the OpenShift console and Kiali dashboard.
+
+### API Changes
+
+We will add an annotation `sail.operator.io/monitoring` in the `Istio` Custom Resource(CR). For example,
```
nrfox: This should follow the format of the existing annotations which should be `sailoperator.io/monitoring` but double check that on a running cluster.

Defined here: https://github.com/istio-ecosystem/sail-operator/blob/82e409da8da9ca18f8559a626c9082bedb22d97f/pkg/constants/constants.go#L19

enhancements/SEP8471-metrics-integration.md
```diff
@@ -0,0 +1,151 @@
...
+
+The **ServiceMonitor** custom resource carries an `ownerReferences` field which points at the owning `IstioRevision` resource. When the `IstioRevision` resource is deleted, Kubernetes garbage collection removes the `ServiceMonitor` resource. 
+
+The **PodMonitor** custom resource does not carry an `ownerReferences` field because cross-namespace owner references are not supported. Those resources are not automatically garbage-collected when the owner `IstioRevision` is deleted. The monitoring controller explicitly handles the cleanup of those `PodMonitor` resources in the following conditions:
+- The owner `IstioRevision` is deleted. 
+- The sidecar injection label(s) are dropped or disabled in application namespaces.
+
+#### Sidecar injection namespace selection
+
+The monitoring controller reconciliation watches namespaces with the Istio sidecar injection labels. For example:
```
nrfox: Are `IstioRevisionTag`s handled as part of the initial implementation?

enhancements/SEP8471-metrics-integration.md
```diff
@@ -0,0 +1,151 @@
...
+
+The controller creates and reconciles `ServiceMonitor` resources named as `{IstioRevision.name}-istiod-metrics` in the control plane namespace. And it creates and reconciles `PodMonitor` resources named as `{IstioRevision.name}-proxies-metrics` in each matching application namespace. It uses Get/Create/Update methods and does not list or bulk-delete monitoring resources in a namespace.
+
+#### Custom Resource Deletion
+
+The **ServiceMonitor** custom resource carries an `ownerReferences` field which points at the owning `IstioRevision` resource. When the `IstioRevision` resource is deleted, Kubernetes garbage collection removes the `ServiceMonitor` resource. 
+
+The **PodMonitor** custom resource does not carry an `ownerReferences` field because cross-namespace owner references are not supported. Those resources are not automatically garbage-collected when the owner `IstioRevision` is deleted. The monitoring controller explicitly handles the cleanup of those `PodMonitor` resources in the following conditions:
```
nrfox: The `IstioRevision` is a cluster scoped resource. I don't remember if you can attach cluster scoped resources as owner refs but it would make handling the cleanup much simpler.

enhancements/SEP8471-metrics-integration.md
```diff
@@ -0,0 +1,151 @@
...
+- The sidecar injection label(s) are dropped or disabled in application namespaces.
+
+#### Sidecar injection namespace selection
+
+The monitoring controller reconciliation watches namespaces with the Istio sidecar injection labels. For example:
+
+| Namespace label | Resolved revision |
+|-----------------|-------------------|
+| `istio-injection: enabled` | `default` |
+| `istio.io/rev: <name>` | `<name>` |
```
nrfox, line 88: This doesn't need to be part of the initial implementation but you can also inject sidecars via a label on the pod itself https://istio.io/latest/docs/reference/config/labels/#SidecarInject. This controller could account for that and create a PodMonitor in namespaces where either the namespace injection label exists or there's a pod with this label.

enhancements/SEP8471-metrics-integration.md
yxun: The monitoring controller is watching both sidecar injection label and istio revision label. 
I will add watching `IstioRevisionTag` CR in this part.

enhancements/SEP8471-metrics-integration.md
yxun: I agree. I have removed this `IstioRevision` condition in the `PodMonitor` cleanup design.
Its cleanup will focus on injection label watches and we will also not cleanup existing monitor resources with users' custom updates after their creation.

nrfox (approved):

dgn (approved):

Issue #1091, opened 2025-08-01
[FEATURE] Support external controlplane setup with lbs with DNS names


nrfox:
### Is this the right place to submit this?

- [x] This is not a question about how to use the sail-operator

### Feature Description

The external controlplane setup described [here](https://github.com/istio-ecosystem/sail-operator/blob/main/docs/deployment-models/multicluster.md#external-control-plane) does not work for lbs with DNS names instead of IPs.

The sail operator remote probe also does not work with discovery addresses with URLs: https://github.com/istio-ecosystem/sail-operator/blob/c07bd705913a8e5b36115eee983509213a4a4ee5/controllers/webhook/webhook_controller.go#L152-L153

Dimss:
@raffamendes @nrfox 
I can confirm, the Istio External Webhook are unreachable when installing Istio Remote if `remotePilotAddress` is set to a DNS name.

When deploying the External Control plane using a [DNS name](https://github.com/istio/istio/blob/master/manifests/charts/istio-control/istio-discovery/templates/remote-istiod-service.yaml#L27) for the `remotePilotAddress`, the remote istiod service in the remote cluster will be set to type: `ExternalName`. As a result, no port mapping will happen for `tcp-webhook` port. However, the `clientConfig` in external istiod webhooks [will continue to point to port 443](https://github.com/istio/istio/blob/master/manifests/charts/istio-control/istio-discovery/templates/mutatingwebhook.yaml#L23), even though the external webhooks will be accessible on 15017

Despite the above, I do not think it is a bug because the Istio documentation clearly states:
* If you are using an IP address instead of the DNS, you must patch the remote-cluster-config 
* This is not recommended in a production environment

In other words: **if you are using `remotePilotAddress` and not DNS with valid certs + `injectionURL`, it is user's responsibility to manually patch the configurations in the way it will suites the needs**. 

There are two options for patching the external deployment:
1. Deploy Istio Gateway for external istiod with `tcp-webhook` port 443, not 15017 -> this will make the whole setup work out of the box 
2. Keep the Istio Gateway for external istiod as is, but patch both validation and mutation webhooks on the remote cluster to use port 15017 instead of 443. 

As for today, the same is true for the Sail Operator because it is fully aligned with the upstream. This means using a non-production `remotePilotAddress` also won't work with the Sail Operator. 

We've two different problems here: 
[...]

PR #2042, opened 2026-06-28
Adding support for DNS based injection URLs in webhook probe check

### LINKED ISSUE #1934: [BUG] Sail Operator does not support  the injectionURL

### Bug Description

Sail Operator does not support for injectionURL when Istio is installed as External-Remote deployment

### Operator Version

1.29.2

### Link to Gist with Logs

_No response_

### Additional Information

_No response_

Dimss:
- [x] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [ ] Test
- [ ] Documentation Update

DNS-based injection URLs (for example, allocated by AWS LB)  will now work correctly in the probe check.

Fixes #1934

Related Issue/PR #
#1091

### DIFF
--- a/controllers/webhook/webhook_controller.go
+++ b/controllers/webhook/webhook_controller.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"net/url"
 	"strconv"
 	"time"
 
@@ -106,8 +107,8 @@ func doProbe(ctx context.Context, webhook *admissionv1.MutatingWebhookConfigurat
 		return false, errors.New("mutatingwebhookconfiguration contains no webhooks")
 	}
 	clientConfig := webhook.Webhooks[0].ClientConfig
-	if clientConfig.Service == nil {
-		return false, errors.New("missing webhooks[].clientConfig.service")
+	if clientConfig.Service == nil && clientConfig.URL == nil {
+		return false, errors.New("missing both webhooks[].clientConfig.service and webhooks[].clientConfig.Url")
 	}
 
 	if len(clientConfig.CABundle) == 0 {
@@ -129,12 +130,12 @@ func doProbe(ctx context.Context, webhook *admissionv1.MutatingWebhookConfigurat
 		},
 	}
 
-	url, err := getReadinessProbeURL(clientConfig)
+	reqUrl, err := getReadinessProbeURL(clientConfig)
 	if err != nil {
 		return false, err
 	}
 
-	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, reqUrl, nil)
 	if err != nil {
 		return false, err
 	}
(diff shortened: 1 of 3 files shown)


controllers/webhook/webhook_controller.go
```diff
@@ -153,7 +154,10 @@ func doProbe(ctx context.Context, webhook *admissionv1.MutatingWebhookConfigurat
 func getReadinessProbeURL(config admissionv1.WebhookClientConfig) (string, error) {
 	switch {
 	case config.URL != nil:
-		return "", errors.New("only webhooks pointing to a Service are supported")
+		if u, err := url.Parse(*config.URL); err == nil && u.Scheme != "" && u.Host != "" {
+			return fmt.Sprintf("%s://%s/ready", u.Scheme, u.Host), nil
```
dgn, line 158: a few issues I see with this approach:
1. we're dropping the path entirely. if the URL is `mydomain.custom/istio/inject` we're rewriting to `mydomain.custom/ready` which does not seem correct
1. we assume blindly that `/ready` is actually exposed. that might not be the case though. especially when the `/inject` endpoint is behind a gateway

dgn (commented):
Thanks for working on this. I don't think we should make assumptions about the URL though (ie that the `/ready` endpoint is available - it often isn't especially when the webhook is behind a gateway). I think a simpler way would be to just send a `GET` request to the actual `/inject` endpoint. If it returns a 4xx error, we're good. If there's a network error, we're not. It's not perfect, but it does at least give us network availability

controllers/webhook/webhook_controller.go
Dimss: Indeed, but I do not think we can get it somewhere else. In External-Remote setup (this is where we mainly  need DNS based mutation webhook URL) The `config.URL` will be something like this `https://DNS-HOST:15017/inject/cluster/cluster2/net/network1`, and it will be always placed behind a Gateway. We can't use `/inject/cluster/cluster2/net/network1` for the health check obviously, and we've no other place where user can define the health check path for the external mutation webhook. 

In addition to that:
1. The probing we do in `webhook_controller.go` does not exists in upstream
2. The sail-operator healthy probe is useful only on the first (create) reconcile, i.e if probe is failing sail-operator won't continue reconciliation loop (no istiod or other related resources will be created). But, if the external webhook become broken after at least one successful reconcile, sail-operator will do nothing with it.

dgn:
hey @Dimss I thought some more about your feedback and I kind of agree that the probes are maybe not the right way. I experimented with a different approach, PTAL: https://github.com/istio-ecosystem/sail-operator/pull/2139

docs/deployment-models/multicluster.adoc
```diff
@@ -1547,3 +1548,349 @@ kubectl delete istios external-istiod --context="${CTX_CLUSTER2}"
 kubectl delete ns external-istiod --context="${CTX_CLUSTER2}"
 kubectl delete ns sample --context="${CTX_CLUSTER2}"
 ----
+
+=== External Control Plane (DNS-based Injection URL)
```
dgn, line 1552: hey, could you extend these steps so they're automatically tested by our docs-tests? thanks!

Dimss: added

dgn (approved):

dgn (changes requested):
tests and lint still failing...

dgn (approved):

dgn (changes requested):
still broken

dgn:
Closing as #2139 added support for DNS-based injection URLs

PR #2048, opened 2026-06-30
Add the correct ciphersuite and ECDH curves on Openshift

rikatz:
#### What type of PR is this?
- [X] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [ ] Test
- [ ] Documentation Update

When fetching TLS configuration from Openshift API, we are transpiling the cipher suite list to meshConfig.

The issue is that Envoy rejects listeners that have min tlsVersion of 1.3 AND establishes a ciphersuite, causing the proxy to never be ready.

This change enforces that when tlsVersion is 1.3, the cipher suite is removed from Envoy/MeshConfig.

Fixes #

Related Issue/PR #

#### Additional information:

Some executed tests:
* Setting TLSProfile as modern, check that just TLS 1.3 is used and that Gateway will provide as the first in the list `X25519MLKEM768` cipher
* Setting TLSProfile as intermediate, both TLS 1.2 and TLS 1.3 are available and the right ciphers are used

### DIFF
--- a/pkg/istiovalues/tls.go
+++ b/pkg/istiovalues/tls.go
@@ -21,6 +21,7 @@ import (
 	"github.com/Masterminds/semver/v3"
 	v1 "github.com/istio-ecosystem/sail-operator/api/v1"
 	"github.com/istio-ecosystem/sail-operator/pkg/config"
+	configv1 "github.com/openshift/api/config/v1"
 
 	"istio.io/istio/pkg/log"
 )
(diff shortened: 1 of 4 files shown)


aslakknutsen:
I wonder if this is the root problem and not needed at all? Unless ztunnel or something else.. 

https://github.com/rikatz/sail-operator/blob/e2695334522b9c7fb4859c285f10911473406f99/pkg/config/tls.go#L127-L132

pkg/istiovalues/tls.go
```diff
@@ -65,6 +74,12 @@ func ApplyTLSConfig(tlsConfig *config.TLSConfig, istioVersion string, values *v1
 		values.MeshConfig.MeshMTLS.MinProtocolVersion = tlsProtocolVersion(tlsConfig.MinVersion)
 	}
 
+	// Envoy does not support setting TLS ciphers when minProtocolVersion is 1.3
+	if values.MeshConfig.TlsDefaults.MinProtocolVersion == v1.MeshConfigTLSConfigTLSProtocolTlsv13 {
+		values.MeshConfig.TlsDefaults.CipherSuites = nil
+	}
+	// TODO: should we reset the MeshMTLS as well? probably
```
dgn: Yeah we should. ZTunnel doesn't (yet?) support configuring ciphersuites anyway

dgn (commented):
Thank you for looking into this! I think we should just remove all cipher_suites when TLS_version>1.3 for now. It doesn't have any effect on any part of the stack anyway.

> The issue is that Envoy rejects listeners that have min tlsVersion of 1.3 AND establishes a ciphersuite, causing the proxy to never be ready.

Looking at the envoy code, this seems to be a problem specific to our OpenSSL backend, because when using boringssl, those ciphers are just silently ignored. For some reason, [the compat code adds another check that verifies that the ciphers are TLS1.2 ciphersuites](https://github.com/envoyproxy/envoy/blob/1609d4614f001c6c66f7b38b54daef831702940c/compat/openssl/source/SSL_CTX_set_strict_cipher_list.cc#L54-L57) (which they are not) and that blocks envoy startup.

dgn (approved):
LGTM, thank you @rikatz

dgn:
ah nevermind there's already #2070

PR #2056, opened 2026-07-01
Capture logs immideately at test failure

MaxBab:
#### What type of PR is this?

- [ ] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [X] Test
- [ ] Documentation Update

Currently, during e2e test failure, logs gathering happens in "AfterAll" section.
But since within the tests, many nested "When" blocks have "BeforeEach" section, when the test fails, ginkgo goes to the next "When" section and executes the "BeforeEach" section there, which leas to a cleanup of different resource, depends on the cleanup steps within specifc section.

That leads to a state, when many gathered logs are empty, since the requested resource already deleted.
That state, makes debugging harded.

Move the logs gathering step to the "JustAfterEach" section. It will capture debug info immediately on test failure (before any cleanup or next BeforeEach runs).

Fixes #

Related Issue/PR #


fjglira (approved):
lgtm

PR #2069, opened 2026-07-03
Fix default tag's VWC Service reference

dgn:
When creating an IstioRevisionTag called `default`, we created a VWC for it, but it would not point to the correct service. This patch fixes that, and at the same time adds some checks in the integration test, and an additional e2e test.

### DIFF
--- a/tests/e2e/controlplane/control_plane_update_test.go
+++ b/tests/e2e/controlplane/control_plane_update_test.go
@@ -32,6 +32,7 @@ import (
 	"github.com/istio-ecosystem/sail-operator/tests/e2e/util/update"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	admissionv1 "k8s.io/api/admissionregistration/v1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


tests/e2e/controlplane/control_plane_update_test.go
```diff
@@ -110,6 +110,21 @@ spec:
...
+				It("istiod-default-validator VWC correctly validates Istio resources", func() {
+					peerAuthYAML := fmt.Sprintf(`
+apiVersion: security.istio.io/v1
+kind: PeerAuthentication
+metadata:
+  name: vwc-test
+  namespace: %s
+spec:
+  mtls:
+    mode: STRICT`, controlPlaneNamespace)
```
sridhargaddam, line 135: If the VWC is broken (i.e., when its pointing to a non-existent service), this would be silently accepted due to "failurePolicy: Ignore". In order to check if the VWC is working, should we actually try creating an istio resource which is invalid (and expect that it fails)?

dgn: Ah, good point. I guess we need to check first whether `failurePolicy: strict` has already been set by istiod, then the positive case should suffice as well

dgn: Actually, I'll test positive and negative

sridhargaddam: Thanks @dgn. I think we should also add a deferCleanup to delete the STRICT PeerAuthentication so that other tests aren't affected.

```
      Expect(k.DeleteFromString(peerAuthYAML)).To(Succeed())
```

dgn: done. thank you @sridhargaddam!

KaelBaldwin:
We're running into this as well, thank you for working on it!

sridhargaddam (approved):
LGTM, thanks @dgn

PR #2139, opened 2026-07-14
webhook controller: remove probing and replace with event monitoring

dgn:
Instead of manually probing the webhook endpoints to determine readiness and health, we should rely on the API server as it is already doing that for us. This rewrite now uses the events emitted by the API server to detect failing webhook endpoints.

### DIFF
--- a/controllers/webhook/webhook_controller.go
+++ b/controllers/webhook/webhook_controller.go
@@ -16,13 +16,9 @@ package webhook
 
 import (
 	"context"
-	"crypto/tls"
-	"crypto/x509"
-	"errors"
-	"fmt"
-	"net"
-	"net/http"
 	"strconv"
+	"strings"
+	"sync"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -33,7 +29,9 @@ import (
 	"github.com/istio-ecosystem/sail-operator/pkg/reconciler"
 	"github.com/istio-ecosystem/sail-operator/pkg/revision"
 	admissionv1 "k8s.io/api/admissionregistration/v1"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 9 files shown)


controllers/webhook/webhook_controller.go
```diff
@@ -190,10 +148,51 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+
+func webhookFailureEventPredicate() predicate.Predicate {
+	return predicate.NewPredicateFuncs(func(obj client.Object) bool {
+		evt, ok := obj.(*corev1.Event)
+		if !ok {
+			return false
+		}
+		return evt.Type == corev1.EventTypeWarning && evt.Reason == "FailedCallingWebhook"
+	})
+}
```
nrfox, line 336: This might be pretty noisy if there's multiple failing webhooks. Can we filter it down more? Using `event.involvedObject`?

controllers/webhook/webhook_controller.go
```diff
@@ -190,10 +148,51 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+// records the failure, and enqueues the webhook for reconciliation.
+func (r *Reconciler) mapFailureEventToWebhook(ctx context.Context, obj client.Object) []reconcile.Request {
+	evt, ok := obj.(*corev1.Event)
+	if !ok {
+		return nil
+	}
+	name := extractWebhookConfigName(evt.Message)
+	if name == "" {
+		return nil
+	}
```
nrfox: Is it possible to use `event.involvedObject` rather than a regex match on the message?

controllers/webhook/webhook_controller.go
```diff
@@ -45,135 +43,95 @@ import (
...
-		if svc.Port != nil {
-			port = int(*svc.Port)
-		}
-		return fmt.Sprintf("https://%s.%s.svc:%d/ready", svc.Name, svc.Namespace, port), nil
-
-	default:
-		return "", errors.New("no URL or Service specified in WebhookClientConfig")
+// isDegraded returns true if a FailedCallingWebhook event was recorded within the recheck period.
+// Failures older than the recheck period are cleared automatically.
+func (r *Reconciler) isDegraded(name string, recheckPeriod time.Duration) bool {
```
nrfox: Does an event only get emitted when the webhook is called? It's a slight behavioral difference over the probe. Does the failure event get cleared from the kube API once the API server can talk to the webhook or will we need to wait until the event expires and gets purged from kube in order to recheck?

controllers/webhook/webhook_controller.go
dgn: I removed the regex and replaced it with something simpler/faster, but we can't use `involvedObject` because that points to the Pod that fails to come up

controllers/webhook/webhook_controller.go
dgn: No event expiry is unrelated and we can't use it.. Basically what I'm doing now is I'm looking at past failures and deduce the back-off time that the ReplicaController is waiting between retries, defaulting to 30s. So if I only see 1 failure event, I wait 30s, and if there's no other failure, I mark the webhook as healthy again. If there's a second event, I double to 60s and so on.

It's not great but I think it's actually more correct than the probing: technically we can't really say whether the webhook is reachable from the API server just because it is from the operator.

controllers/webhook/webhook_controller.go
dgn: Agreed it's going to be a bit noisy, but no easy way to filter (we can't use `involvedObject`, see below). Worst case we're calling `List()` twice (once for Mutating, once for Validating) and that goes straight to the cache so should be fine

controllers/webhook/webhook_controller.go
nrfox: When the webhook fails to get called, the pod doesn't actually get created right? Is it enough to say `event + no pod = isDegraded` and `event + pod = !isDegraded`? Assuming that if there's an event and no pod then the webhook is still failing but if there's an event with a pod then the event is old.

tests/e2e/operator/webhook_event_test.go
```diff
@@ -0,0 +1,162 @@
...
+		extractedName := webhook.ExtractWebhookName(matchedEvent.Message)
+		Expect(extractedName).To(Equal(webhookName),
+			"ExtractWebhookName should extract the webhook name from the real Kubernetes event")
+		Success("Extraction code correctly matched real Kubernetes webhook failure event")
+	})
+})
+
+func strPtr(s string) *string {
+	return &s
+}
```
nrfox: You can just use the builtin `new` now with `go 1.26`.

controllers/webhook/webhook_controller.go
```diff
@@ -45,135 +43,147 @@ import (
...
+// +kubebuilder:rbac:groups=admissionregistration.k8s.io,resources=validatingwebhookconfigurations,verbs=get;list;watch;create;update;patch
+// +kubebuilder:rbac:groups="",resources=events,verbs=get;list;watch
 
-// Reconcile is part of the main kubernetes reconciliation loop which aims to
-// move the current state of the cluster closer to the desired state.
-//
-// For more details, check Reconcile and its Result here:
-// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.14.1/pkg/reconcile
-func (r *Reconciler) Reconcile(ctx context.Context, webhook *admissionv1.MutatingWebhookConfiguration) (ctrl.Result, error) {
+func (r *Reconciler) ReconcileMutating(ctx context.Context, webhook *admissionv1.MutatingWebhookConfiguration) (ctrl.Result, error) {
```
nrfox: When the webhook is not found e.g. the `IstioRevision` got deleted, shouldn't we also clear `failureHistory` for that webhook so as not to build up entries over time?

nrfox (commented):
@dgn just [one comment](https://github.com/istio-ecosystem/sail-operator/pull/2139/changes#r3897468325) otherwise LGTM

tests/e2e/operator/webhook_event_test.go
dgn: actually that didn't really help here. I ended up using `ptr.To()` from `k8s.io/utils/ptr`

controllers/webhook/webhook_controller.go
dgn: I added a janitor ticker that cleans up

controllers/webhook/webhook_controller_test.go
```diff
@@ -53,331 +45,719 @@ var ctx = context.Background()
...
+	g.Expect(r.isDegraded("test")).To(BeZero(), "no failure recorded")
+
+	// Recent failure: still within the degraded window
+	r.recordFailure("test")
+	g.Expect(r.isDegraded("test")).To(BeNumerically(">", 0), "recent single failure")
+
+	// Failure older than the degraded window has aged out
+	r.mu.Lock()
+	r.failureHistory["test"] = time.Now().Add(-(degradedWindow + time.Minute))
+	r.mu.Unlock()
```
nrfox, line 337: nit: something like `r.recordFailureHistory("test", time.Now().Add(-(degradedWindow + time.Minute)))` that handled the locking would move the lock handling out of the tests.

nrfox (approved):

controllers/webhook/webhook_controller_test.go
dgn: thanks, I implemented a separate type `failureHistory` with its own Get(), Add() etc methods that encapsulate the mutex

PR #2151, opened 2026-07-16
Fix race condition between install library reconciler and Uninstall

gcs278:
## Summary

The install library's `Uninstall()` can race with an in-flight `libraryReconciler.Reconcile()`, causing istiod to be reinstalled immediately after being uninstalled. Restores the `lifecycleMu` serialization that was removed in the 3.4 rewrite, and adds back `Reconciling`/`Reconcile complete`/`Uninstalling`/`Uninstall complete` log messages that were also lost.

## Root Cause

The 3.4 rewrite (https://github.com/istio-ecosystem/sail-operator/pull/1857) replaced the workqueue-based lifecycle (`lifecycle.go`) with a controller-runtime manager (`reconciler.go`). This removed the `lifecycleMu` mutex that previously serialized `Apply` and `Uninstall` with the reconciliation loop.

Without it, the following race occurs:

1. `libraryReconciler.Reconcile()` copies `desiredOpts` and starts a Helm install
2. The consumer calls `Uninstall()`, which nils `desiredOpts` and runs Helm uninstall
3. The in-flight reconciler finishes its Helm install **after** the uninstall, reinstalling istiod

```mermaid
sequenceDiagram
    participant R as libraryReconciler.Reconcile()
    participant U as Uninstall()
    participant H as Helm State

    R->>R: opts = desiredOpts (copied)
    R->>H: inst.reconcile(opts) starts...
    Note over R,H: Helm install in progress

    U->>U: desiredOpts = nil
    U->>H: inst.uninstall()
    H-->>U: istiod deleted
    U->>U: return nil (success)

    H-->>R: inst.reconcile(opts) finishes
    Note over R,H: istiod reinstalled!
```

The consumer sees a successful `Uninstall()` return, but istiod is still running.

The previous implementation in `lifecycle.go` prevented this with `lifecycleMu`, which serialized `Apply`, `Uninstall`, and the processing loop. It also stopped informers and drained the work queue before running the Helm uninstall.

**Previous implementation (release-3.3):**
- [`lifecycleMu` field](https://github.com/openshift-service-mesh/sail-operator/blob/release-3.3/pkg/install/library.go#L198)
- [`Apply` lock](https://github.com/openshift-service-mesh/sail-operator/blob/release-3.3/pkg/install/lifecycle.go#L63-L64)
- [`Uninstall` lock + informer drain](https://github.com/openshift-service-mesh/sail-operator/blob/release-3.3/pkg/install/lifecycle.go#L132-L168)
- [Reconcile/Uninstall log messages](https://github.com/openshift-service-mesh/sail-operator/blob/release-3.3/pkg/install/lifecycle.go#L246-L249)

## Fix

Restore `lifecycleMu` on the `Library` struct. Hold it in:
- `Apply()` — while setting `desiredOpts` and sending the trigger
- `Uninstall()` — across the entire Helm uninstall
- `libraryReconciler.Reconcile()` — across the Helm reconcile

This ensures `Uninstall` cannot run while a reconcile is in progress, and vice versa.

## Reproduction

Rapidly creating and deleting a GatewayClass while the library's reconciler is mid-install consistently reproduces the race. In OpenShift CI, the failure manifested as a flaky test where istiod pods remained after GatewayClass deletion, with the Helm release secret still present. This caused the 3.4 vendor bump to be reverted: https://github.com/openshift/cluster-ingress-operator/pull/1507

Confirmed via debug logging:
```
libraryReconciler.Reconcile: starting Helm reconcile     ← reconciler grabs opts, starts install
Library.Uninstall: starting Helm uninstall               ← CIO calls Uninstall, nils desiredOpts
libraryReconciler.Reconcile: finished Helm reconcile     ← reconciler finishes AFTER, reinstalls
```

### DIFF
--- a/pkg/install/library.go
+++ b/pkg/install/library.go
@@ -33,6 +33,8 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/event"
+
+	"istio.io/istio/pkg/log"
 )
 
 const (
@@ -156,8 +158,9 @@ type Status struct {
 // Library manages the lifecycle of an istiod installation without
 // requiring the full Sail Operator to be deployed.
 type Library struct {
-	mu       sync.Mutex
-	statusMu sync.RWMutex
+	mu          sync.Mutex
+	lifecycleMu sync.Mutex
+	statusMu    sync.RWMutex
 
 	kubeConfig             *rest.Config
 	resourceFS             fs.FS
@@ -259,6 +262,8 @@ func (l *Library) Apply(opts Options) error {
 	if err := ValidateOptions(opts); err != nil {
 		return fmt.Errorf("invalid options: %w", err)
 	}
+	l.lifecycleMu.Lock()
+	defer l.lifecycleMu.Unlock()
 	l.mu.Lock()
 	if l.desiredOpts != nil && optionsEqual(*l.desiredOpts, opts) {
 		l.mu.Unlock()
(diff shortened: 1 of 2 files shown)


gcs278:
I'm still testing that this properly fixes our problem, but ready for review.
/hold

dgn (approved):
Thank you! 🙏

dgn:
I will add some tests to cover this and other scenarios in a follow-up

nrfox (approved):

gcs278:
I got 20 successful CI runs, and my local testing showed that this fix a rapid install/uninstall reproducer. I think it's tested enough.
/unhold

PR #2156, opened 2026-07-17
Enable native nftables by default on OpenShift

sridhargaddam:
On OpenShift, Sail Operator now sets global.nativeNftables=true
for Istio and IstioCNI unless the user explicitly overrides it. Other
platforms keep the existing chart default (disabled).

This avoids requiring every OpenShift install to opt in manually,
while still honoring spec.values.global.nativeNftables=false when set.

**Alternatives Considered:**
1. Update Istio's profile-platform-openshift.yaml

One option is to enable `global.nativeNftables=true` in Istio's OpenShift platform profile so that Helm and istioctl use the same default. My main concern with this approach is that it would affect every OpenShift installation, including older clusters where the host might not have a usable nft binary. Keeping this default in Sail limits the change to operator-managed installations, while still leaving the door open for an upstream change in the future if we decide to go in that direction.

2. Midstream-only change in openshift-service-mesh/sail-operator

Another option is to make this a midstream-only change so that only the productized builds enable nftables by default. The downside is that community Sail users on OpenShift would still need to opt in manually, creating an unnecessary difference in behavior between upstream and midstream. Setting the default in the community Sail operator keeps the experience consistent for OpenShift users and avoids maintaining a downstream-only change.

### DIFF
--- a/docs/common/istio-nftables.adoc
+++ b/docs/common/istio-nftables.adoc
@@ -33,11 +33,14 @@ Many major Linux distributions are actively moving towards adopting native nftab
 === Installation
 
 The support for native nftables in Istio sidecar mode was implemented in the upstream istio https://istio.io/latest/news/releases/1.27.x/announcing-1.27[release-1.27].
-It is disabled by default. To enable it, you can set a feature flag as `+values.global.nativeNftables=true+`.
+On Kubernetes platforms other than OpenShift, it is disabled by default. To enable it, set `+values.global.nativeNftables=true+`.
+
+On OpenShift, the Sail Operator automatically sets `+global.nativeNftables=true+` for both `+Istio+` and `+IstioCNI+` resources.
+If you explicitly set `+spec.values.global.nativeNftables=false+`, that value is honored and overrides the OpenShift default.
 
 ==== Install in Sidecar Mode
 
-When installing `Istio` and `IstioCNI` resources with the Sail Operator, you can enable nftables by setting `spec.values.global.nativeNftables=true` in the resource. This option configures Istio to use the nftables backend for traffic redirection instead of iptables.
+When installing `Istio` and `IstioCNI` resources with the Sail Operator on non-OpenShift platforms, you can enable nftables by setting `spec.values.global.nativeNftables=true` in the resource. This option configures Istio to use the nftables backend for traffic redirection instead of iptables. On OpenShift, this setting is applied automatically unless you override it.
 
 . Create the `+istio-system+` and `+istio-cni+` namespaces.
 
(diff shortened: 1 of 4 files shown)


dgn:
> **Alternatives Considered:**
> 
>     1. Update Istio's profile-platform-openshift.yaml
> 
> 
> One option is to enable `global.nativeNftables=true` in Istio's OpenShift platform profile so that Helm and istioctl use the same default. My main concern with this approach is that it would affect every OpenShift installation, including older clusters where the host might not have a usable nft binary. Keeping this default in Sail limits the change to operator-managed installations, while still leaving the door open for an upstream change in the future if we decide to go in that direction.

Didn't we implement it so that it falls back to iptables in that case?

I think making this change in the upstream chart's profile is cleaner. The only rationale for me to do it in the Sail Operator would be to include an OCP version check, but we're not doing that in this PR.

sridhargaddam:
> Didn't we implement it so that it falls back to iptables in that case?

I'm afraid not. It falls back to iptables only in ambient mode, and only in the following scenarios:

1. During upgrades, when the CNI node agent detects existing iptables artifacts on the host network. In this case, it falls back to iptables and [logs a message](https://github.com/istio/istio/blob/77bfdc743714c3565c181c9b1d842af4c77fa931/cni/pkg/nodeagent/server_linux.go#L61) such as: "To complete migration to nftables, reboot the node."
2. When the `nft` binary is [built without](https://github.com/istio/istio/blob/77bfdc743714c3565c181c9b1d842af4c77fa931/cni/pkg/nodeagent/server_linux.go#L79) `libjansson` support.

In all other cases (f.e., when the `nft` binary is missing), and always in sidecar mode, the `nft` binary is treated as a hard requirement. In those cases, it [logs an error such as](https://github.com/istio/istio/blob/77bfdc743714c3565c181c9b1d842af4c77fa931/tools/istio-nftables/pkg/capture/run.go#L56): "nativeNftables is enabled, but the nft binary is missing in the sidecar container image (or on the host)."

This behavior is intentional. The goal is to avoid silently falling back to iptables when users have explicitly configured nftables mode. A silent fallback could give the impression that everything is working with nftables when it is actually running on iptables. That is especially important for users who are validating or migrating their platforms to nftables and need a clear indication that their environment is fully ready for nftables.

> I think making this change in the upstream chart's profile is cleaner. The only rationale for me to do it in the Sail Operator would be to include an OCP version check, but we're not doing that in this PR.

My thought behind this PR is that the OSSM and Sail Operators target specific OCP versions. For example, OSSM 3.4 is supported only on OCP 4.20, 4.21, and 4.22. Since these platforms include the `nft` binary, we know it is safe to enable `nativeNftables` on them.

sridhargaddam:
> I think making this change in the upstream chart's profile is cleaner. 

@dgn, I thought about this further, and as you mentioned, it is probably better to update the OpenShift profiles in Istio to enable nftables by default.

When `nativeNftables` is set to `true`, the `nft` binary is expected to be available in the following places:

1. **Sidecar + init container:** `proxyv2` container image
2. **Sidecar + CNI:** Host filesystem
3. **Ambient mode:** Istio CNI pod (i.e., `install-cni` image)

Istio images such as `proxyv2` and Istio CNI already include the `nft` binary for both distroless as well as debug images. The host filesystem, such as RHEL 9, also includes `nft`, so we should be fine enabling nftables by default.

In the worst-case scenario, if the `nft` binary is missing from the host, users will still have the option to disable `nativeNftables` through the Helm configuration.

Any thoughts before I start working on the necessary changes in Istio to set `nativeNftables` to `true` by default for the OpenShift profiles?

dgn:
@sridhargaddam should we close this issue?

sridhargaddam:
> @sridhargaddam should we close this issue?

Yup. I'm working on the necessary changes in Istio.

PR #2174, opened 2026-07-23
Add integrations SEP

nrfox:
Add a SEP that aims to simplify integrating with `Istio`. Note that this builds on the work being done for the [monitoring SEP](github.com/istio-ecosystem/sail-operator/pull/1959). It does not replace it.

 

#### What type of PR is this?

- [ ] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [ ] Test
- [ ] Documentation Update

Fixes #

Related Issue/PR #

### DIFF
--- a/enhancements/SEP3-integrations.md
+++ b/enhancements/SEP3-integrations.md
@@ -386,15 +386,15 @@ Some users will already have configured their integrations. They may already hav
 
 The integrations controller could either:
 
-1. Adopt the resources i.e. add the Integrations controller's OwnerRef to them. Adopting the resources is the best user experience but potentially the most problematic since these resources may have customizations or be managed by another system.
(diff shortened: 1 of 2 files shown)


dgn:
Notes from our discussion
* Need to investigate how GitOps tools play with SSA - will they go into a tightloop if we e.g. make changes to a gitops-controlled `Kiali` CR? 
* We should list all possible integrations (SPIRE, CertManager, Tracing, Metrics, Kiali), find possible categories and try to find out where to draw the line between CRDs (e.g. one CRD per category vs 1 or 2 CRDs in total)

dgn (approved):
I think this looks good! We can still update as we go. I'd probably implement the controllers separately (one per CRD)

PR #2177, opened 2026-07-23
Add nil pointer check in ApplyTLSConfig func

### LINKED ISSUE #2096: [BUG] Nil-pointer panic on version-parse failure in TLS values
In `ApplyTLSConfig`, a failed `semver.NewVersion()` logs a warning but then dereferences the nil `v` pointer. Unlike the sibling code in `fips.go` (which checks `v != nil`), this path panics. Currently unreachable because callers pass resolved version names, but it is one refactor away from a crash-loop on OpenShift+TLS-profile clusters.

https://github.com/istio-ecosystem/sail-operator/blob/86fac7a0a15c7e48014af334974ec3e25d6ed276/pkg/istiovalues/tls.go#L94-L105

Compare with the safe version in fips.go:

https://github.com/istio-ecosystem/sail-operator/blob/86fac7a0a15c7e48014af334974ec3e25d6ed276/pkg/istiovalues/fips.go#L79-L88

## Suggested Fix

`if v != nil && v.GreaterThanEqual(istio1_29)` with an explicit else-log.

---

MaxBab:
#### What type of PR is this?

- [X] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [ ] Test
- [ ] Documentation Update

The ApplyTLSConfig function needs to check for version value nil pointer to avoid false path check failure.

Fixes #2096 

Related Issue/PR #

### DIFF
--- a/pkg/istiovalues/tls.go
+++ b/pkg/istiovalues/tls.go
@@ -93,8 +93,9 @@ func ApplyTLSConfig(tlsConfig *config.TLSConfig, istioVersion string, values *v1
 
 	if minVersionName := tlsVersionName(tlsConfig.MinVersion); minVersionName != "" {
 		v, err := semver.NewVersion(istioVersion)
-		if err != nil {
+		if err != nil || v == nil {
 			log.Warnf("failed to parse Istio version %q: %v", istioVersion, err)
+			return
 		}
 
 		// This flag is only supported on Istio 1.29+. TODO: Remove this check when we drop support for Istio 1.28


pkg/istiovalues/tls.go
```diff
@@ -93,8 +93,9 @@ func ApplyTLSConfig(tlsConfig *config.TLSConfig, istioVersion string, values *v1
 
 	if minVersionName := tlsVersionName(tlsConfig.MinVersion); minVersionName != "" {
 		v, err := semver.NewVersion(istioVersion)
-		if err != nil {
+		if err != nil || v == nil {
```
nrfox, line 96: nit: Usually you don't need to check if the value is nil. Most `New` functions won't return nil for the value if `err == nil`. But it also doesn't hurt.

nrfox (approved):

PR #2187, opened 2026-07-28
fix: apply WebhookFilter to webhook watches

dgn:
This avoids unnecessary reconciliations by ignoring changes to fields we know istiod sets after installation.

### DIFF
--- a/tests/integration/api/istiorevisiontag_test.go
+++ b/tests/integration/api/istiorevisiontag_test.go
@@ -165,6 +165,48 @@ var _ = Describe("IstioRevisionTag resource", Label("istiorevisiontag"), Ordered
 						Expect(webhook.Webhooks[0].ClientConfig.Service.Namespace).To(Equal(istio.Spec.Namespace),
 							"VWC should point to the correct namespace for the referenced revision")
 					})
+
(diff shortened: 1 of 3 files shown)


tests/integration/api/istiorevisiontag_test.go
```diff
@@ -165,6 +165,48 @@ var _ = Describe("IstioRevisionTag resource", Label("istiorevisiontag"), Ordered
 						Expect(webhook.Webhooks[0].ClientConfig.Service.Namespace).To(Equal(istio.Spec.Namespace),
 							"VWC should point to the correct namespace for the referenced revision")
 					})
+
+					It("skips reconcile when only caBundle and failurePolicy are updated on ValidatingWebhookConfiguration", func() {
+						time.Sleep(5 * time.Second)
```
nrfox, line 170: This is for in flight reconciles to settle?

nrfox (approved):

tests/integration/api/istiorevisiontag_test.go
dgn: yup, this mirrors what we do in the istiorevision controller tests

PR #2195, opened 2026-07-30
feat: support ZTunnel vendor defaults and add v1.30.3 ztunnel resource limits

chinaran:
## Why

Per the resource specification, every istio component must define resource limits. In an ambient-mode deployment, ztunnel was the only component shipped without limits:

| Component | Limits | Source |
|---|---|---|
| istiod | 2 / 4Gi | vendor_defaults.yaml |
| istio-cni-node | 1 / 1Gi | vendor_defaults.yaml |
| waypoint | 2 / 1Gi | chart default |
| ztunnel | **none** | — |

Root cause: the vendor defaults mechanism only supported the `Istio` and `IstioCNI` resources (`vendor_defaults.go`, "current valid resources are: istio and istiocni"), so ztunnel had no vendor-level channel to receive default limits.

## What

- Add `ApplyZTunnelVendorDefaults` to `pkg/istiovalues/vendor_defaults.go`, following the same pattern used when IstioCNI support was added (istio-ecosystem#845).
- Wire it into `ZTunnelReconciler.ComputeValues` (after image digests, before FIPS values), mirroring the CNI reconciler.
- Set default resource limits for **v1.30.3 only** in `vendor_defaults.yaml`:
  ```yaml
  ztunnel:
    ztunnel:
      resources:
        limits:
          cpu: "2000m"
          memory: 1024Mi
  ```
  The values match the default limits of the other data plane proxies (sidecar `2000m/1024Mi`, waypoint `2/1Gi`). Older versions (v1.28.x) are intentionally left unchanged.
- Merge semantics: vendor defaults are the base and user values are the override, so any `spec.values.ztunnel.resources` set explicitly in a ZTunnel CR still takes precedence.

## Testing

- New `TestApplyZTunnelVendorDefaults` unit test: basic merge, user-override deep-merge (user cpu wins, vendor memory fills in), version without ztunnel defaults, malformed defaults error.
- Extended `TestValidateVendorDefaultsFile` to type-check the `istio`, `istiocni` **and** `ztunnel` sections of `vendor_defaults.yaml` (previously only `istio` was validated).
- `make build`, `golangci-lint` (0 issues), and the full unit test suite pass locally. (`lint-scripts` / `lint-copyright-banner` fail on `main` as well due to pre-existing `.claude/skills` shell scripts — unrelated to this change.)


chinaran:
Opened against the wrong repository by mistake (this change targets the alauda-mesh fork). Closing, sorry for the noise.

PR #2217, opened 2026-08-04
Use `fips-140-3` compliance policy

nrfox:
#### What type of PR is this?

- [ ] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [ ] Test
- [ ] Documentation Update

When Istio version > 1.30, `fips-140-3` policy should be used since Istio 1.31 is built with go1.26 and go's FIPS mode is certified on platforms that have openssl crypto backends.

Fixes #

Related Issue/PR #

### DIFF
--- a/pkg/istiovalues/fips.go
+++ b/pkg/istiovalues/fips.go
@@ -15,8 +15,9 @@
 package istiovalues
 
 import (
-	"crypto/fips140"
 	"fmt"
+	"os"
+	"strings"
 
 	"github.com/Masterminds/semver/v3"
 	v1 "github.com/istio-ecosystem/sail-operator/api/v1"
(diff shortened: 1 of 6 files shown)


pkg/istiovalues/fips.go
```diff
@@ -15,64 +15,63 @@
...
 )
 
 var istio1_30 = semver.MustParse("1.30.0")
 
-// detectFipsMode checks if FIPS mode is enabled in the system.
-func init() {
-	detectFipsMode(FipsEnableFilePath)
-}
+// This is separated out solely to let tests override it.
+var fipsEnabled = fips140.Enabled
```
dgn: hmm. I'm not sure this is correct. Aren't we setting COMPLIANCE_POLICY now always, because FIPS in Go is enabled at build time? and won't that break the other TLS settings work?

nrfox: I'll revert this change. I don't think there's support for autodetecting FIPS from the hardware in upstream golang. Even if there was it shouldn't be included with this change.

nrfox: Reverted this part of the PR. Now this PR only is concerned with conditionally setting `fips-140-3`.

dgn (approved):

PR #2258, opened 2026-08-17
Add 1.31.0-beta.1 charts

bmangoen:
Add 1.31.0-beta.1 charts for early testing phase


pkg/istioversion/versions.yaml
```diff
@@ -15,6 +15,28 @@
...
+    version: 1.31.0-beta.1
+    repo: https://github.com/istio/istio
+    commit: 1.31.0-beta.1
+    charts:
+      - https://istio-release.storage.googleapis.com/charts/base-1.31.0-beta.1.tgz
+      - https://istio-release.storage.googleapis.com/charts/istiod-1.31.0-beta.1.tgz
+      - https://istio-release.storage.googleapis.com/charts/gateway-1.31.0-beta.1.tgz
+      - https://istio-release.storage.googleapis.com/charts/cni-1.31.0-beta.1.tgz
+      - https://istio-release.storage.googleapis.com/charts/ztunnel-1.31.0-beta.1.tgz
+  - name: v1.30.2
```
unsortedhashsets: all that block for v1.30.2 seems redundant with same on L52 and also located above 1.30.3 so suppose can be deleted, and hope it will also resolve unnit test 👍

unsortedhashsets (changes requested):

pkg/istioversion/versions.yaml
bmangoen: Good catch, thanks !

unsortedhashsets (approved):
LGTM

MaxBab (approved):

PR #2263, opened 2026-08-18
Enable NetworkPolicy defaults for OCP 5+ clusters

cam-garrison:
Bringing https://github.com/openshift-service-mesh/sail-operator/pull/970 to upstream, as there are API changes in the PR. 

Jira:
https://redhat.atlassian.net/browse/OSSM-14869

Detect OCP version at startup by querying the ClusterVersion API (config.openshift.io/v1) and store it in ReconcilerConfig
On OCP 5+, automatically set global.networkPolicy.enabled
On OCP 4 (or non-OpenShift), preserve existing opt-in behavior, no NetworkPolicies unless explicitly enabled
Users can always override by setting spec.values.global.networkPolicy.enabled on any CR
Manual testing:

Tested on an OCP 4.23 nightly cluster
OCP 4 (real cluster, detected major=4 minor=23):

Deployed Istio, IstioCNI, and ZTunnel CRs with no networkPolicy setting — no NetworkPolicies created (correct)
Patched all three CRs with networkPolicy.enabled: true — NPs created for istiod, istio-cni, ztunnel (correct)
Removed the networkPolicy setting — NPs deleted (correct)
OCP 5 (spoofed by hardcoding major=5 minor=0 on same cluster):

Deployed same CRs with no networkPolicy setting — NPs auto-created for all three components (correct)
Patched all three CRs with networkPolicy.enabled: false — NPs removed (correct)


dgn (commented):
This generally LGTM. Do we need e2e tests for it?

I'm a little worried that people's meshes will stop working with the upgrade to OCP5, because of this.

cam-garrison:
> This generally LGTM. Do we need e2e tests for it?
> 
> I'm a little worried that people's meshes will stop working with the upgrade to OCP5, because of this.

What do you think @dgn, should we consider only auto-enabling NPs for fresh installs on OCP 5 (not upgrades)? if someone has workloads that need non-standard access istiod/ztunnel/cni (custom ports, unusual network paths), the new NPs could block them unexpectedly during an OCP 4 to 5 upgrade.

 existing explicit false settings should be preserved, so we could list this as a release note/warning in the related z-streams.

dgn:
> > This generally LGTM. Do we need e2e tests for it?
> > I'm a little worried that people's meshes will stop working with the upgrade to OCP5, because of this.
> 
> What do you think @dgn, should we consider only auto-enabling NPs for fresh installs on OCP 5 (not upgrades)? if someone has workloads that need non-standard access istiod/ztunnel/cni (custom ports, unusual network paths), the new NPs could block them unexpectedly during an OCP 4 to 5 upgrade.
> 
> existing explicit false settings should be preserved, so we could list this as a release note/warning in the related z-streams.

Applying the behavior change only on fresh installs makes sense, I think we should go with that. Maybe we just set it as a default value in mutating webhook?

cam-garrison:
hey @dgn, reworked the PR - PTAL when you get a chance! ty

cam-garrison:
@nrfox could you take a look at this when you're able? tysm

cmd/main.go
```diff
@@ -150,14 +150,21 @@ func main() {
...
 			setupLog.Error(err, "unable to create temporary client")
 			os.Exit(1)
 		}
 
+		reconcilerCfg.OCPVersion, err = config.DetectOCPVersion(ctx, cl)
+		if err != nil {
+			setupLog.Error(err, "unable to detect OCP version, NetworkPolicy defaults will be disabled")
+		} else {
+			setupLog.Info("detected OCP version", "major", reconcilerCfg.OCPVersion.Major, "minor", reconcilerCfg.OCPVersion.Minor)
+		}
```
nrfox, line 166: Thinking out loud here but what happens if this changes from when we initially detected it? e.g.

User is on 4.22
Starts upgrade to 5.0
Operator reconciles Istio with stale 4.22 as the desired version

Since we only update the defaults for new instances on 5.0, I think this would only be an issue if you create a new `Istio` and your operator pod hasn't been restarted as part of the upgrade process from 4.22 --> 5.0. Assuming that the operator pod will get restarted as part of the upgrade process.

controllers/istiocni/istiocni_controller.go
```diff
@@ -120,6 +120,7 @@ func (r *Reconciler) newCNIReconciler() *sharedreconcile.CNIReconciler {
 	return sharedreconcile.NewCNIReconciler(sharedreconcile.Config{
 		ResourceFS:        r.Config.ResourceFS,
 		Platform:          r.Config.Platform,
+		OCPVersion:        r.Config.OCPVersion,
```
nrfox, line 123: Plumbing OCP specific values through isn't ideal but I'm not sure of a better alternative. At least we will only need this for operator versions that support both 4.2x and 5.x?

You can put a `TODO` to remove after.

pkg/reconcile/ztunnel.go
```diff
@@ -70,7 +70,9 @@ func (r *ZTunnelReconciler) Validate(ctx context.Context, version, namespace str
 // ComputeValues computes the final Helm values by applying digests, profiles, and user overrides.
 // If baseValues are provided (e.g. from a referenced IstioRevision), they are treated like an additional
 // profile layer: applied on top of profile defaults, with user values then applied on top.
-func (r *ZTunnelReconciler) ComputeValues(version string, userValues *v1.ZTunnelValues, baseValues ...helm.Values) (helm.Values, error) {
+func (r *ZTunnelReconciler) ComputeValues(
+	version string, userValues *v1.ZTunnelValues, existingReleaseValues helm.Values, baseValues ...helm.Values,
+) (helm.Values, error) {
```
nrfox, line 75: How is `existingReleaseValues` different from `baseValues`?

pkg/reconcile/ztunnel.go
```diff
@@ -120,6 +128,9 @@ func (r *ZTunnelReconciler) ComputeValues(version string, userValues *v1.ZTunnel
 	if err != nil {
 		return nil, fmt.Errorf("failed to apply user overrides: %w", err)
 	}
+	if err := istiovalues.ApplyNetworkPolicyDefaults(r.cfg.OCPVersion, &finalHelmValues, existingReleaseValues); err != nil {
+		return nil, fmt.Errorf("failed to apply network policy defaults: %w", err)
+	}
```
nrfox, line 133: This should be applied before user overrides. If users provide an override for the network policy then we should respect that.

pkg/reconcile/ztunnel.go
```diff
@@ -130,7 +141,12 @@ func (r *ZTunnelReconciler) ComputeValues(version string, userValues *v1.ZTunnel
 func (r *ZTunnelReconciler) Install(
 	ctx context.Context, version, namespace string, values *v1.ZTunnelValues, ownerRef *metav1.OwnerReference, baseValues ...helm.Values,
 ) error {
-	finalHelmValues, err := r.ComputeValues(version, values, baseValues...)
+	existingReleaseValues, err := helm.GetReleaseValues(ctx, r.cfg.ChartManager, namespace, ztunnelReleaseName)
+	if err != nil {
+		return fmt.Errorf("failed to get existing Helm chart %q: %w", ztunnelChartName, err)
+	}
```
nrfox, line 147: What's the reason for passing the current release values? Do we just need to check if a release exists or not inside `r.ComputeValues`?

Ok I think I can trace what's happening now. We need to know if there's a release in `istiovalues.ApplyNetworkPolicyDefaults` but the `istiovalues` funcs are meant to just set values. Can you move checking if a release exists and getting the current values down into `r.ComputeValues`? That way we don't have to plumb 3 different values sets into `ComputeValues`.

tests/e2e/operator/operator_install_test.go
```diff
@@ -40,11 +41,13 @@ import (
...
 	. "github.com/onsi/gomega"
 	configv1 "github.com/openshift/api/config/v1"
+	crtls "github.com/openshift/controller-runtime-common/pkg/tls"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	networkingv1 "k8s.io/api/networking/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
```
nrfox, line 50: Don't use `ptr` just use the builtin `new()`

cam-garrison: c2721f3 done here

pkg/reconcile/ztunnel.go
cam-garrison: Moved CNI and ztunnel Helm-release lookup into their ComputeValues methods 
c2721f3

pkg/reconcile/ztunnel.go
cam-garrison: agreed, done 
c2721f3

pkg/reconcile/ztunnel.go
cam-garrison: got rid of the redundancy 
c2721f3

controllers/istiocni/istiocni_controller.go
cam-garrison: added note
c2721f3

cmd/main.go
cam-garrison: I added a release note advising users to restart the operator after upgrading openshift if they want new releases to receive the OCP 5 default immediately 
c2721f3

nrfox (approved):

PR #2273, opened 2026-08-20
Fix nondeterministic finalizer ordering

### LINKED ISSUE #2090: [BUG] Nondeterministic finalizer ordering causes spurious diffs
`sets.UnsortedList()` produces nondeterministic finalizer ordering, causing spurious object updates on every reconcile.

In `RemoveFinalizer`:

https://github.com/istio-ecosystem/sail-operator/blob/86fac7a0a15c7e48014af334974ec3e25d6ed276/pkg/kube/finalizers.go#L37-L57

In `AddFinalizer`:

https://github.com/istio-ecosystem/sail-operator/blob/86fac7a0a15c7e48014af334974ec3e25d6ed276/pkg/kube/finalizers.go#L59-L79

## Suggested Fix

Use `SortedList()` instead of `UnsortedList()`.

---
*From code inspection (2026-07-05, main @ 86fac7a0). Severity: 🟡 Low*

sridhargaddam:
In Go, map iteration order is unspecified and isn’t guaranteed to be the same from one iteration to the next, which makes using `UnsortedList()` nondeterministic. While describing the issue as happening on “every reconcile” is a bit of an exaggeration, the underlying bug is real. Using `SortedList()` is harmless regardless of how often it actually occurs. When `AddFinalizer` or `RemoveFinalizer` runs on an object that already has other finalizers, the write can otherwise randomly reorder all of them instead of simply appending or removing the relevant entry.

Fixes: https://github.com/istio-ecosystem/sail-operator/issues/2090

### DIFF
--- a/changelog/fix-finalizer-nondeterministic-order.yaml
+++ b/changelog/fix-finalizer-nondeterministic-order.yaml
@@ -1,8 +1,0 @@
-category: fixed
-title: Fix nondeterministic finalizer ordering
-description: |
-  `AddFinalizer` and `RemoveFinalizer` used `UnsortedList()` when writing
-  finalizers back to the object, which could reorder existing finalizers
-  randomly whenever one was added or removed. They now use `SortedList()`
-  for deterministic ordering.
-issueLink: https://github.com/istio-ecosystem/sail-operator/issues/2090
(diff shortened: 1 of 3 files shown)


dgn (approved):
Thanks!

PR #2292, opened 2026-08-31
Avoid needless webhook config updates

### LINKED ISSUE #2084: [BUG] Webhook probe writes annotations on every tick; period/timeout accept invalid values
**1.** `Reconcile` unconditionally `Update`s the MutatingWebhookConfiguration every period (default 3s) even when the probe status/reason did not change — constant etcd write load and audit-log noise per remote control plane, forever.

https://github.com/istio-ecosystem/sail-operator/blob/86fac7a0a15c7e48014af334974ec3e25d6ed276/controllers/webhook/webhook_controller.go#L90-L101

**2.** `getPeriod`/`getTimeout` parse the annotation with `strconv.Atoi` and accept `0` and negative values: a negative period yields `RequeueAfter < 0` (immediate requeue → hot loop); a zero timeout disables the HTTP timeout entirely.

https://github.com/istio-ecosystem/sail-operator/blob/86fac7a0a15c7e48014af334974ec3e25d6ed276/controllers/webhook/webhook_controller.go#L230-L246

## Suggested Fix

Only call `Update` when the annotation values actually change (compare first); clamp parsed values to a sane minimum (e.g. ≥1s) and fall back to defaults otherwise.

---

sridhargaddam:
The webhook controller called Update on the MutatingWebhookConfiguration on every reconcile (every 3s by default), even when the readiness status and reason hadn't changed. The API server dedups no-op updates so this usually dosent hit etcd, but it still meant a needless API request, admission processing every few seconds for each remote control plane. Now we only call Update when a value actually changes.

Also, the period and timeout annotations were parsed without checking the result, so a zero or negative value slipped through. A negative period made the controller requeue immediately in a tight loop, and a zero timeout turned off the HTTP timeout completely. We now fall back to the defaults for any value that isn't a positive number, and log it at debug level so a misconfigured annotation is easier to spot.

Fixes: https://github.com/istio-ecosystem/sail-operator/issues/2084

### DIFF
--- a/changelog/fix-webhook-probe-redundant-updates.yaml
+++ b/changelog/fix-webhook-probe-redundant-updates.yaml
@@ -1,0 +1,11 @@
+category: fixed
+title: Avoid redundant webhook config updates and reject invalid probe annotations
+description: |
+  The webhook readiness controller called `Update` on the
+  MutatingWebhookConfiguration on every reconcile, even when the readiness
+  status and reason were unchanged, causing needless API requests and
+  audit-log entries for each remote control plane. It now updates only when a
+  value actually changes. Zero, negative, or non-numeric probe period/timeout
+  annotations are also rejected in favor of the defaults, preventing a
+  requeue hot loop and a disabled probe timeout.
+issueLink: https://github.com/istio-ecosystem/sail-operator/issues/2084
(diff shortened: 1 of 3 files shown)


sridhargaddam:
We no longer need this as [PR#2139](https://github.com/istio-ecosystem/sail-operator/pull/2139) includes the necessary changes.

PR #2293, opened 2026-08-31
Return API errors from IstioRevision tag conflict check

### LINKED ISSUE #2082: [BUG] `validateNoTagConflict` swallows non-NotFound API errors
A transient API error from `Client.Get` is indistinguishable from "no tag with this name", so a conflicting IstioRevision can proceed to install and fight with the identically-named IstioRevisionTag over the same webhook resources.

https://github.com/istio-ecosystem/sail-operator/blob/86fac7a0a15c7e48014af334974ec3e25d6ed276/controllers/istiorevision/istiorevision_controller.go#L179-L187

Note that the mirror-image check in the IstioRevisionTag controller already handles this correctly:

https://github.com/istio-ecosystem/sail-operator/blob/86fac7a0a15c7e48014af334974ec3e25d6ed276/controllers/istiorevisiontag/istiorevisiontag_controller.go#L131-L139

## Suggested Fix

`if err != nil && !apierrors.IsNotFound(err) { return err }` — match the tag controller's existing pattern.

---

sridhargaddam:
validateNoTagConflict only reacted when Client.Get returned no error, so a transient API error looked the same as "no tag with this name". That let a conflicting IstioRevision pass validation and install, fighting an identically-named IstioRevisionTag over the same webhook resources. Now we return any non-NotFound error, matching the check that already exists in the IstioRevisionTag controller.

Fixes: https://github.com/istio-ecosystem/sail-operator/issues/2082

### DIFF
--- a/controllers/istiorevision/istiorevision_controller.go
+++ b/controllers/istiorevision/istiorevision_controller.go
@@ -182,6 +182,8 @@ func (r *Reconciler) validateNoTagConflict(ctx context.Context, rev *v1.IstioRev
 		if validation.ResourceTakesPrecedence(&tag.ObjectMeta, &rev.ObjectMeta) {
 			return reconciler.NewNameAlreadyExistsError("an IstioRevisionTag exists with this name", nil)
 		}
+	} else if !apierrors.IsNotFound(err) {
+		return err
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


nrfox (approved):

dgn (approved):
Thank you @sridhargaddam

PR #2302, opened 2026-09-02
Make bundle-publish idempotent by reusing existing PRs

FilipB:
When re-running the release workflow, the PR creation step would fail with a 422 because a PR from the same branch already exists. Now the script checks for an existing open PR first and updates it via PATCH instead of unconditionally creating a new one.


mkralik3 (approved):

dgn (approved):
Thanks @FilipB!

PR #2315, opened 2026-09-04
Adapt TLSAdherence test to support OCP 5.0+

mkralik3:
TLSAdherence is a Tech Preview feature controlled by the FeatureGate resource in OCP 4.22+. Although it is expected to reach GA in version 5.0, it is not yet GA.

To support current testing, the version comparison logic was updated as follows:

- Verify whether the cluster version is 4.22+ using the full version string rather than just the minor version.
- Check if TLSAdherence is listed under the FeatureGate resource; 
  - if present, enable it.
  - If it is not under the FeatureGate resource, assume TLSAdherence is a GA feature.


tests/e2e/operator/operator_install_test.go
```diff
@@ -253,54 +253,69 @@ spec:
...
+						Eventually(func(g Gomega) {
+							fg := &configv1.FeatureGate{}
+							g.Expect(cl.Get(ctx, client.ObjectKey{Name: "cluster"}, fg)).To(Succeed())
+							found := false
+							for _, details := range fg.Status.FeatureGates {
+								for _, enabled := range details.Enabled {
+									if enabled.Name == "TLSAdherence" {
+										found = true
+										break
+									}
```
nrfox, line 315: There's a lot of nesting going on now. I think we'll want to do other refactors to make this test more manageable like fetching the ocp version in the suite setup. But this can be done after.

nrfox (approved):

tests/e2e/operator/operator_install_test.go
mkralik3: yeah :( , I wanted to catch all cases so

- when cluster is 4.22+
  - if yes, check if TLSAdherence is in feature gate
    - if yes, check if TLSAdherence is disabled 
      - if yes, enable it
    - if no, already enabled, no need to do anything
  - if no, cluster is 4.22+ so assuming TLSAdherence is GA, no need to do anything
- if no, 4.21 and lower don't have TLSAdherence

PR #2324, opened 2026-09-09
Use quay.io/sail-dev for 1.31+

nrfox:
#### What type of PR is this?

- [x] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [ ] Test
- [ ] Documentation Update

Istio is no longer publishing images to registry.istio.io so we will copy them to `quay.io/sail-dev` in the update deps job.

Fixes #

Related Issue/PR #

### DIFF
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -47,7 +47,7 @@ ISTIO_PROXY_IMAGE_NAME ?= proxyv2
 
 # GitHub creds
 GITHUB_USER ?= openshift-service-mesh-bot
-GITHUB_TOKEN ?= 
+GITHUB_TOKEN ?=
 
 SOURCE_DIR := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
 
@@ -573,7 +573,7 @@ update-istio-samples: ## Update the Istio samples files located in the samples f
 	@hack/update-istio-samples.sh
 
 # Istio minor releases component images are only publisheed to Docker Hub. We mirror these
-# quay.io/sail-dev. This is the single source of
+# to quay.io/sail-dev. This is the single source of
 # truth: hack/istio-hub.sh reads it to decide both which images to mirror and which hub
 # to write into the downloaded charts. Add a minor here when Istio ships a new one.
 # TODO: Remove this once 1.30 drops out of support.
@@ -581,7 +581,7 @@ MIRROR_ISTIO_MINORS ?= 1.31
 
 .PHONY: mirror-istio-images
 mirror-istio-images: crane ## Mirror the Istio operand images that upstream only publishes to Docker Hub into quay.io/sail-dev.
-	@CRANE=$(CRANE) hack/mirror-istio-images.sh $(MIRROR_ISTIO_MINORS)
+	@CRANE=$(CRANE) hack/mirror-istio-images.sh $(VERSIONS_YAML_DIR)/$(VERSIONS_YAML_FILE) $(MIRROR_ISTIO_MINORS)
 
 .PHONY: update-deps
 update-deps: ## Update all dependencies including tools, Go modules, and operator components.
@@ -653,7 +653,7 @@ $(OPERATOR_SDK): $(LOCALBIN)
 
 # ISTIOCTL_DOWNLOAD_URL defines the url where istioctl will be downloaded
 # By default, it is not set and it uses the istio/istio release download artifact
-ISTIOCTL_DOWNLOAD_URL ?= 
+ISTIOCTL_DOWNLOAD_URL ?=
 
 # ISTIOCTL_FROM_CONTAINER_IMAGE defines whether istioctl should be downloaded by extracting it from a container image.
 # If set to true, the istioctl binary will be pulled from a specified container image instead of the default release artifact.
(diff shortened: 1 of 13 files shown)


Makefile.core.mk
```diff
@@ -572,6 +572,17 @@ update-istio: ## Update the Istio commit hash in the 'latest' entry in versions.
 update-istio-samples: ## Update the Istio samples files located in the samples folder to match the latest Istio upstream version of the charts.
 	@hack/update-istio-samples.sh
 
+# Istio minor releases component images are only publisheed to Docker Hub. We mirror these
+# quay.io/sail-dev. This is the single source of
```
FilipB:
```suggestion
# to quay.io/sail-dev. This is the single source of
```

hack/mirror-istio-images.sh
```diff
@@ -0,0 +1,120 @@
...
+set -euo pipefail
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+REPO_ROOT=$(dirname "${SCRIPT_DIR}")
+
+# shellcheck source=hack/istio-hub.sh
+source "${SCRIPT_DIR}/istio-hub.sh"
+
+VERSIONS_YAML_DIR=${VERSIONS_YAML_DIR:-"${REPO_ROOT}/pkg/istioversion"}
+VERSIONS_YAML_FILE=${VERSIONS_YAML_FILE:-"versions.yaml"}
```
FilipB: Don't we need to pass this variable from the Makefile explicitly? We have to respect whatever Makefile sets as this variable might be changed in Makefile.vendor.mk

nrfox: Yes that's a good point. If you are calling this from the `Makefile` the variable would need to be exported first. Probably better to make this an argument to the script rather than an env var that gets defaulted.

nrfox: Alright this is passed through to the script as an arg now rather than an env var

nrfox:
@FilipB I can run the script manually to get e2e tests passing?

FilipB:
> @FilipB I can run the script manually to get e2e tests passing?

Let's do that. I assume it will be also necessary to manually mark those images as public in quay.io after first push.

hack/istio-hub.sh
```diff
@@ -0,0 +1,66 @@
...
+UPSTREAM_HUB=${UPSTREAM_HUB:-"registry.istio.io/release"}
+
+# Operand images referenced by the Istio charts.
+ISTIO_IMAGES=(pilot proxyv2 install-cni ztunnel)
+
+# usesMirror returns 0 if the images for the given Istio version are served from
+# MIRROR_HUB rather than UPSTREAM_HUB.
+# $1: the Istio version, e.g. 1.31.0 or 1.31.0-beta.1
+function usesMirror() {
+  local version="$1" minor
```
MaxBab: What does "minor" here stands for?

Makefile.core.mk
```diff
@@ -572,6 +572,17 @@ update-istio: ## Update the Istio commit hash in the 'latest' entry in versions.
 update-istio-samples: ## Update the Istio samples files located in the samples folder to match the latest Istio upstream version of the charts.
 	@hack/update-istio-samples.sh
 
+# Istio minor releases component images are only publisheed to Docker Hub. We mirror these
```
MaxBab: Typo in "published"

hack/mirror-istio-images.sh
```diff
@@ -0,0 +1,123 @@
...
+  fi
+
+  for version in ${versions}; do
+    echo "mirroring ${version}"
+    for image in "${ISTIO_IMAGES[@]}"; do
+      src="${SOURCE_HUB}/${image}:${version}"
+      dst="${MIRROR_HUB}/${image}:${version}"
+
+      src_digest=$("${CRANE}" digest "${src}" 2>/dev/null || true)
+      if [ -z "${src_digest}" ]; then
```
MaxBab, line 114: The "crane digest" failure always counted as missing image.
But it could be any other issue as well, like network or auth issue.

Worth verifying the error.
It just a misleading thing, not a global issue.

hack/download-charts.sh
```diff
@@ -189,14 +193,29 @@ EOF
...
+  local hub="${UPSTREAM_HUB}"
+  if usesMirror "${ISTIO_VERSION}"; then
+    hub="${MIRROR_HUB}"
+  fi
+
+  echo "setting hub to ${hub} in all charts"
+
+  local sed_args=() known_hub
+  for known_hub in "${SOURCE_HUB}" "${UPSTREAM_HUB}" "${MIRROR_HUB}"; do
+    sed_args+=(-e "s|hub: ${known_hub}|hub: ${hub}|g")
```
MaxBab, line 218: When `hub` is set to "MIRROR_HUB" in line 208, the sed replaces itself in a loop.

nrfox:
> Let's do that. I assume it will be also necessary to manually mark those images as public in quay.io after first push.

@FilipB I don't have access to push but here's the command I tried:
```
hack/mirror-istio-images.sh pkg/istioversion/versions.yaml 1.31
```

fjglira (approved):
lgtm and all the test passed already. Adding a hold until Filip or Maxim take a look because they already did review also

MaxBab (approved):

MaxBab:
I'm good with the changes.

FilipB (approved):

PR #2325, opened 2026-09-10
Reconcile OwnerRef on IstioRevision object during update

### LINKED ISSUE #2083: [BUG] `revision.CreateOrUpdate` never reconciles ownerReferences (or namespace)
On the update path only `Spec.Version` and `Spec.Values` are written; `OwnerReferences` are set only at creation. If the owner reference is lost or mutated (manual edit, restore from backup, velero-style re-creation), the Istio controller silently loses prune/adoption semantics (`ListOwned` matches by owner UID) and the orphaned revision is never pruned nor reflected in `status.revisions`.

Update path (only sets Version and Values):

https://github.com/istio-ecosystem/sail-operator/blob/86fac7a0a15c7e48014af334974ec3e25d6ed276/pkg/revision/reconcile.go#L41-L48

Create path (sets OwnerReferences):

https://github.com/istio-ecosystem/sail-operator/blob/86fac7a0a15c7e48014af334974ec3e25d6ed276/pkg/revision/reconcile.go#L50-L61

## Suggested Fix

On update, ensure the expected owner reference is present (add if missing), and set `rev.Spec.Namespace = namespace` (a no-op while the field is immutable, self-healing if that ever changes).

---

MaxBab:
#### What type of PR is this?

- [X] Bug Fix
- [ ] Refactor
- [ ] Optimization
- [ ] Test
- [ ] Documentation Update

When revision.CreateOrUpdate reconciles during update, the OwnerReference is not being updated, which could lead to a state, where the orphaned revision is never pruned or reflected in "status.revisions".

Add OwnerReference during reconcile update.

Fixes #2083 

Related Issue/PR #

### DIFF
--- a/pkg/revision/reconcile.go
+++ b/pkg/revision/reconcile.go
@@ -38,33 +38,39 @@ func CreateOrUpdate(
 		return fmt.Errorf("failed to get active IstioRevision: %w", err)
 	}
 
+	rev.Spec.Version = version
+	rev.Spec.Namespace = namespace
+	rev.Spec.Values = values
+	setOwnerReference(&rev, ownerRef)
+
 	if found {
 		// update
-		rev.Spec.Version = version
-		rev.Spec.Values = values
 		log.Info("Updating IstioRevision")
 		if err = cl.Update(ctx, &rev); err != nil {
 			return fmt.Errorf("failed to update IstioRevision %q: %w", rev.Name, err)
 		}
 	} else {
 		// create new
-		rev = v1.IstioRevision{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:            revName,
-				OwnerReferences: []metav1.OwnerReference{ownerRef},
-			},
-			Spec: v1.IstioRevisionSpec{
-				Version:   version,
-				Namespace: namespace,
-				Values:    values,
-			},
-		}
+		rev.Name = revName
 		log.Info("Creating IstioRevision")
 		if err = cl.Create(ctx, &rev); err != nil {
 			return fmt.Errorf("failed to create IstioRevision %q: %w", rev.Name, err)
 		}
 	}
 	return nil
+}
+
+// setOwnerReference ensures the expected ownerReference is present on the revision.
+// It replaces any existing ownerRef with the same Kind+Name (e.g. after the owner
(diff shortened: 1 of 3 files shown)


pkg/revision/reconcile.go
```diff
@@ -67,6 +60,19 @@ func CreateOrUpdate(
...
+// was re-created with a new UID), or appends if none matches.
+func setOwnerReference(rev *v1.IstioRevision, ownerRef metav1.OwnerReference) {
+	for i, ref := range rev.OwnerReferences {
+		if ref.Kind == ownerRef.Kind && ref.Name == ownerRef.Name {
+			rev.OwnerReferences[i] = ownerRef
+			return
+		}
+	}
+	rev.OwnerReferences = append(rev.OwnerReferences, ownerRef)
+}
```
nrfox, line 74: nit: you can use [this helper](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.25.0/pkg/controller/controllerutil#SetControllerReference) rather than defining a new one. You may need to alter the args passed to `CreateOrUpdate` a bit.

dgn: I guess you mean [`SetOwnerReference()`](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.25.0/pkg/controller/controllerutil#SetOwnerReference)? That might be a good idea, but we'd have to refactor how to create the ownerRef a little bit..

dgn: maybe for a follow-up?

dgn (approved):
LGTM, thanks @MaxBab

MaxBab:
@dgn @nrfox 

Hi Daniel,

Thanks for the approval.
I actually started to work on changing the PR based on Nick suggestion.

His suggestion of using the `SetControllerReference` rather than `SetOwnerReference` looks correct to me, because the `SetControllerReference` sets `controller: true` and `BlockOwnerDeletion: true` automatically. Also, it makes sure that the object is owned by a single controller and makes sure it remain that way.

All that is not done by the `SetOwnerReference` function.

I'll create a new PR to apply the use of the `SetControllerReference` function and remove the duplication I've created by a manual `setOwnerReference` function.

PR #2329, opened 2026-09-14
Use standard "SetControllerReference" func for IstioRev reconcile

MaxBab:
Replace the custom "setOwnerReference" helper function with the standard "controllerutil.SetControllerReference", which runs on both create and update paths, self-healing any stale or missing ownerRef.

 

#### What type of PR is this?

- [ ] Bug Fix
- [X] Refactor
- [ ] Optimization
- [ ] Test
- [ ] Documentation Update

Replace the custom "setOwnerReference" helper function with the standard "controllerutil.SetControllerReference", which runs on both create and update paths, self-healing any stale or missing ownerRef.

Fixes #

Related Issue/PR #


MaxBab:
Hi @dgn @nrfox 

This is the follow-up pr, of the [following](https://github.com/istio-ecosystem/sail-operator/pull/2325#issuecomment-5660941112).

changelog/use-setcontrollerreference-func-for-istiorev-reconcile-flow.yaml
dgn, line 1: I don't think this needs a changelog; it's not user-facing

MaxBab: NP. Will remove that.

MaxBab: Removed.

dgn (commented):
Generally LGTM, my only concern is setting `BlockOwnerDeletion: true` will result in a slight behavior change in the GC. We should investigate if that might cause disruptions

MaxBab:
> Generally LGTM, my only concern is setting `BlockOwnerDeletion: true` will result in a slight behavior change in the GC. We should investigate if that might cause disruptions

The `BlockOwnerDeletion` controls the deletion ordering, by making sure the owner object is not being deleted before the owned objects.

https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/

Also, that flow is relevant during foreground cascading deletion.
During background deletion, the owner is being deleted immediately and dependents are being cleaned afterward.

dgn:
> > Generally LGTM, my only concern is setting `BlockOwnerDeletion: true` will result in a slight behavior change in the GC. We should investigate if that might cause disruptions
> 
> The `BlockOwnerDeletion` controls the deletion ordering, by making sure the owner object is not being deleted before the owned objects.
> 
> https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/
> 
> Also, that flow is relevant during foreground cascading deletion. During background deletion, the owner is being deleted immediately and dependents are being cleaned afterward.

Right. But we're already using finalizers to achieve that same behavior, so we might be able to remove that code?

dgn (approved):
Nevermind, we previously set `BlockerOwnerDeletion: true` anyway, and the reconcilers on the operator CRs we use are for a different thing. LGTM

PR #2330, opened 2026-09-14
fix: map new upstream structpb.Struct fields in api_transformer

dgn:
Map new/renamed CNI, Pilot and Waypoint fields to concrete k8s types in transform.yaml. Also make api_transformer fail fast with the exact field names when a type mapping is missing, and run goimports on the output so unused addImports entries don't break the build.

Example output of the changes in api_transformer:
```
Upstream added or renamed field(s) that transform.yaml doesn't know how to map yet:
  CNIConfig.UpdateStrategy: *structpb.Struct
  CNIConfig.SeLinuxOptions: *structpb.Struct
  CNIConfig.Tolerations: []*structpb.Struct
  PilotConfig.InitContainers: []*structpb.Struct
  PilotConfig.Pdb: *structpb.Struct
  PilotConfig.DeploymentAnnotations: *structpb.Struct
  PilotConfig.SidecarInjectorWebhookAnnotations: *structpb.Struct
  WaypointConfig.Tolerations: []*structpb.Struct
Add a replaceFieldTypes (or preserveTypes) entry for each field above in hack/api_transformer/transform.yaml, then re-run `make gen`.
exit status 1
```

This should help with future fixes.


nrfox (approved):
Much better output

PR #2335, opened 2026-09-15
docs: clarify RBAC trust model for cluster-scoped CRs

### LINKED ISSUE #2114: [SECURITY] No admission webhook validates `spec.namespace`, allowing namespace escalation
The operator has **no admission webhooks** (no `CustomValidator`, `CustomDefaulter`, or `SetupWebhookWithManager`). The only validation is declarative CEL rules that make `spec.namespace` immutable *after creation* — but there is no validation on the initial value. A user with `create` permission on the CR can target any namespace for Istio deployment.

Istio CRD — namespace immutable after creation but not validated on create:

https://github.com/istio-ecosystem/sail-operator/blob/86fac7a0a15c7e48014af334974ec3e25d6ed276/api/v1/istio_types.go#L34

https://github.com/istio-ecosystem/sail-operator/blob/86fac7a0a15c7e48014af334974ec3e25d6ed276/api/v1/istio_types.go#L60

Same pattern in IstioCNI:

https://github.com/istio-ecosystem/sail-operator/blob/86fac7a0a15c7e48014af334974ec3e25d6ed276/api/v1/istiocni_types.go#L43-L47

Same pattern in ZTunnel:

https://github.com/istio-ecosystem/sail-operator/blob/86fac7a0a15c7e48014af334974ec3e25d6ed276/api/v1/ztunnel_types.go#L35-L38

## Suggested Fix

Add a validating admission webhook (or CEL validation rule) that restricts `spec.namespace` to namespaces the creating user has privileges in.

---
*From code inspection (2026-07-05, main @ 86fac7a0). Severity: 🔴 High*

dgn:
Creating/updating Istio, IstioRevision, IstioCNI, or ZTunnel resources is equivalent to cluster-admin, since the operator's own ServiceAccount can act on any namespace on the caller's behalf.

Fixes #2114.


FilipB (commented):
While the documentation is clear we are still not forcing this as the issue is suggesting. It's probably not worth the effort as anyone who is able to create/edit cluster-wide resources should be consider as cluster admin? There is no room for error in this case?

dgn:
> While the documentation is clear we are still not forcing this as the issue is suggesting. It's probably not worth the effort as anyone who is able to create/edit cluster-wide resources should be consider as cluster admin? There is no room for error in this case?

From my perspective, we have two options:
a) continue with the current policy of simply accepting that access to sail CRs equals cluster-admin access
b) build a complex policy engine that verifies for each CR edit that the user in question does actually have permissions for all changes that will be applied in the cluster

I think to fully solve the issue, b) is the only option, but I'm not sure it's worth the effort, or even really viable*. Giving these sorts of guarantees would also mean that every possible privilege escalation is a critical bug in Sail Operator, creating a lot of work for us.

*: we could e.g. only verify this at edit-time using a webhook. At reconciliation time, we simply cannot know which user made which changes to the CR. So effectively, any user with edit permission on webhook configs can circumvent the checks and use Sail Operator for privilege escalation.

FilipB:
> > While the documentation is clear we are still not forcing this as the issue is suggesting. It's probably not worth the effort as anyone who is able to create/edit cluster-wide resources should be consider as cluster admin? There is no room for error in this case?
> 
> From my perspective, we have two options: a) continue with the current policy of simply accepting that access to sail CRs equals cluster-admin access b) build a complex policy engine that verifies for each CR edit that the user in question does actually have permissions for all changes that will be applied in the cluster
> 
> I think to fully solve the issue, b) is the only option, but I'm not sure it's worth the effort, or even really viable*. Giving these sorts of guarantees would also mean that every possible privilege escalation is a critical bug in Sail Operator, creating a lot of work for us.
> 
> *: we could e.g. only verify this at edit-time using a webhook. At reconciliation time, we simply cannot know which user made which changes to the CR. So effectively, any user with edit permission on webhook configs can circumvent the checks and use Sail Operator for privilege escalation.

Thank you. Option a) seems good enough.
Approving and adding hold so others can also review.

FilipB (approved):

docs/general/getting-started.adoc
```diff
@@ -31,6 +32,15 @@ link:../README.adoc[Return to Project Root]
 
 [[getting-started]]
 
+[[rbac-and-trust-model]]
+== RBAC and Trust Model
+
+The `Istio`, `IstioRevision`, `IstioCNI`, and `ZTunnel` resources are cluster-scoped, and the operator's own `ServiceAccount` is granted broad, cluster-wide permissions (including the ability to create and bind `RoleBinding`/`ClusterRoleBinding` objects, deploy privileged workloads, and manage `Secret`/`ServiceAccount` objects in any namespace) so that it can reconcile the control plane into whatever namespace `spec.namespace` names.
+
+Because of this, granting a user permission to create or update these resources is equivalent to granting that user `cluster-admin`: they can direct the operator's elevated privileges at any namespace in the cluster, including namespaces they otherwise have no access to. Do not delegate `create`/`update` permissions on `Istio`, `IstioRevision`, `IstioCNI`, or `ZTunnel` to any user that is not already fully trusted at the cluster-admin level.
```
nrfox, line 40: > Because of this, granting a user permission to create or update these resources is equivalent to granting that user `cluster-admin`: they can direct the operator's elevated privileges at any namespace in the cluster

They only can interact with these resources indirectly though. A user can WRITE an `Istio` resource but may not be able to interact with the resources that the operator creates for the `Istio` resource. In other words, creating an `Istio` resource may cause a `Secret` to be created in namespace `istio-system` but it doesn't give you access to that secret.

This could change if we gave users the ability to customize the helm templates.

PR #2340, opened 2026-09-16
fix: go get in update_deps.sh should not use -u

dgn:
When the -u argument is used, transitive dependencies are updated to their latest version, irrespective of what's actually required by the module in question. This broke our update-deps PRs because it updated structured-merge-diff to v7 which is API-incompatible with v6.


nrfox (approved):

FilipB:
Lint is failing because 1.31.0 was added to release-1.31. I will add 1.31.0 to main as well.

FilipB:
Lint should be fixed by https://github.com/istio-ecosystem/sail-operator/pull/2343

Issue #2338, opened 2026-09-15
[FEATURE] Introduce analytic custom metrics and a controller for recording APIs, custom resources usage


yxun:
### Is this the right place to submit this?

- [x] This is not a question about how to use the sail-operator

### Feature Description

This is a feature request about adding a set of custom metrics that help to measure sail operator and Istio component APIs usage. For example, 

- Istio version counts (number of control planes at each Istio version)
- ZTunnel proxy counts (number of ZTunnel proxies)
- Envoy Proxy counts (number of Envoy Proxies managed by an Istiod)
- Namespace counts (number of namespaces configured in Istio sidecar mode or Istio Ambient mode)

Because the operator is running a controller runtime manager and it's metrics server is exposing registered metrics under its `deploymentName + "-metrics-service"` path /metrics.

The goal is to have a new package for analytics purpose and define new custom metrics type as standard Prometheus metric types.
Next, implement a controller for watching the operator API custom resource events and recording metric count value.

**Benefits**

- Outlines Operator APIs and components usage
- Help integrating with a telemetry pipeline from the operator to the cluster Prometheus instance. 

**Proposed solution**

- Implement a package `analytics` for introducing custom metric types and configuration resources.
- Implement a metric doc generator for adding a doc page.
- Implement an `analytics` controller for watching events and recording metric value

Non-goals
- Modifying the existing CRDs or APIs
- Aggregating custom metrics data from all operator instances

### Additional Information

Related work: OSSM-6616 
https://github.com/istio-ecosystem/sail-operator/pull/2349
https://github.com/openshift-service-mesh/sail-operator/pull/1132

PR #2349, opened 2026-09-17
Analytics: Add operator analytics metrics and controller

yxun:
This PR adds a package `analytics` that aims to introduce custom metrics for Sail Operator components. A new analytics_controller is implemented for recording metric values when custom resources such as `Istio`, `IstioRevision`, `ZTunnel` are created or deleted. It also creates `ServiceMonitor` and `PrometheusRule` resources for integrating with in-cluster Prometheus or telemeter.

By default, Sail operator does not run this new feature. This PR introduces a new environment variable `ENABLE_ANALYTICS` for the operator. When it is set, the custom metric controller starts running.

More details are in the `enhancements/SPE6616-analytics-metrics.md`

 Enhancement / New Feature
 
Reference:
https://github.com/maistra/istio-operator/pull/1119

Resolve:
https://github.com/istio-ecosystem/sail-operator/issues/2338


enhancements/SEP6616-analytics-metrics.md
nrfox, line 1: Hey @yxun this change is larger than I thought it would be. I was expecting there to be some custom metrics added to the builtin controller-runtime [registry](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/metrics@v0.25.1#RegistererGatherer) and a loop that sets these custom metrics every x seconds. These metrics are exported out through [this service](https://github.com/istio-ecosystem/sail-operator/blob/main/chart/templates/auth_proxy_service.yaml).

Do we need to dynamically create a `ServiceMonitor`? Can this instead be static configuration that you can include/exclude with the operator chart through helm values? We can include it by default with the OLM bundle.

These two changes, I think, would greatly reduce the size of this change:
1. Export metrics every x seconds. Reuse controller-runtime metrics registry. Basically list all CRs and call `gauge.Set(...)` every x seconds. You don't need to setup `Watch` on resources for this. You may still need to handle removing old labels though.
2. Move `ServiceMonitor` to static configuration and include with bundle/chart.

At that point you shouldn't need a SEP for this as the feature is relatively small and there's no API changes.

nrfox (changes requested):

enhancements/SEP6616-analytics-metrics.md
yxun: Hi @nrfox , 

> I was expecting there to be some custom metrics added to the builtin controller-runtime [registry](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/metrics@v0.25.1#RegistererGatherer) and a loop that sets these custom metrics every x seconds.

I checked the operator code and tests. I am following this existing e2e tests :
https://github.com/istio-ecosystem/sail-operator/blob/main/tests/e2e/operator/operator_install_test.go#L210

We have `sail-operator-metrics-service`  and `sailoperator-metrics-reader` ClusterRole ready in the bundle. However, we still need to create a `ClusterRoleBinding` for allowing access the metrics operator path.

There is **no existing loop** that sets Sail Operator's custom metrics other than controller runtime default ones.

This PR is registering custom metrics to the builtin controller-runtime registry (See the implementation in `pkg/analytics/metrics.go func ReisterMetrics`)

> Do we need to dynamically create a ServiceMonitor? 

`ServiceMonitor` need to be created so that in cluster Prometheus can scrape those metrics from operator. I can follow your suggestion for putting that in the operator chart through helm values.

> Reuse controller-runtime metrics registry. Basically list all CRs and call gauge.Set(...) every x seconds. 

Jamie wants to collect custom metrics about how many namespaces enrolled in Istio sidecar mode and how many in Istio ambient mode. So I still need to watch resources such as `namespace` update events when they get label updates. 

If that causes too large changes , I can move those custom metrics implementation in next phase. But eventually, we need a design and a controller for recording and sending more custom metrics such as Istio namespaces, Envoy proxy version counts, multicluster mesh cluster counts etc.

yxun: Because there is no existing loop that sets Sail Operator's custom metrics e.g. Istio, IstioRevision, ZTunnel counts, How to ` list all CRs and call gauge.Set(...) every x seconds.` without a new controller ? There is no goal for changing existing controller and API.

nrfox: You're right there's not an existing controller for this today but it can be done simply with
```
ticker := time.Ticker(time.Minute)
for {
select {
 case <-ctx.Done():
   return
 case <-ticker.C:
   // List all the resources and call gauge.Set
}
}
```
It's ok for this use case if we don't respond immediately to events by watching all resources. There is some delay on the scrape as well.

In the controller model, what object is being reconciled in the request
```
Reconcile(ctx context.Context, req ctrl.Request)
```

yxun: thanks, I will try your suggestion above today and I'm moving this PR controller objects to static helm and bundle configuration.

pkg/scheme/scheme.go
```diff
@@ -38,5 +40,9 @@ func init() {
 	utilruntime.Must(v1alpha1.AddToScheme(Scheme))
 	utilruntime.Must(v1.AddToScheme(Scheme))
 	utilruntime.Must(apiextensionsv1.AddToScheme(Scheme))
+	utilruntime.Must(monitoringv1.AddToScheme(Scheme))
```
nrfox: I don't think we need this anymore as part of this change since the metrics are static configuration now?

chart/templates/olm/metrics_reader_clusterrole.yaml
nrfox: Same with these roles? The operator shouldn't need these permissions anymore for exporting metrics if we are including the service/podmonitors as static configuration.

pkg/scheme/scheme.go
yxun: This `utilruntime.Must(monitoringv1.AddToScheme(Scheme))` was aligned with Mayleigh's PR change. I have merged those changes from main branch. This PR does not update that scheme.

This PR adds `analyze.RegisterMetrics()` for registering custom metrics to the default metric server.

chart/templates/olm/metrics_reader_clusterrole.yaml
yxun: I am working on an e2e test for checking permissions. 
Yes, we don't need that metrics reader role and binding if using the default sail operator ClusterRole defined in `chart/templates/rbac/role.yaml` . I was thinking to use a new ClusterRole specifically for reading metrics.

yxun: I am working on a permission issue. 

Because the [this service] https://github.com/istio-ecosystem/sail-operator/blob/main/chart/templates/auth_proxy_service.yaml only has https port.

And the existing [metrics-reader ClusterRole] https://github.com/istio-ecosystem/sail-operator/blob/main/chart/templates/auth_proxy_client_clusterrole.yaml only has infra rules.

The OpenShift UWM skips ServiceMonitor which uses `bearerTokenFile` / TLS files. UWM is not allowed to use that for security concern.


In this case, we need to ensure the existing `ServiceAccount` used by the in cluster `Prometheus` is bound to the sail operator `ClusterRole` via a `ClusterRoleBinding`. So Prometheus has cross-namespace permission to read the service and metric endpoint. 
And the `ServiceMonitor` can use the `ServiceAccount` token from its `authorization` field.

pkg/analyze/metrics.go
```diff
@@ -0,0 +1,272 @@
...
+	client.Client
+	interval time.Duration
+	ticker   *time.Ticker
+	done     chan struct{}
+}
+
+// metricsDescription is a map of string keys (metrics) to MetricDescription values (Name, Help).
+var metricDescription = map[string]MetricDescription{
+	"IstioVersionTotal": {
+		Name: "servicemesh_istiod_total",
```
dgn: let's make the `servicemesh_` prefix configurable and use the .Values.name from the chart

chart/templates/olm/metrics_reader_servicemonitor.yaml
```diff
@@ -0,0 +1,24 @@
+{{ if .Values.bundleGeneration }}
```
dgn: we should also gate on whether we're creating an OLM bundle for OCP

PR #2359, opened 2026-09-21
e2e: improve test robustness through retries

dgn:
We have a bunch of places where we e.g. exec into pods right when they report healthy, which can fail intermittently in CI. This patch adds a bunch of Eventually() wrappers and retries in some of those places.


dgn:
This should fix flakes like https://prow.istio.io/view/s3/istio-prow/pr-logs/pull/istio-ecosystem_sail-operator/2356/e2e-kind-dualstack_sail-operator_main/2102021930277146624 or https://prow.istio.io/view/s3/istio-prow/pr-logs/pull/istio-ecosystem_sail-operator/2356/e2e-kind-dualstack_sail-operator_main/2102017966483705856

fjglira (approved):

PR #2362, opened 2026-09-22
[release-1.31] Add Istio versions 1.31.1, 1.30.5, 1.29.8

mkralik3:
## Summary
- Add missing Istio versions: 1.31.1, 1.30.5, 1.29.8
- Update VERSION to 1.31.1 ~~and PREVIOUS_VERSION to 1.31.0~~ (reverted since 1.31.0 was not released yet. so still the latest previous OLM version is 1.30.4)
- Pin `istio.io/istio` to the 1.31.1 release commit and `istio.io/api` to v1.31.1


FilipB:
@bmangoen what is the plan with 1.31.0? Do you want to release it as it is now or include 1.31.1?

bmangoen:
> @bmangoen what is the plan with 1.31.0? Do you want to release it as it is now or include 1.31.1?

We have to discuss with the team about this but IMO we need to include Istio 1.31.1 into Sail Operator 1.31.0 (adding a `do-not-merge/hold` for now). 

@nrfox also may have changes to include to it. 

cc. @dgn

fjglira:
I'm in favor to release directly with 1.31.1

mkralik3:
why not release directly 1.31.1 ?  so path will be 1.30.4 -> 1.31.1 

I know from product perspective it is weird to have first release with Z version = 1  and not 0, but since sail operator version is in line with istio version, it make sense to release the one with latest istio version. 

BTW, if changes are midstream related, it doesn't need to block upstream release (midstream is sync up periodically with release branches, not only with upstream release)

FilipB:
> why not release directly 1.31.1 ? so path will be 1.30.4 -> 1.31.1
> 
> I know from product perspective it is weird to have first release with Z version = 1 and not 0, but since sail operator version is in line with istio version, it make sense to release the one with latest istio version.
> 

I don't remember details why but we don't want to do that - https://github.com/istio-ecosystem/sail-operator#versioning-and-support-policy
`Also, the Sail Operator patch version will not correspond to the Istio patch version. For example, the 1.25.0 Sail Operator may only support installing Istio 1.25.1 but not Istio 1.25.0.`

mkralik3:
@FilipB  that will be confusing than.
If we strictly want to release 1.31.**0** , than release it without 1.31.1 (hold this PR) and only after that, do another release again for 1.31.1 and this PR. 
What is point to use the same version as istio when the versions will be de-synchronized to each other?

BTW, in 1.30, there were 1.30.0 and than 1.30.3 Sail Operator release since we forget to do sail operator release with 1.30.1 and 1.30.2 istio release.

bmangoen:
@mkralik3 @FilipB 
Actually I already triggered the [GH release workflow for 1.31.0](https://github.com/istio-ecosystem/sail-operator/actions/runs/35351716426/job/105621191635) (it failed at creating the GH release step), https://github.com/redhat-openshift-ecosystem/community-operators-prod/pull/11219 was created and merged.
So we have to release 1.31.0 as is now

mkralik3:
@bmangoen  ack, so I will revert [Fix PREVIOUS version](https://github.com/istio-ecosystem/sail-operator/pull/2362/changes/cd80cf1ebb696be2faa8269e5900fa3c1abe3f41) from this PR since 1.30.0 as previous version will be valid

FilipB (approved):

